PR #170, opened 2022-12-01
chore(EMS-585): bump govuk-frontend

ttbarnes:
This PR updates govuk-frontend to the latest version. Also: 

- Add `preventDoubleClick` config for all buttons.
- Use GOV header component instead of hard coded header.
- Simplify header E2E tests by removing the SVG check. We are not rendering this ourselves now as it's handled by the library/component itself. We now only test what we are controlling/specifying i.e the GOV link and service name/link.

Note: currently we cannot consume the GOV footer component due to current design. This may change in the future.


IgnatG (approved):
LGTM

PR #176, opened 2022-12-06
feat(EMS-503): Policy and export - Type of policy 'save and go back' functionality

ttbarnes:
This PR adds the first "save and go back" functionality to an EXIP insurance/application.

## Summary of changes

- Add "save and go back" route/controller for the "type of policy" page.
- Create "save data" function for all pages in the `policy-and-export` section of an application. This function:
  - Strips any invalid fields from submitted form data, if an `errorList` is provided.
  - Gets the application.
  - Sanitises the submitted form data using existing function (e.g transform a string of 'true' to a boolean).
  - Get the application's `policyAndExport` id from the returned application.
  - Call the API to update the application's `policyAndExport`.
- Consume the above function in 2 endpoints:
  - "type of policy" POST
  - "type of policy save and go back" POST
  - In the near future, the function can be consumed by all POST routes in the "policy and export" section.

## Other
- Create `hasFormData` function to check if any fields have been submitted.
- Create `getValidFields` helper function that consumes form data and an `errorList`, and returns only valid fields.
- Create `getApplication` function that gets an application, also checking that the application has required child data.

## Moving forwards
The plan is to use the same approach for all other application pages that require "save and go back" functionality. For a different page/section, we'd just have to:

- Setup "save and go back" route/controller.
- Create `save-data` function to be consumed in all POST endpoints for the given page/section.
- Write e2e tests.


src/ui/server/helpers/has-form-data/index.test.ts
```diff
@@ -0,0 +1,26 @@
...
+      };
+
+      const result = hasFormData(mockFormData);
+
+      expect(result).toEqual(true);
+    });
+  });
+
+  describe('when a form does NOT have a field other than _csrf', () => {
+    it('should return false', () => {
```
abhi-markan, line 18: Do we need to add a coverage where the form field is empty or void?

```
  const mockFormData = {
        _csrf: '1234',
        mockField,
      };

File: src/ui/server/helpers/has-form-data/index.ts [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * hasFormData
+ * Checks if form data has values
+ * @param {Express.Request.body} Form data
+ * @returns {Boolean}
+ */
+const hasFormData = (formBody: RequestBody) => {
+  const { _csrf, ...formData } = formBody;
+
+  if (Object.keys(formData).length) {
```
abhi-markan, line 12: Might be safe to add this in case `formData` is undefined

```
if (!formDate) {
  return false;
}

**Review by abhi-markan — CHANGES_REQUESTED** [review]
Good PR, well done.
Few minor comments please.

File: src/ui/server/helpers/has-form-data/index.test.ts (hunk 1 above)
Comment (ttbarnes) [reply src/ui/server/helpers/has-form-data/index.test.ts:18]: Excellent point @abhi-markan yes need to improve this it thinks this is "a form with data"

```js
Object.keys({ test: '' }).length
Object.keys({ test: null }).length
Object.keys({ test: undefined }).length
```

...Which is technically true 😉 I'll add some checks to account for this

ttbarnes: Added a function to handle this, `stripEmptyFormFields` Great spot 👍

src/ui/server/helpers/get-valid-fields/index.ts
```diff
@@ -0,0 +1,26 @@
...
+    Object.keys(formData).forEach((fieldName) => {
+      if (!fieldsWithErrors.includes(fieldName)) {
+        validFields[fieldName] = formData[fieldName];
+      }
+    });
+
+    return validFields;
+  }
+
+  return {};
```
Zainzzkk, line 23: if there is no error list, shouldn't all the fieldNames be returned?

ttbarnes: Excellent spot! Will fix

ttbarnes: Added this in

Zainzzkk (dismissed):
LGTM

src/ui/server/helpers/has-form-data/index.ts
```diff
@@ -0,0 +1,53 @@
...
+export const stripEmptyFormFields = (formBody: RequestBody) => {
+  if (formBody) {
+    const fieldsWithValues = {};
+
+    const keys = Object.keys(formBody);
+
+    keys.forEach((key) => {
+      const value = formBody[key];
+
+      if (value !== '' && value !== null && value !== undefined) {
```
abhi-markan, line 18: Might be efficient to write `Boolean(value)`

src/ui/server/helpers/has-form-data/index.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param {Express.Request.body} Form data
+ * @returns {Boolean}
+ */
+const hasFormData = (formBody: RequestBody) => {
+  if (!formBody) {
+    return false;
+  }
+
+  const { _csrf, ...formData } = formBody;
+
```
abhi-markan, line 41: A check to ensure `formData` is not empty unless this will never be a case?

abhi-markan (changes requested):
Two minor comments to address please.

src/ui/server/helpers/has-form-data/index.ts
ttbarnes: Already have a check for this just above https://github.com/UK-Export-Finance/exip/pull/176/files/197ca2b410b89bcbcbd329b3da63db4485e10663#diff-9385d5383b0ef70440e9453d1771ee81991bc905c6b21b77c85a006e04dbf131R36

src/ui/server/helpers/has-form-data/index.ts
ttbarnes: Great shout, changed to this and allowing for false boolean as well

abhi-markan (approved):
LGTM

PR #177, opened 2022-12-07
fix(EMS-623): Insurance need to start route

Zainzzkk:
# Issue: 
* In the insurance section, need to start again link would take you to insurance start page
* This would wipe all the data stored on filled application

# Changes made: 
* Changed route to countries page on link
* Updated unit and e2e tests

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/eligibility/need-to-start-again.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/eligibility/need-to-start-again.spec.js
@@ -7,6 +7,7 @@ import { completeStartForm, completeCheckIfEligibleForm } from '../../../../supp
 const CONTENT_STRINGS = PAGES.NEED_TO_START_AGAIN_PAGE;
 
 const insuranceStartRoute = ROUTES.INSURANCE.START;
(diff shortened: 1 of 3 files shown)


e2e-tests/cypress/e2e/journeys/insurance/eligibility/need-to-start-again.spec.js
```diff
@@ -82,10 +83,10 @@ context('Insurance Eligibility - Need to start again exit page', () => {
   });
 
   describe('clicking the submit button', () => {
-    it(`should redirect to ${ROUTES.INSURANCE.START}`, () => {
+    it(`should redirect to ${ROUTES.INSURANCE.buyerCountryRoute}`, () => {
       submitButton().click();
 
-      cy.url().should('include', ROUTES.INSURANCE.START);
+      cy.url().should('include', buyerCountryRoute);
```
ttbarnes, line 89: This is my fault for not doing this, but can we use `eq` instead of `include`? e.g

```js
const expected = `${Cypress.config('baseUrl')}/the-url;

cy.url().should('eq', expected);
```

Zainzzkk: done

ttbarnes (approved):

PR #192, opened 2022-12-15
feat(EMS-506): Single policy type - requested start date

ttbarnes:


src/ui/server/helpers/create-date-from-numbers/index.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const expected = new Date(`${month} ${day} ${year}`);
+
+    expect(result).toEqual(expected);
+  });
+
+  describe('when day is not provided', () => {
+    it('should return null', () => {
+      // @ts-ignore
+      const result = createDateFromNumbers(undefined, month, year);
```
abhi-markan, line 19: Any specific reason for instructing TS complier to ignore following lines in this file?

abhi-markan (changes requested):

src/ui/server/helpers/create-date-from-numbers/index.test.ts
ttbarnes: The function expects day, month and year and we're testing that it returns null when one of these params is not provided. Typescript complains about field not provided so add this to ignore the TS error

abhi-markan (approved):

PR #193, opened 2022-12-15
feat(EMS-621 EMS-771): move insurance eligibility exit pages into insurance route

ttbarnes:
This PR moves the 'speak to UKEF EFM' and 'apply offline' exit pages in insurance eligibility, into a generic insurance route so that the URL makes sense in both contexts.

Previously:
- http://exip.com/insurance/eligibility/speak-to-a-UKEF-EFM
- http://exip.com/insurance/eligibility/apply-offline

Now:
- http://exip.com/insurance/speak-to-a-UKEF-EFM
- http://exip.com/insurance/apply-offline


Zainzzkk (approved):
LGTM

src/ui/server/controllers/insurance/speak-to-ukef-efm/index.ts
```diff
@@ -0,0 +1,18 @@
+import { PAGES } from '../../../content-strings';
+import { TEMPLATES } from '../../../constants';
+import { Request, Response } from '../../../../types';
+import corePageVariables from '../../../helpers/page-variables/core/insurance';
+
+const get = (req: Request, res: Response) => {
+  const EXIT_REASON = req.flash('exitReason');
+
```
abhi-markan, line 8: Any possiblity of `req` being empty, do we need to redirect user to a problem page?

abhi-markan (changes requested):
Good PR, one minor comment to address please.

ttbarnes:
Merging this in - concern that @abhi-markan raised is extremely unlikely if not impossible and the whole UI server would probably not be working.

src/ui/server/controllers/insurance/speak-to-ukef-efm/index.ts
ttbarnes: Marking as resolved so we can merge - extremely unlikely if not impossible and the whole UI server would probably not be working. Let me know if still concerned @abhi-markan can have a chat or look into what we could do with all controllers.

PR #195, opened 2022-12-19
feat(EMS-506) - Single contract policy - save data and render submitted values

ttbarnes:
This PR adds save and "save and go back" functionality to the single contract policy page and when revisiting the page, render previously submitted date values in the form fields.

## Summary of changes
- Add "save and go back" route and controller.
- Add save API call to the single contract policy POST controller.
- Changed single contract policy `PAGE_VARIABLES` to a function so that `referenceNumber` can be passed to generate a save and go back URL.
- Create a `mapSubmittedData` function to transform day/month/year fields into a date timestamp - this is what we need to send to the API/DB.
- Create a `getDateFieldsFromTimestamp` helper function that will be used for all date fields in the application. This function consumes a timestamp and field ID and returns an object with day, month and year fields, prefixed with the field ID.
- Create a generic `mapApplicationToFormFields`helper function that can be used for any form. The goal is to have a single place where, if we need to transform some data from the DB into a different structure, for UI component consumption, it can happen here. It currently just maps the field `requestedStartDate` with `getDateFieldsFromTimestamp`.

## Other improvements 
- Rename `createDateFromNumbers` helper function to `createTimestampFromNumbers`.
- Simplify some jest spy definitions in various policy and export unit tests.
- Simplify the type of policy controller by removing unnecessary top level try/catch wrapper.
- Improved test coverage for the type of policy controller.
- Restructure single contract policy E2E tests so that they belong to a `single-contract-policy` directory.


src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/index.ts
```diff
@@ -132,5 +147,20 @@ export const post = async (req: Request, res: Response) => {
     }
   }
 
-  return res.redirect(`${INSURANCE_ROOT}/${referenceNumber}${ROUTES.INSURANCE.POLICY_AND_EXPORTS.ABOUT_GOODS_OR_SERVICES}`);
+  try {
+    // save the application
+    const populatedData = mapSubmittedData(req.body);
+
```
abhi-markan, line 153: Can we add a void check for `populatedData` please.

abhi-markan (changes requested):

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/index.ts
ttbarnes: Hey @abhi-markan I don't think we need to, `mapSubmittedData` will always return an object?

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/single-contract-policy/save-and-back.spec.js
```diff
@@ -0,0 +1,171 @@
...
+
+    taskList.prepareApplication.tasks.policyTypeAndExports.link().click();
+
+    singlePolicyField.input().click();
+    submitButton().click();
+
+    getReferenceNumber().then((id) => {
+      referenceNumber = id;
+
+      const expected = `${Cypress.config('baseUrl')}${ROUTES.INSURANCE.ROOT}/${referenceNumber}${ROUTES.INSURANCE.POLICY_AND_EXPORTS.SINGLE_CONTRACT_POLICY}`;
```
Zainzzkk, line 53: can we destructure please? this one is big :P

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/index.ts
```diff
@@ -43,7 +53,8 @@ export const PAGE_VARIABLES = {
       ...FIELDS.CONTRACT_POLICY[POLICY_CURRENCY_CODE],
     },
   },
-};
+  SAVE_AND_BACK_URL: `${INSURANCE_ROOT}/${referenceNumber}${INSURANCE.POLICY_AND_EXPORTS.SINGLE_CONTRACT_POLICY_SAVE_AND_BACK}`,
```
Zainzzkk, line 56: can we destructure these at the top?

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/map-submitted-data/index.ts
```diff
@@ -0,0 +1,43 @@
...
+const mapSubmittedData = (formBody: RequestBody): object => {
+  const requestedStartDateFormIds = {
+    dayId: `${REQUESTED_START_DATE}-day`,
+    monthId: `${REQUESTED_START_DATE}-month`,
+    yearId: `${REQUESTED_START_DATE}-year`,
+  };
+
+  const { dayId, monthId, yearId } = requestedStartDateFormIds;
+
+  if (formBody) {
```
Zainzzkk, line 26: could we do all the ifs on one line? might look nicer as compared to nested ifs?

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/single-contract-policy/save-and-back.spec.js
ttbarnes: Good shout

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/map-submitted-data/index.ts
ttbarnes: I did think of this but we will only have to move it back to this in the next ticket as we'll need to add some more logic here

src/ui/server/helpers/date/get-date-fields-from-timestamp/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    const year = getYear(date);
+
+    return {
+      [`${fieldId}-day`]: day,
+      [`${fieldId}-month`]: month,
+      [`${fieldId}-year`]: year,
+    };
+  }
+
+  return null;
```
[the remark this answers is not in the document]
ttbarnes: I think either is good, but object makes more sense so it's a consistent type. Changed to return an empty object

abhi-markan (approved):

Zainzzkk (approved):
LGTM

PR #196, opened 2022-12-19
feat(EMS-508):  Single contract policy - add total value of contract field and improve form sanitisation functions

ttbarnes:
Waiting on this PR first https://github.com/UK-Export-Finance/exip/pull/195 so we can change the branch to `main-application`.

This PR adds the "total value of contract field" to the policy and export section.

## Summary of changes
- Add validation rules and tests for "total value of contract"
- Add "total value of contract" and "currency of the policy" fields to keystone schema.
- Update `mapApplicationToFormFields` helper function so that it returns other fields that don't require special logic/transformation.

## Other improvements
- Extract `stripEmptyFormFields` helper function into it's own directory.
- Update `sanitiseData` helper function to automatically:
  - Strip out any fields that have fields ending with `-day`, `-month` or `-year`.
  - Strip out any empty fields.
- Split up "single contract policy" E2E tests so all field validation tests are in their own file.


database/exip.sql
```diff
@@ -337,6 +337,8 @@ CREATE TABLE IF NOT EXISTS `PolicyAndExport` (
   `policyType` varchar(191) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   `application` varchar(191) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
 	`requestedStartDate` datetime(3) DEFAULT NULL,
+	`totalValueOfContract` int DEFAULT NULL,
```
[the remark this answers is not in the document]
abhi-markan: As I believe we do not expect negative values in this field?

src/api/.keystone/config.js
```diff
@@ -170,7 +170,14 @@ var lists = {
           { label: APPLICATION.POLICY_TYPE.MULTI, value: APPLICATION.POLICY_TYPE.MULTI }
         ]
       }),
-      requestedStartDate: (0, import_fields.timestamp)()
+      requestedStartDate: (0, import_fields.timestamp)(),
+      totalValueOfContract: (0, import_fields.integer)({
+        validation: {
+          min: 1,
+          max: 499999
```
abhi-markan, line 177: Is this `max`  value correct? Does this mean we can not have a total value of contract over £499,999?

src/api/schema.ts
```diff
@@ -149,6 +149,13 @@ export const lists = {
         ],
       }),
       requestedStartDate: timestamp(),
+      totalValueOfContract: integer({
+        validation: {
+          min: 1,
+          max: 499999,
```
abhi-markan, line 155: Same as above please.

src/ui/server/helpers/sanitise-data/index.ts
```diff
@@ -48,6 +49,20 @@ const sanitiseValue = (value: string | number | boolean) => {
...
 };
 
+/**
+ * isDayMonthYearField
+ * Checks if a field is a day, month or year field
+ * @param {String} Form field name
+ * @returns {Boolean}
+ */
+const isDayMonthYearField = (fieldName: string): boolean => {
+  if (fieldName.includes('-day') || fieldName.includes('-month') || fieldName.includes('-year')) {
```
abhi-markan, line 59: Do we need to add an empty `fieldName` check or is this never be the case?

abhi-markan (changes requested):

database/exip.sql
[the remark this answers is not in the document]
ttbarnes: That's correct, no negative values. We can add this to the SQL dump, but it's currently not possible to add `UNSIGNED` flag with keystone schema, so it will not get added to the DB if this dump is not used and it would get overridden when the schema changes.

Happy to add this in, but I don't see much point and think it should be the same as what keystone generates.

The UI will not call the API to save `totalValueOfContract` if there are validation errors e.g if there is a negative value

src/api/.keystone/config.js
ttbarnes: Yes that's right - business requirement.

I will move this to constants actually

src/ui/server/helpers/sanitise-data/index.test.ts
```diff
@@ -99,13 +99,51 @@ describe('server/helpers/sanitise-data', () => {
     });
   });
 
+  describe('isDayMonthYearField', () => {
+    describe('when a field name contains `-day`', () => {
+      it('should return true', () => {
```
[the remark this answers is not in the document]
ttbarnes: I've been thinking about this as we do these checks in other places. I don't think there is a need to because of typescript. If this function is attempted to be called from anywhere without `fieldName` string parameter, the code will not compile.

So I don't think there is any value in doing the param check and adding a unit test for this with `@ts-ignore` (for not providing a string param), because outside of tests, typescript will prevent the code from compiling.

What do you think?

src/ui/server/helpers/sanitise-data/index.ts
ttbarnes: Same comment as above

src/api/.keystone/config.js
ttbarnes: Moved to constants - marking as resolved

e2e-tests/content-strings/error-messages.js
```diff
@@ -92,6 +92,13 @@ export const ERROR_MESSAGES = {
             NOT_A_NUMBER: 'Enter a policy start date in the correct format - for example, 06 11 2023',
             BEFORE_EARLIEST: 'You cannot enter a policy start date in the past - enter a future date',
           },
+          [FIELD_IDS.INSURANCE.POLICY_AND_EXPORTS.CONTRACT_POLICY.SINGLE.TOTAL_CONTRACT_VALUE]: {
+            IS_EMPTY: 'Enter your contract value as a whole number - do not enter decimals',
+            NOT_A_NUMBER: 'Enter your contract value as a whole number - do not enter decimals',
+            NOT_A_WHOLE_NUMBER: 'Enter your contract value as a whole number - do not enter decimals',
+            BELOW_MINIMUM: 'Your contract value must be 1 or more',
+            ABOVE_MAXIMUM: 'The maximum the buyer will owe cannot be more than £499.999. If you need cover for more than this, fill in this form instead.',
```
Zainzzkk, line 100: not sure if this is github or needs linting?

ttbarnes: I think it's your screensize 😄 Looks fine if increase the browser window width

abhi-markan (approved):

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/validation/rules/total-contract-value.ts
```diff
@@ -0,0 +1,70 @@
...
+  INSURANCE: {
+    POLICY_AND_EXPORTS: {
+      CONTRACT_POLICY: {
+        SINGLE: { [FIELD_ID]: ERROR_MESSAGE },
+      },
+    },
+  },
+} = ERROR_MESSAGES;
+
+const MINIMUM = 1;
```
Zainzzkk, line 28: can we get these from constants please?

ttbarnes: Create a ticket here https://ukef-dtfs.atlassian.net/browse/EMS-783

Zainzzkk (approved):
LGTM

PR #199, opened 2022-12-20
fix(EMS-654): Generic combine validation rules function

Zainzzkk:
This PR adds a generic validation function that can consume multiple validation rule function results, combine them, and return as a single object. This helps to be consistent and removes repeated code.

## Issue
- Validation functions were using for loops
- Typings had `any`
- Repeated validation looping code for any form that had multiple rules.

## Changes
- Create a generic function, `combineValidationRules` that accepts an array of validation rule function results and combines into a single object of all errors.
- Changed to use forEach instead for all validation functions.
- Added test coverage

## Note
There is an exception where this function cannot currently be used - in the insurance companies house validation. This is because the form consumes validation errors from multiple places - UI validation, and API/companies house response. Maybe could be improved in the future, but not a blocker for this tech improvement.


abhi-markan (approved):

abhi-markan (changes requested):
Failing tests to address please.

src/ui/server/helpers/combine-validation-rules/index.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * combineValidationRules
+ * validates company details page response
+ * throws validation errors if any fields are not completed or incorrectly completed
+ * @param {Express.Request.body} responseBody containing an object with the company details response body
+ * @returns {object} object containing errors or blank object
+ */
+/* eslint-disable no-unused-vars, prettier/prettier */
+const combineValidationRules = (rules: Array<(formBody: object, errors: ValidationErrors) => ValidationErrors>, data: object) => {
+  /* eslint-enable no-unused-vars, prettier/prettier */
```
ttbarnes, line 12: Note that this eslint disable is required, unfortunately. WIthout this, TS complains about an unused variable. If we don't have the variable, things won't work.

This issue has been flagged in other repos/discussions. It occurs when you have a function type definition like above (array of functions: `Array<(formBody: object, errors: ValidationErrors)`).

src/ui/server/controllers/insurance/business/company-details/validation/company-details/index.ts
```diff
@@ -10,9 +10,11 @@ import { RequestBody } from '../../../../../../../types';
 const validation = (responseBody: RequestBody, errors: object) => {
   let updatedErrors = errors;
 
-  for (let i = 0; i < companyDetailsResponseRules.length; i += 1) {
-    updatedErrors = companyDetailsResponseRules[i](responseBody, updatedErrors);
-  }
+  /* eslint-disable no-unused-vars, prettier/prettier */
```
abhi-markan, line 13: Please remove `inline` es-lint rule.

src/ui/server/controllers/insurance/business/company-details/validation/company-details/index.ts
```diff
@@ -10,9 +10,11 @@ import { RequestBody } from '../../../../../../../types';
...
+  companyDetailsResponseRules.forEach((rule: (ruleData: object, ruleErrors: ValidationErrors) => ValidationErrors) => {
+    /* eslint-enable no-unused-vars, prettier/prettier */
```
abhi-markan, line 15: Please remove `inline` es-lint rule.

abhi-markan (changes requested):
Two minor comments to address please.

src/ui/server/controllers/insurance/business/company-details/validation/company-details/index.ts
ttbarnes: @abhi-markan We have to keep this in unfortunately, same reason as above https://github.com/UK-Export-Finance/exip/pull/199#discussion_r1055611819

But the good news is we now only have this inline eslint rule in 3 places rather than having it littered in every single validation function.

ttbarnes:
@abhi-markan Any ideas why DB queries have started failing in the API unit tests? https://github.com/UK-Export-Finance/exip/actions/runs/3767113914/jobs/6404425035 This PR hasn't touched the API at all, purely UI changes. It looks like the DB is not able to be queried

abhi-markan (approved):

PR #200, opened 2022-12-20
feat(EMS-509): Single contract policy - add credit period with buyer field

ttbarnes:
This PR adds the "credit period with buyer" field to the Single contract policy page.

## Summary of changes
- Add validation rules and tests for "credit period with buyer".
- Add "credit period with buyer" field to keystone schema.
- Ensure that the value is retained in the textarea when revisiting the page.


abhi-markan (approved):

abhi-markan (changes requested):
Various failing tests to address please.

abhi-markan (approved):

PR #201, opened 2022-12-22
feat(EMS-507): Single contract policy - add contract completion date

ttbarnes:
This PR adds the "contract completion date" to the Single contract policy page.

## Summary of changes

- Add contract completion date to the schema (API/DB, application types) and mock application used for unit tests.
- Add contract completion validation rules:
  - Cannot be empty.
  - Must be a number.
  - Must be in the future.
  - Cannot be the same as the requested start date.
  - Cannot be before the requested start date.
  - Cannot be past the maximum years of cover (via `MAX_COVER_PERIOD_YEARS` constant)
- Update `mapSubmittedData` to return a timestamp from day/month/year fields, for "contract completion date"
- Update `mapApplicationToFormFields`to return day/month/year fields from the saved timestamp.
- Update E2E tests to:
  - Test all possible completion date error/validation scenarios.
  - Test form submission with additional field.
  - Check that the field renders submitted values, when coming back to the page/form.

## Other improvements
- Improve/clean `mapSubmittedData` function, now that there are multiple date fields.
- Simplify the field ID by renaming from `expectedCompletionOfContractDate` to `contractCompletionDate`.


src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/validation/rules/contract-completion-date.test.ts
```diff
@@ -0,0 +1,196 @@
...
+      };
+
+      const result = contractCompletionDateRules(mockSubmittedData, mockErrors);
+
+      const expected = generateValidationErrors(CONTRACT_COMPLETION_DATE, ERROR_MESSAGE.IS_EMPTY, mockErrors);
+
+      expect(result).toEqual(expected);
+    });
+  });
+
```
abhi-markan, line 78: Can we please add a test case when neither of all three (day, month and year) are provided.

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/validation/rules/contract-completion-date.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  describe('when any day/month/year field is NOT a number', () => {
+    it('should return validation error', () => {
+      const mockSubmittedData = {
+        [`${CONTRACT_COMPLETION_DATE}-day`]: 'One',
+        [`${CONTRACT_COMPLETION_DATE}-month`]: 'Two',
+        [`${CONTRACT_COMPLETION_DATE}-year`]: 'Three',
```
abhi-markan, line 84: Can we add one with special characters and white space please.

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/validation/rules/contract-completion-date.ts
```diff
@@ -0,0 +1,122 @@
...
+ * Returns generateValidationErrors if there are any errors.
+ * @param {Express.Response.body} Express response body
+ * @param {Object} Errors object from previous validation errors
+ * @returns {Object} Validation errors
+ */
+const contractCompletionDateRules = (formBody: RequestBody, errors: object) => {
+  const updatedErrors = errors;
+  let errorMessage;
+
+  const inputValues = getDateInputValues(formBody);
```
abhi-markan, line 71: Do we need void `inputValues` check and/or ensure essential properties exists before proceeding further `inputValues.end.day`?

abhi-markan (changes requested):

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/validation/rules/contract-completion-date.test.ts
ttbarnes: We don't need to 😄 We have logic to check if all 3 fields are provided and have tests to make sure that if any single one of these fields is missing, an error is returned. 100% test coverage

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/validation/rules/contract-completion-date.ts
ttbarnes: I don't believe so. `getDateInputValues` will always return the object structure with `start.day` and `end.day` etc. If the values do not exist in the form body, the object structure will still be generated and the validation will still work.

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/single-contract-policy/single-contract-policy.spec.js
```diff
@@ -51,7 +51,7 @@ const {
         CREDIT_PERIOD_WITH_BUYER,
```
Zainzzkk, line 51: getDate import is not used.  Does not let me comment there (line 3)

ttbarnes: Sorry forgot to push, fixed

Zainzzkk (dismissed):
LGTM after tests pass

Zainzzkk (approved):
LGTM

abhi-markan (approved):

PR #204, opened 2022-12-23
feat(EMS-504): About goods or services - initial page/form setup

ttbarnes:
This PR adds the initial "About goods or services" page to the Type of policy section of an application.

## Summary of changes

- Setup GET route and controller.
- Setup template, Render the first form field.
- Add E2E test

## Other improvements

- Create a `completeAndSubmitSingleContractPolicyForm` cypress command


src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export const get = (req: Request, res: Response) => {
+  const { application } = res.locals;
+
+  if (!application) {
+    return res.redirect(ROUTES.PROBLEM_WITH_SERVICE);
+  }
+
+  const { referenceNumber } = req.params;
+  const refNumber = Number(referenceNumber);
```
abhi-markan, line 57: Do we need a void check for `refNumber`?

src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    return res.render(TEMPLATE, {
+      ...insuranceCorePageVariables({
+        PAGE_CONTENT_STRINGS: PAGES.INSURANCE.POLICY_AND_EXPORTS.ABOUT_GOODS_OR_SERVICES,
+        BACK_LINK: req.headers.referer,
+      }),
+      ...pageVariables(refNumber),
+      application,
+    });
+  } catch (err) {
+    console.error('Error getting currencies ', { err });
```
abhi-markan, line 69: Can we please update the error message to `Error rendering the application - goods or services page` please?

abhi-markan (changes requested):

src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.ts
ttbarnes: We don't - the controller is tied specifically to a route that will always have a reference number - we also have some middleware that is called before this function, that checks for a reference number (and fetches application), before getting to this controller.  https://github.com/UK-Export-Finance/exip/blob/main-application/src/ui/server/middleware/insurance/get-application/index.ts#L14

If the middleware does not have a reference number it'll redirect to a 404 page instead of hitting this controller. This applies to all `/insurance` routes that do _not_ include `/insurance/eligibility`

src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.ts
ttbarnes: Good spot sorry, copy and paste error. Removed this and the try/catch, we don't need these at all as there are no async calls.

Zainzzkk (approved):
LGTM after tests pass

Zainzzkk (commented):
1 failing test to fix

Zainzzkk (approved):
LGTM

abhi-markan (approved):

PR #205, opened 2022-12-23
chore(EMS-696): Update all controllers to have the same template/export pattern

ttbarnes:
This PR tidies up all controllers to have the same structure/pattern for template definitions and exports.

## All controllers now:

- Define a single `TEMPLATE` constant instead of defining it in multiple places for GET and POST.
- Export on the function level instead of exporting everything at the end of the file.


Zainzzkk (commented):
I am wondering if we should be more descriptive with 'TEMPLATE' incase we have multiple templates for any reason?

ttbarnes:
@Zainzzkk I can't forsee a page/controller having multiple templates. Any page differences should be a different route/URL and therefore would have their own controller? To be accessible and semantic.

Zainzzkk (approved):
LGTM

PR #209, opened 2022-12-29
feat(EMS-505): All sections - render submission deadline and reference number

ttbarnes:
This PR adds the submission deadline and reference number to the "all sections" page.

## Summary of changes

- Consume `mapApplicationToFormFields` in the "all sections" controller.
- Render a heading and paragraph for the submission deadline and reference number.
- Update `mapApplicationToFormFields` so that it transforms the submission deadline timestamp into a readable, formatted string via new helper function, `formatDate`.
- Add E2E tests.

## Other improvements
- Add `application` constants to the UI.
- In the `field-values` constants, consume single/multi contract policy names from the `application` constants (same as the API).
- Remove unused custom styles.
- Rename a companies house custom class name to be more specific.
- Change the date format in `mock-application` to match exactly how dates are saved.


src/ui/server/helpers/date/format-date/index.test.ts
```diff
@@ -0,0 +1,14 @@
+import { format } from 'date-fns';
+import formatDate from '.';
+
+describe('server/helpers/date/format-date', () => {
+  const mockTimestamp = new Date();
+
+  it('should return a formatted date', () => {
```
abhi-markan, line 7: A test when `undefined` or `null` argument is passed?

abhi-markan (changes requested):
LGTM, one minor comment to address please.

src/ui/server/helpers/date/format-date/index.test.ts
ttbarnes: @abhi-markan We don't need to 😃 the `timestamp` parameter is required - if there is an attempt to call this function without it, the code won't compile because of typescript. 

We could add a test for this, but doesn't seem to be value in doing this, would also have to add `ts-ignore` because it would flag that the parameter is required

Zainzzkk (approved):
LGTM

PR #212, opened 2022-12-30
feat(EMS-571): Policy and exports - Multiple contract policy page - cover start date validation

ttbarnes:
This PR adds requested cover start date validation to the multiple contract policy page.

## Summary of changes
- Add validation rules to the multiple contract policy POST controller.
- Create E2E helper functions to check the cover start date field and validation so that we can test the same validation in multiple places by running e.g
  - `requestedCoverStartDate.checkValidation.day.notProvided();`
- Move the cover start date field outside of `SINGLE` in field constants/content strings as it is now used in both single and multiple policy pages.


e2e-tests/content-strings/error-messages.js
```diff
@@ -119,6 +119,7 @@ export const ERROR_MESSAGES = {
             ABOVE_MAXIMUM: 'The maximum the buyer will owe cannot be more than £499.999. If you need cover for more than this, fill in this form instead.',
           },
         },
+        MULTIPLE: {},
```
Zainzzkk, line 122: can we remove this if empty?

ttbarnes: Good shout but TBH in the next PR (coming shortly), this will be populated shall we leave this in for now?

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/validation/multiple-contract-policy-validation-requested-start-date.spec.js
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  beforeEach(() => {
+    Cypress.Cookies.preserveOnce('_csrf');
+    Cypress.Cookies.preserveOnce('connect.sid');
+  });
+
+  describe('when day is not provided', () => {
+    it('should render a validation error', () => {
+      requestedCoverStartDate.checkValidation.day.notProvided();
```
Zainzzkk, line 48: can this be dereferenced for all occurrences?

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/single-contract-policy/validation/single-contract-policy-validation-requested-start-date.spec.js
```diff
@@ -69,137 +43,45 @@ context('Insurance - Policy and exports - Single contract policy page - form val
...
-      checkText(
-        partials.errorSummaryListItems().eq(0),
-        CONTRACT_ERROR_MESSAGES.SINGLE[REQUESTED_START_DATE].IS_EMPTY,
-      );
-
-      checkText(
-        field.errorMessage(),
-        `Error: ${CONTRACT_ERROR_MESSAGES.SINGLE[REQUESTED_START_DATE].IS_EMPTY}`,
-      );
+      requestedCoverStartDate.checkValidation.day.notProvided();
```
Zainzzkk, line 48: can this be dereferenced for all occurrences

src/ui/server/controllers/insurance/policy-and-export/multiple-contract-policy/index.ts
```diff
@@ -104,3 +107,68 @@ export const get = async (req: Request, res: Response) => {
...
+      const mappedCurrencies = mapCurrencies(currencies);
+
+      return res.render(TEMPLATE, {
+        ...insuranceCorePageVariables({
+          PAGE_CONTENT_STRINGS: PAGES.INSURANCE.POLICY_AND_EXPORTS.MULTIPLE_CONTRACT_POLICY,
+          BACK_LINK: req.headers.referer,
+        }),
+        ...pageVariables(refNumber),
+        application: mapApplicationToFormFields(application),
+        submittedValues: req.body,
```
Zainzzkk, line 147: in the future, I think we should santise the req.body before passing it back? future PR

e2e-tests/content-strings/error-messages.js
ttbarnes: Removed it

src/ui/server/controllers/insurance/policy-and-export/multiple-contract-policy/index.ts
ttbarnes: Interesting, good idea - i'm wondering what value we'd get from that though - because when rendered, it would be transformed back into what it was before it was posted? We'd also need a different (or adapt existing) "sanitise data" function because in this scenario, we wouldn't want to for example, remove commas from a number

Zainzzkk (approved):
LGTM

PR #214, opened 2022-12-30
chore(EMS-786): update README

ttbarnes:
## Summary of changes

- Add links to the tech stack list
- Add a section briefly explaining what keystone does, including:
  - What keystone generates when we add something to the keystone schema
  - What keystone hooks are
  - How we customise the generated graphql schema
  - Why we use keystone
  - How and when the UI calls the API
- Remove the deployment section as this has changed and is currently undergoing discussions of what we should/shouldn't put in the README.

Also, improved the documentation/description for the policy and export's "save data" function.

Quick link of the updated README in generated format: https://github.com/UK-Export-Finance/exip/blob/f393d29dae0c5885591d400c263e753a8c5d70d7/README.md


abhi-markan (approved):

Zainzzkk (approved):
LGTM

PR #219, opened 2023-01-04
feat(EMS-697): Your buyer - Buyer’s company - country field

vensurag:


src/ui/server/constants/templates/index.ts
```diff
@@ -1,12 +1,14 @@
...
 import { SHARED_PAGES } from './shared-pages';
+import { YOUR_BUYER_DETAILS_TEMPLATES } from './insurance/your-buyer-details';
 
 export const TEMPLATES = {
   COOKIES: 'cookies.njk',
   PROBLEM_WITH_SERVICE: 'problem-with-service.njk',
   CANNOT_APPLY: 'cannot-apply.njk',
   QUOTE: QUOTE_TEMPLATES,
   INSURANCE: INSURANCE_TEMPLATES,
+  YOUR_BUYER: YOUR_BUYER_DETAILS_TEMPLATES,
```
Zainzzkk, line 12: @ttbarnes do you think we should put these in 'INSURANCE' routes?

src/ui/server/controllers/insurance/your-buyer/your-buyer-details/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  let mappedCountries;
+
+  if (countryValue) {
+    mappedCountries = mapCountries(countries, countryValue.isoCode);
+  } else {
+    mappedCountries = mapCountries(countries);
+  }
+  // // eslint-disable-next-line no-console
+  // console.log(mappedCountries);
```
Zainzzkk, line 43: can this be removed please

src/ui/templates/insurance/your-buyer/your-buyer-details.njk
```diff
@@ -0,0 +1,91 @@
...
+      <div class="govuk-grid-column-one-half-from-desktop">
+
+        {% set class = "govuk-form-group" %}
+
+        {% if validationErrors.errorList[FIELD_ID] %}
+          {% set class = "govuk-form-group govuk-form-group--error" %}
+        {% endif %}
+
+
+{{ govukSelect({
```
Zainzzkk, line 46: indentation fix please

src/ui/templates/insurance/your-buyer/your-buyer-details.njk
```diff
@@ -0,0 +1,91 @@
...
+  id: CONTENT_STRINGS.BUYER_COUNTRY,
+  name: CONTENT_STRINGS.BUYER_COUNTRY,
+  label: {
+    text: CONTENT_STRINGS.COUNTRY_LABEL
+  },
+  items: countries
+}) }}
+      </div>
+
+            <div class="govuk-grid-column-three-quarters-from-desktop">
```
Zainzzkk, line 56: same as above

src/ui/server/routes/insurance/your-buyer-details/index.ts
```diff
@@ -0,0 +1,11 @@
...
+// import { YOUR_BUYER } from '../../../constants/routes/insurance/your-buyer';
+// // import { ROUTES } from '../../../constants';
+// import { get as getaYourBuyerOrganisation } from '../../../controllers/insurance/your-buyer/buyer-oganisation';
+
+// // @ts-ignore
+// const yourBuyerRouter = express.Router();
+
+// yourBuyerRouter.get(YOUR_BUYER.YOUR_BUYER_DETAILS, getaYourBuyerOrganisation);
+
+// export default yourBuyerRouter;
```
Zainzzkk, line 11: is there a reason these are commented out? And @ttbarnes should we follow the same thing for 'EXPORTER_BUSINESS' where routes are in 'YOUR_BUYER' file and then dereferenced in src/ui/server/routes/insurance/index.ts

src/ui/server/helpers/mappings/map-countries.ts
```diff
@@ -82,6 +82,7 @@ export const filterCisCountries = (countries: Array<CisCountry>) => countries.fi
 export const mapCountry = (country: CisCountry, selectedIsoCode?: string): Country => {
   const mapped = {
     name: country.marketName,
+    text: country.marketName,
```
Zainzzkk, line 85: can we use name instead where 'text' is being referred to? as it seems like duplication otherwise? Ignore me if I am misunderstanding

src/ui/server/constants/routes/insurance/index.ts
```diff
@@ -26,6 +26,7 @@ export const INSURANCE_ROUTES = {
     NEED_TO_START_AGAIN: `${INSURANCE_ROOT}${ELIGIBILITY_ROOT}/need-to-start-again`,
   },
   APPLY_OFFLINE: `${INSURANCE_ROOT}/apply-using-our-form`,
+  YOUR_BUYER_DETAILS: `${INSURANCE_ROOT}/your-buyer-details`,
```
ttbarnes, line 29: Can we extract this into it's own object/file like `EXPORTER_BUSINESS` and `POLICY_AND_EXPORTS` (line 33/34) please? As the buyer section grows they'll be more "your buyer" routes

src/ui/server/constants/templates/index.ts
ttbarnes: 100% yes, it's part of insurance

src/ui/server/content-strings/pages/index.ts
```diff
@@ -126,4 +130,5 @@ export const PAGES = {
   PROBLEM_WITH_SERVICE_PAGE,
   QUOTE: QUOTE_PAGES,
   INSURANCE: INSURANCE_PAGES,
+  YOUR_BUYER: YOUR_BUYER_DETAILS_PAGE,
```
ttbarnes, line 133: Can we move this into insurance please and export like the other insurance sections, e.g

```js
import * as EXPORTER_BUSINESS from './exporter-business';
```

src/ui/server/controllers/insurance/your-buyer/your-buyer-details/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  PAGE_CONTENT_STRINGS: PAGES.YOUR_BUYER,
+};
+
+export const get = async (req: Request, res: Response) => {
+  if (!req.session.submittedData || !req.session.submittedData.insuranceEligibility) {
+    req.session.submittedData = {
+      ...req.session.submittedData,
+      insuranceEligibility: {},
+    };
+  }
```
ttbarnes, line 21: Don't need these checks, this is only required for eligibility routes, not insurance

ttbarnes: Instead we need to check the application exists - full example here https://github.com/UK-Export-Finance/exip/blob/main-application/src/ui/server/controllers/insurance/policy-and-export/multiple-contract-policy/index.ts#L79

src/ui/server/controllers/insurance/your-buyer/your-buyer-details/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  if (!countries || !countries.length) {
+    return res.redirect(ROUTES.PROBLEM_WITH_SERVICE);
+  }
+
+  let countryValue;
+
+  if (req.session.submittedData && req.session.submittedData.insuranceEligibility[FIELD_IDS.BUYER_COUNTRY]) {
+    countryValue = req.session.submittedData.insuranceEligibility[FIELD_IDS.BUYER_COUNTRY];
+  }
```
ttbarnes, line 33: This won't work for the full application, need to get the country value from `res.locals.application` instead.

This will be empty until we are doing data saving/sending to API in the POST

src/ui/server/controllers/insurance/your-buyer/your-buyer-details/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  if (countryValue) {
+    mappedCountries = mapCountries(countries, countryValue.isoCode);
+  } else {
+    mappedCountries = mapCountries(countries);
+  }
+  // // eslint-disable-next-line no-console
+  // console.log(mappedCountries);
+  return res.render(TEMPLATES.YOUR_BUYER.BUYER_COMPANY_DETAILS, {
+    ...singleInputPageVariables({
```
ttbarnes, line 45: `insuranceCorePageVariables`

src/ui/server/helpers/mappings/map-countries.ts
ttbarnes: We need to revert these changes to `map-countries` and the country type change (in `types/country.d.ts`), this is my fault for not making this clear. 

`map-countries` was written a while ago and is used for mapping countries we get from an external API. it's not used to map data to the structure we need for the GOV select component, like `map-currencies`.

I will create a seperate PR to rename `map-countries` so this is clearer, then we can create a new `map-countries` like `map-currencies`. This is required to pass the A/C of this ticket - without a new mapping, the validation won't work because something will always be selected.

src/ui/server/routes/insurance/your-buyer-details/index.ts
ttbarnes: Yes exactly 😄 we follow the URL structure of each page, for every section - e.g these URLs defined in the design:

- exip.com/insurance/policy-and-exports/type-of-policy
- exip.com/insurance/your-business/company-details
- exip.com/insurance/your-buyer/buyer-company-organisation


Should be structured like this in the codebase:

```js
src/ui/server/routes/insurance/policy-and-exports/type-of-policy
src/ui/server/routes/insurance/your-business/company-details
src/ui/server/routes/insurance/your-buyer/buyer-company-organisation
```

Not like this:
```js
src/ui/server/routes/insurance/buyer-company-organisation
src/ui/server/routes/insurance/your-buyer-details
```
Because the "organisation" route has no context of what section it belongs to.

Same applies to all sepects e.g controllers, field definitions, e2e tests etc.

The template location of `your-buyer-details.njk`is perfect 👍

src/ui/templates/insurance/your-buyer/your-buyer-details.njk
```diff
@@ -0,0 +1,91 @@
...
+    {{ govukErrorSummary({
+      titleText: "There is a problem",
+      errorList: validationErrors.summary
+    }) }}
+  {% endif %}
+
+  <form method="POST" novalidate>
+
+    <input type="hidden" name="_csrf" value="{{ csrfToken }}">
+    <input type="hidden" name="{{ HIDDEN_FIELD_ID }}" id="{{ HIDDEN_FIELD_ID }}" value="{{ submittedValues[HIDDEN_FIELD_ID].name }}">
```
ttbarnes, line 34: No need for `HIDDEN_FIELD_ID` - seems this is a copy and paste from the buyer-country page in eligibility - that uses a search box whereas here we have simpler select drop down

src/ui/templates/insurance/your-buyer/your-buyer-details.njk
```diff
@@ -0,0 +1,91 @@
...
+    <input type="hidden" name="{{ HIDDEN_FIELD_ID }}" id="{{ HIDDEN_FIELD_ID }}" value="{{ submittedValues[HIDDEN_FIELD_ID].name }}">   
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-half-from-desktop">
+
+        {% set class = "govuk-form-group" %}
+
+        {% if validationErrors.errorList[FIELD_ID] %}
+          {% set class = "govuk-form-group govuk-form-group--error" %}
+        {% endif %}
```
ttbarnes, line 43: Same as above we don't need this, it can just be in the select component e.g


```js
{{ govukSelect({
  ...
  errorMessage: validationErrors.errorList[FIELDS.COUNTRY.ID] and {
  text: validationErrors.errorList[FIELDS.COUNTRY.ID].text,
    attributes: {
      "data-cy": FIELDS.COUNTRY.ID + "-error-message"
    }
  }
}) }}
```

Can we also use a `FIELDS` structure because there are (or will be) multiple fields in this page.

src/ui/templates/insurance/your-buyer/your-buyer-details.njk
```diff
@@ -0,0 +1,91 @@
...
+
+  <script src="/assets/js/accessibleAutocomplete.js" type="text/javascript"></script>
+
+  <script type="text/javascript">
+    var element = document.getElementById('{{ FIELD_ID }}');
+    accessibleAutocomplete.enhanceSelectElement({
+      selectElement: element,
+      defaultValue: ''
+    });
+  </script>
```
ttbarnes, line 89: No need for the scripts, this is for eligibility buyer country which is different requirement

ttbarnes (changes requested):

PR #222, opened 2023-01-05
feat(EMS-576): Policy and exports - Multiple contract policy - currency field

ttbarnes:
This PR adds validation to the "currency" field in the multiple contract policy page.

## Changes

- Move the currency code validation function used in "single contract policy" into the shared validation directory.
- Consume the currency code validation in "multiple contract policy" validation.
- Add/update tests.
- Also ensured that the "months of cover" and currency code is rendered correctly in the form if submitted.

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/validation/multiple-contract-policy-validation-policy-currency-code.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/validation/multiple-contract-policy-validation-policy-currency-code.spec.js
@@ -1,5 +1,5 @@
 import { submitButton } from '../../../../../pages/shared';
-import { typeOfPolicyPage } from '../../../../../pages/insurance/policy-and-export';
+import { typeOfPolicyPage, multipleContractPolicyPage } from '../../../../../pages/insurance/policy-and-export';
 import partials from '../../../../../partials';
 import { ERROR_MESSAGES } from '../../../../../../../content-strings';
 import { FIELD_IDS, ROUTES } from '../../../../../../../constants';
@@ -9,7 +9,7 @@ import checkText from '../../../../../helpers/check-text';
 const { taskList } = partials.insurancePartials;
 
 const multiplePolicyFieldId = FIELD_IDS.INSURANCE.POLICY_AND_EXPORTS.POLICY_TYPE;
-const multiplePolicyField = typeOfPolicyPage[multiplePolicyFieldId].multiple;
+const multiplePolicyField = typeOfPolicyPage[multiplePolicyFieldId].single;
 
 const { INSURANCE } = ROUTES;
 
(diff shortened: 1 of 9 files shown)


e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/validation/multiple-contract-policy-validation-policy-currency-code.spec.js
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  beforeEach(() => {
+    Cypress.Cookies.preserveOnce('_csrf');
+    Cypress.Cookies.preserveOnce('connect.sid');
+  });
+
+  const field = singleContractPolicyPage[POLICY_CURRENCY_CODE];
+
+  describe('when policy currency code is not provided', () => {
```
Zainzzkk, line 67: just wondering since this test is so short, can this be combined with any others? would save the time of creating a new application for this test

src/ui/server/controllers/insurance/policy-and-export/multiple-contract-policy/index.ts
```diff
@@ -93,7 +93,13 @@ export const get = async (req: Request, res: Response) => {
       return res.redirect(ROUTES.PROBLEM_WITH_SERVICE);
     }
 
-    const mappedCurrencies = mapCurrencies(currencies);
+    let mappedCurrencies;
+
+    if (application.policyAndExport[POLICY_CURRENCY_CODE]) {
```
Zainzzkk, line 98: can we use objectHasProperty for this?

Zainzzkk (commented):
linting fail

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/validation/multiple-contract-policy-validation-policy-currency-code.spec.js
ttbarnes: I like the thinking - could have a chat/session about it?

Thing is, for these kind of pages, where there are a lot of form fields and validation rules - one single field has a validation E2E test that is 150 lines long. Others like this one have a lot less.

I think if we were to combine all of these into a single "validation rules" E2E test - even if we extract the checks into seperate files/functions - I think the test would be too big.

The application creation doesn't take very long right now, but it's great to think about this and ultimately the time the entire E2E suite takes to run. When we add login, the time for E2E suite will increase. At that time I think we should have a think about anything we can do to decerease E2E time.

But I dont think this is an immediate issue we need to tackle. What do you think?

ttbarnes: One thing we could do is instead of going through the eligibility and "create application", maybe we could add an API call to skip that part. That'd shave off 3-5 seconds for each application test (if I correctly remember how long it takes)

Zainzzkk (approved):
LGTM

PR #224, opened 2023-01-05
feat(EMS-570): Policy and export - multiple contract policy - save and go back. Improve save data approach

ttbarnes:
This PR adds "save and go back" functionality to the "multiple contract policy" page.

## Summary of changes

- Add "save and go back" route and POST controller.
- Add E2E test.
- Add documentation RE submit button requirements, update "what happens when a form is submitted" section. Here is the markdown generated section: https://github.com/UK-Export-Finance/exip/blob/e0b3c3552c0a1f9926b38ad6ffcc0a3385a97d9c/README.md#requirements-for-submitting-a-form

## Technical improvements

- Create a `mapAndSave.policyAndExport` function. This can be used in all POST methods for any page in the policy and export section. This function:
  - Maps any form data (if required). E.g, transform day/month/year fields into a timestamp.
  - Calls `save.policyAndExport` with the mapped data and any passed validation errors.  This save function then sanitises the data and calls the API.
- Update all POST controllers in policy and export to call `mapAndSave.policyAndExport` instead of mapping and passing the mapping to `save.policyAndExport`. DRY
- Update all policy and export E2E tests to have a single declaration of the relevant task.

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/save-and-back.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/save-and-back.spec.js
@@ -1,0 +1,182 @@
+import {
+  add,
+  getDate,
+  getMonth,
+  getYear,
+  sub,
+} from 'date-fns';
+import { submitButton, saveAndBackButton } from '../../../../pages/shared';
+import { typeOfPolicyPage, multipleContractPolicyPage } from '../../../../pages/insurance/policy-and-export';
+import partials from '../../../../partials';
+import { TASKS } from '../../../../../../content-strings';
+import { ROUTES, FIELD_IDS } from '../../../../../../constants';
+import getReferenceNumber from '../../../../helpers/get-reference-number';
+import application from '../../../../../fixtures/application';
+
+const { taskList } = partials.insurancePartials;
+
+const multiplePolicyFieldId = FIELD_IDS.INSURANCE.POLICY_AND_EXPORTS.POLICY_TYPE;
+const multiplePolicyField = typeOfPolicyPage[multiplePolicyFieldId].multiple;
+
+const {
+  INSURANCE: {
+    ROOT: INSURANCE_ROOT,
+    ALL_SECTIONS,
+    POLICY_AND_EXPORTS: {
+      MULTIPLE_CONTRACT_POLICY,
+    },
+  },
+} = ROUTES;
+
+const {
+  INSURANCE: {
+    POLICY_AND_EXPORTS: {
+      CONTRACT_POLICY: {
+        MULTIPLE: { TOTAL_SALES_TO_BUYER },
+      },
+    },
+  },
+} = FIELD_IDS;
+
+const {
+  INSURANCE: {
+    POLICY_AND_EXPORTS: {
+      CONTRACT_POLICY: {
+        REQUESTED_START_DATE,
+      },
+    },
+  },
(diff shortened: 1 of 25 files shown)


e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/save-and-back.spec.js
```diff
@@ -0,0 +1,163 @@
...
+    it(`should redirect to ${ALL_SECTIONS}`, () => {
+      saveAndBackButton().click();
+
+      const expected = `${Cypress.config('baseUrl')}${INSURANCE_ROOT}/${referenceNumber}${ALL_SECTIONS}`;
+
+      cy.url().should('eq', expected);
+    });
+
+    it('should retain the `type of policy and exports` task status as `in progress`', () => {
+      const task = taskList.prepareApplication.tasks.policyTypeAndExports;
```
Zainzzkk, line 106: can this be declared at the top so it can be used multiple times?

src/ui/server/controllers/insurance/policy-and-export/map-and-save/index.ts
```diff
@@ -0,0 +1,43 @@
+import hasFormData from '../../../../helpers/has-form-data';
+import mapSubmittedData from '../map-submitted-data';
+import save from '../save-data';
+import { Application, RequestBody, ValidationErrors } from '../../../../../types';
+
+/**
+ * mapAndSave
+ * Map and save any valid Single contract policy fields
+ * @param {Express.Request.body} Express request body
+ * @param {Object} Validation errors
```
Zainzzkk, line 10: missing application here

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/save-and-back/index.ts
```diff
@@ -29,19 +27,13 @@ export const post = async (req: Request, res: Response) => {
...
-
-      let saveResponse;
+      const saved = await mapAndSave.policyAndExport(req.body, application, validationErrors);
 
-      if (validationErrors) {
-        saveResponse = await save.policyAndExport(application, populatedData, validationErrors.errorList);
-      } else {
-        saveResponse = await save.policyAndExport(application, populatedData);
+      if (saved) {
+        return res.redirect(`${INSURANCE_ROOT}/${referenceNumber}${ROUTES.INSURANCE.ALL_SECTIONS}`);
```
Zainzzkk, line 33: can we add this route as a const since it's being used twice?

Zainzzkk: And all sections could be dereferenced on line 7

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/save-and-back.spec.js
ttbarnes: Excellent idea - i've updated all policy and export E2E tests to have a single decleration of this

src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/save-and-back/index.ts
ttbarnes: We can't declare it outside of the function as it needs the reference number param. Improved the dereferencing - think it's nice now

Zainzzkk (approved):
LGTM

PR #230, opened 2023-01-09
feat(EMS-808): API/DB structure for exporter company

ttbarnes:
This PR adds the ability to get and save exporter company details, for the "Company details" page.

## Summary of changes

- Add `exporterCompany` and `exporterCompanyAddress` to the keystone schema/DB/API. These are available to be queried from the top level of the application.
- Add placeholder `ExporterCompanySicCode` and `ExporterBusiness` to the schema, ready for further development.
- Create a new custom resolver in keystone schema to update an exporter company _and_ address.
  - The requirement we have is to update an exporter's company for an application, as well as the company address. We want to keep these pieces of data seperate to have a clean database structure.
  - For various reason, it is troublesome/not possible or results in poor code quality, to update 2 tables at the same time with one query. Alternative options are to either make 2 api calls from the UI (bad), or have a custom resolver that can handle the request and data saving as we need.
- Update application keystone schema so that when an application is created:
  - `exporterCompany` and `exporterCompanyAddress` rows are created and associated with the application. This avoids us having to create _and_ update further down the line.
- Add GQL mutation to the UI, for updating an exporter company.
- Update the "get application GQL query in the UI to return exporter company fields.

## Other improvements
- Rename trading name and trading address field IDs to reflect it being a boolean/yes/no question.
- Rename companies house naming in the custom schema to avoid confusion between `ExporterCompanyAddress`.

### DIFF
--- a/src/ui/server/api/keystone/application/index.ts
+++ b/src/ui/server/api/keystone/application/index.ts
@@ -120,7 +120,6 @@ const application = {
         throw new Error(`Updating application policy and export ${err}`);
       }
     },
-    // TODO: align error message
     exporterCompany: async (companyId: string, companyAddressId: string, update: object) => {
       try {
         console.info('Updating exporter company');
--- a/src/ui/server/api/keystone/application/index.ts
+++ b/src/ui/server/api/keystone/application/index.ts
@@ -6,6 +6,7 @@ import countries from '../countries';
 import createApplicationMutation from '../../../graphql/mutations/create-application';
 import getApplicationQuery from '../../../graphql/queries/application';
 import updateApplicationPolicyAndExportMutation from '../../../graphql/mutations/update-application/policy-and-export';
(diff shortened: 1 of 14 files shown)


src/ui/server/api/keystone/application/index.ts
```diff
@@ -6,6 +6,7 @@ import countries from '../countries';
 import createApplicationMutation from '../../../graphql/mutations/create-application';
 import getApplicationQuery from '../../../graphql/queries/application';
 import updateApplicationPolicyAndExportMutation from '../../../graphql/mutations/update-application/policy-and-export';
+import updateExporterCompanytMutation from '../../../graphql/mutations/update-application/exporter-company';
```
Zainzzkk, line 9: typo

Zainzzkk (approved):
LGTM

PR #231, opened 2023-01-10
feat(EMS-609): Improve country and currency mapping functions

ttbarnes:
This PR adds a `mapSelectOption` function that can be used in any select options mapping functions, for example in currencies and countries. EXIP has many instances of currency and country mappings and there will be more. As the application grows, we need to ensure such mapping functions are DRY.

## Summary of changes
- Create `mapSelectOption` function for mapping an option into the required GOV design structure.
- Rename `mapCountries` to `mapCisCountries`. This function only maps the API response from the CIS (country information system) API.
- Create a `mapCountries` function to map countries, regardless of where the countries are obtained from (our API, external API).

## Other improvements

- Update mock countries.
- Update mock currencies to include all supported currencies.
- Ensure that we are using the same mock countries and currencies from the test mocks, throughout the codebase.
- Remove `mapCountriesSelect` function and instead use the new and improved `mapCountries` function.
- Update application "your buyer" controller to call keystone countries API instead of the CIS API.
- Update quote "your buyer" controller to use a try/catch. Improve test coverage by adding promise rejection tests.


src/ui/server/controllers/quote/buyer-country/index.ts
```diff
@@ -71,9 +71,9 @@ export const get = async (req: Request, res: Response) => {
   let mappedCountries;
 
   if (countryValue) {
-    mappedCountries = mapCountries(countries, countryValue.isoCode);
+    mappedCountries = mapCisCountries(countries, countryValue.isoCode);
```
Zainzzkk, line 74: can we try catch the api call on line 59?

ttbarnes: Great spot - this controller was created ages ago (first page!) I've added a try/catch and improved test coverage by adding a unit test for when the API call/promise is rejected

Zainzzkk (approved):
LGTM

PR #232, opened 2023-01-10
feat(EMS-609): Policy and export - About goods or services - final destination field

ttbarnes:
This PR adds the "final destination" field to the About goods or services page.

## Summary of changes

- Add `finalDestinationCountryCode` field to the schemas, GQL application get query and DB dump.
- Add final destination validation rule for "is empty".
- Get countries from keystone API when visiting the page or on POST when there are validation errors.
- Add countries fixtures to E2E tests.
- Removed about goods or services "description"  field validation E2E test. This is already tested in the main validation E2E test, no need to be duplicated.


src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.test.ts
```diff
@@ -83,27 +95,106 @@ describe('controllers/insurance/policy-and-export/about-goods-or-services', () =
...
+        await get(req, res);
 
         expect(res.redirect).toHaveBeenCalledWith(ROUTES.PROBLEM_WITH_SERVICE);
       });
     });
+
+    describe('api error handling', () => {
+      describe('when there are no countries returned from the API', () => {
+        beforeEach(() => {
+          // @ts-ignore
```
Zainzzkk, line 149: do we need this?

src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.ts
```diff
@@ -59,14 +62,35 @@ export const get = (req: Request, res: Response) => {
...
-  return res.render(TEMPLATE, {
-    ...insuranceCorePageVariables({
-      PAGE_CONTENT_STRINGS: PAGES.INSURANCE.POLICY_AND_EXPORTS.ABOUT_GOODS_OR_SERVICES,
-      BACK_LINK: req.headers.referer,
-    }),
-    ...pageVariables(refNumber),
-    application,
-  });
+  try {
+    const countries = await api.keystone.countries.getAll();
```
Zainzzkk, line 66: should the api call have its own try catch?

src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.test.ts
ttbarnes: Unfortunately yes - have this in a few places. We are testing here that if the API call is successful, but returns an empty body, we redirect to "problem with service" page.

Typescript does not allow an empty `Promise.resolve()` but here we are stubbing this to mock an empty body from the API

Happy to remove these tests, but feels worth having? We had issues like this with the CIS API before so introduced this test, have been using for all API calls since.

Zainzzkk (approved):
LGTM

PR #236, opened 2023-01-11
feat(EMS-607): Policy and exports - callMapAndSave function

ttbarnes:
This PR adds a new function to the policy and export controllers directory - `callMapAndSave`. This is to reduce repeated logic in the "save and back" controllers.

## Summary of changes
- Extract repeated logic and unit tests from "save and back" controllers and move into a single function.
- In the "type of policy" validation, return an empty object instead of null. This was causing issues with the improved codebase changes. Add a `objectHasValues` check to the controller so it works as before.

### DIFF
--- a/src/ui/server/controllers/insurance/policy-and-export/call-map-and-save/index.ts
+++ b/src/ui/server/controllers/insurance/policy-and-export/call-map-and-save/index.ts
@@ -1,0 +1,24 @@
+import mapAndSave from '../map-and-save';
+import { Application, RequestBody, ValidationErrors } from '../../../../../types';
+
+const callMapAndSave = async (formData: RequestBody, application: Application, validationErrors?: ValidationErrors) => {
+  try {
+    let saveResponse;
+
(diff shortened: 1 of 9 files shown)


src/ui/server/controllers/insurance/policy-and-export/call-map-and-save/index.ts
```diff
@@ -0,0 +1,24 @@
+import mapAndSave from '../map-and-save';
+import { Application, RequestBody, ValidationErrors } from '../../../../../types';
+
+const callMapAndSave = async (formData: RequestBody, application: Application, validationErrors?: ValidationErrors) => {
```
Zainzzkk, line 4: documentation please

ttbarnes: Whops sorry, added

Zainzzkk (approved):
LGTM

PR #242, opened 2023-01-16
chore(EMS-606): Improve summary list generation functions

ttbarnes:
This PR improves all summary list generation functions for the Quote tool answers list, Quote list and companies house summary list. As the application grows and we need to add more summary lists, we need to make sure we are DRY.

## Summary of changes 

- Simplify data content mapping functions to return just a text string instead of an object with `.text`.
- Create a generic `generateSummaryListRows` for all summary lists. DRY.
- Create a new mapping function `generateSummaryListItemData` to reduce the amount of lines for each summary list row in each generation function. This function maps a field with the submitted value and also gets the field title (or "summary title" if it exists - design differs for some fields). The list item data function is then ready to be mapped into the (harder to map over) format required for GOV summary lists.
- Update alll summary list generation functions to use `generateSummaryListItemData`.

## Other improvements
- Create an empty string dash content string, update all usages.
- Improve documentation for companies house summary list generation functions.


src/ui/server/helpers/summary-lists/generate-field-group-item/index.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param {String} Field ID
+ * @param {String} Custom field value
+ * @returns {String} Field value
+ */
+export const getSummaryListItemDataValue = (data: object, fieldId: string, customValue?: string): string => {
+  if (customValue) {
+    return customValue;
+  }
+
+  if (data[fieldId]) {
```
Zainzzkk, line 19: objectHasProperty instead?

ttbarnes: Great shout

Zainzzkk (dismissed):
LGTM

Zainzzkk (approved):
LGTM

PR #243, opened 2023-01-16
feat(EMS-613-637): Company details - populate fields

Zainzzkk:
# Changes made: 
* Populated fields on company details page in get controller if they exist
* Added helper to go map through application and populate companies house summary list
* Added tests and added to e2e test


e2e-tests/cypress/e2e/journeys/insurance/your-business/company-details/company-details-save-and-back.spec.js
```diff
@@ -54,6 +54,13 @@ describe('Insurance - Your business - Company details page - Save and go back',
 
       cy.url().should('eq', `${Cypress.config('baseUrl')}${ROUTES.INSURANCE.ROOT}/${referenceNumber}${ALL_SECTIONS}`);
     });
+
+    it('should display the valid inputs on fields when going back to the page', () => {
+      cy.navigateToUrl(url);
+
+      yesRadioInput().first().should('be.checked');
+      yesRadioInput().eq(1).should('be.checked');
```
ttbarnes, line 62: Could we say "shouldrender submitted values" intead of the "valid inputs"?

Also do you think it's worth having cypress selectors for each radio e.g `hasTradingNameRadio.yes` `hasTradingAddressRadio.yes`etc instead of `first()` / `eq(1)`?

src/ui/server/controllers/insurance/business/company-details/helpers/populate-companies-house-summary-list.ts
```diff
@@ -0,0 +1,44 @@
...
+ * maps database response and returns body in correct format for summary list
+ * @param {object} exporterCompany
+ * @returns {CompanyHouseResponse} reformatted exporterCompany
+ */
+const mapDatabaseResponse = (exporterCompany: object) => {
+  return {
+    [COMPANY_NAME]: exporterCompany[COMPANY_NAME],
+    [COMPANY_ADDRESS]: {
+      ...exporterCompany[COMPANY_ADDRESS],
+    },
```
ttbarnes, line 22: Do we need the spread operator here? Could it just be 

```js
[COMPANY_ADDRESS]: exporterCompany[COMPANY_ADDRESS],
```

src/ui/server/controllers/insurance/business/company-details/helpers/populate-companies-house-summary-list.ts
```diff
@@ -0,0 +1,44 @@
...
+const populateCompaniesHouseSummaryList = (exporterCompany: object) => {
+  if (!exporterCompany[COMPANY_NUMBER]) {
+    return null;
+  }
+
+  const companyDetails = mapDatabaseResponse(exporterCompany);
+
+  return companyHouseSummaryList(companyDetails);
+};
+
```
ttbarnes, line 43: Could you check out this `helpers/summary-lists/answers-summary-list`, i'm working on the summary lists and trying to align everything.

This answers summary list is the same requirement we have for the other summary lists we'll be building soon; Answers summary list has very similar to this - a function that maps the data to content and is in `helpers/data-content-mappings`.

Currently in the "check answers" GET controller, summary list is generated like this:

```js
const answers = mapAnswersToContent(req.session.submittedData.quoteEligibility);

const summaryList = answersSummaryList(answers);
```


What do you think? Should we continue with the way answers are mapped and passed into summary list? Or we could do this as part of the `answersSummaryList` function

Either way I think the answers mapping to content should be in `helpers/data-content-mappings` and should name it similar to `mapAnswersToContent`; There is another one called `mapQuoteToContent`

src/ui/server/controllers/insurance/business/company-details/index.test.ts
```diff
@@ -56,15 +60,59 @@ describe('controllers/insurance/business/companies-details', () => {
   });
 
   describe('get', () => {
-    it('should render the company-details template with correct variables', () => {
+    it('should render the company-details template with correct variables if application has populated exporterCompany data', () => {
```
ttbarnes, line 63: Can we wrap the conditional in a `describe`? Easier to read IMO e.g

```js
describe('When the application has populated exporterCompany data', () => {
  it('should....
});
```
Same below

src/ui/server/controllers/insurance/business/company-details/index.ts
```diff
@@ -48,22 +49,41 @@ const exitReason = {
  * gets the template for company details page
  * @param {Express.Request} Express request
  * @param {Express.Response} Express response
- * @returns {Express.Response.render} renders company details page
+ * @returns {Express.Response.render} renders company details page with/without previously submitted details
  */
 const get = async (req: Request, res: Response) => {
-  const { application } = res.locals;
+  try {
```
ttbarnes, line 55: Seems that we can remove `async` 😊

src/ui/server/controllers/insurance/business/map-submitted-data/mocks/index.ts
```diff
@@ -11,7 +11,7 @@ export const mockBody = {
   [INPUT]: '8989898',
   __typename: 'CompaniesHouseResponse',
   ...mockCompany,
-  registeredOfficeAddress: {
+  address: {
```
ttbarnes, line 14: Shouldn't this be `registeredOfficeAddress` ?

src/ui/server/helpers/mappings/map-sic-codes.test.ts
```diff
@@ -0,0 +1,25 @@
...
+          sicCode: '12345',
+          __typename: 'ExporterCompanySicCode',
+        } as SicCodeResponse,
+      ];
+
+      const response = mapSicCodes(sicCodes);
+
+      expect(response).toEqual([sicCodes[0].sicCode, sicCodes[1].sicCode]);
+    });
+  });
```
ttbarnes, line 24: Nice, could we move the mock sic codes into `test-mocks`?

Would be nice to have the `toEqual([...` definition into an `expected` constant so it's easier to read

src/ui/server/controllers/insurance/business/map-submitted-data/mocks/index.ts
Zainzzkk: no, this is a temporary address which is processed and is passed to the api as `address` as `registeredOfficeAddress` is deleted as it contains nulls and unwanted fields like `__typename`

src/ui/server/helpers/summary-lists/company-house-summary-list.ts
```diff
@@ -22,7 +22,7 @@ const generateAddressHTML = (address: object) => {
 
   Object.keys(address).forEach((field) => {
     // if the address field exists and not the typename part
-    if (address[field] && field !== '__typename') {
+    if (address[field] && field !== '__typename' && field !== 'id') {
```
ttbarnes, line 25: Do we need to update the unit test?

src/ui/types/company-house-response.d.ts
```diff
@@ -15,8 +15,8 @@ interface CompanyHouseResponse {
   companyNumber: string;
   dateOfCreation: string;
   sicCodes: Array<string>;
-  success: boolean;
-  apiError: boolean;
+  success?: boolean;
+  apiError?: boolean;
```
ttbarnes, line 19: What's the reason for this change?

Zainzzkk: the mapping from the database comes as  CompanyHouseResponse which does not have success or apiError and did not want to set a value from them.  Also used by the API which has those flags

ttbarnes: Could we use `ApplicationExporterCompany` type instead for these instances?

e2e-tests/cypress/e2e/journeys/insurance/your-business/company-details/company-details-save-and-back.spec.js
Zainzzkk: yes, but separate ticket please as this will need a few changes

src/ui/server/controllers/insurance/business/company-details/helpers/populate-companies-house-summary-list.ts
Zainzzkk: I think the check answers approach is correct, but I think this should have a separate ticket to get working

src/ui/types/company-house-response.d.ts
Zainzzkk: I would need to remap the api response, do I do it in this ticket?

ttbarnes: I think so if it's not a big job, to avoid introducing tech debt

e2e-tests/cypress/e2e/journeys/insurance/your-business/company-details/company-details-save-and-back.spec.js
ttbarnes: https://ukef-dtfs.atlassian.net/browse/EMS-843

src/ui/server/controllers/insurance/business/company-details/helpers/populate-companies-house-summary-list.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    it('should map exporterCompany to required format', () => {
+      const exporterCompany = {
+        ...mockApplication.exporterCompany,
+      };
+
+      const response = mapDatabaseResponse(exporterCompany);
+
+      const expected = {
+        ...exporterCompany,
+        [COMPANY_SIC]: [exporterCompany[COMPANY_SIC][0].sicCode],
```
ttbarnes, line 53: This should be asserted with `mapSicCodes` as that's what the code is doing, if we use `mapSicCodes` here and that function changes, won't won't need to update this test 

Not blocker for this PR

src/ui/types/company-details.d.ts
```diff
@@ -12,4 +12,9 @@ interface CompanyDetailsFieldGroups {
   COMPANY_DETAILS: Array<SummaryListItemData>;
 }
 
-export { CompanyDetails, CompanyDetailsFieldGroups };
+interface SicCodeResponse {
+  id: string;
+  sicCode: string;
+}
```
ttbarnes, line 18: Can we just call this `SicCode`?

ttbarnes (dismissed):

ttbarnes (approved):

PR #244, opened 2023-01-16
feat(EMS-695): Your Buyer code improvements, add company/organisation address field

vensurag:
This PR improves code structure/approach for the the "your buyer" section, aligning with other areas of the codebase. Also adds an address field to the "Company or organisation" page.

## Summary of changes

- Align the structure and approach with other sections of the application codebase.
- Fix/update naming conventions, directories and URLs to align with design.
- Use field IDs and field content strings instead of having field IDs inside the field content strings.
- Add page content strings for the “company or organisation” page.
- Add E2E test coverage for “company or organisation” validation.
- Add and use page variables with FIELDS approach in the  “company or organisation” validation.
- Fix typos and indentation
- Add the "address" field to the "Company or organisation" page.


e2e-tests/cypress/e2e/journeys/insurance/your-buyer/your-buyer.spec.js
```diff
@@ -16,6 +16,8 @@ const {
   BUYER_COUNTRY_LABEL,
   BUYER_ORGANISATION,
   BUYER_ORGANISATION_LABEL,
+  BUYER_ADDRESS,
+  BUYER_ADDRESS_LABEL,
```
ttbarnes, line 20: The destructuring is labelled as `YOUR_BUYER_FIELD_IDS` but it's importing `FIELDS`, need to rename

e2e-tests/cypress/e2e/pages/insurance/your-buyer.js
```diff
@@ -3,6 +3,7 @@ import { YOUR_BUYER_FIELD_IDS } from '../../../../content-strings/fields/insuran
 const {
   BUYER_COUNTRY,
   BUYER_ORGANISATION,
+  BUYER_ADDRESS,
```
ttbarnes, line 6: Same comment as above

src/ui/server/controllers/insurance/your-buyer/validation/rules/address.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+describe('controllers/insurance/your-buyer/validation/address/', () => {
+  const mockErrors = {
+    summary: [],
+    errorList: {},
+  };
+
+  describe('when address name field  is not provided', () => {
+    it('should return validation error', () => {
+      const mockSubmittedData = { organisation: 'xdfd', country: 'XAD', address: '' };
```
ttbarnes, line 22: Can we update and use `fixtures/application` for these values please so that we have the values we submit with in tests all in one place?

src/ui/server/controllers/insurance/your-buyer/validation/rules/address.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @returns {Object} Validation errors
+ */
+export const addressRules = (formBody: RequestBody, errors: object) => {
+  const updatedErrors = errors;
+  // check if the Country field is empty.
+  if (!objectHasProperty(formBody, YOUR_BUYER.BUYER_ADDRESS.ID)) {
+    return generateValidationErrors(YOUR_BUYER.BUYER_ADDRESS.ID, BUYER_ADDRESS.IS_EMPTY, errors);
+  }
+  // check the  length of character
+  if (Object.keys(formBody[YOUR_BUYER.BUYER_ADDRESS.ID]).length > YOUR_BUYER.BUYER_ADDRESS.MAX_LENGTH) {
```
ttbarnes, line 30: No need for `Object.keys` here

e2e-tests/cypress/e2e/pages/insurance/your-buyer.js
vensurag: resolved

e2e-tests/content-strings/fields/insurance/your-buyer/index.js
```diff
@@ -3,4 +3,7 @@ export const YOUR_BUYER_FIELD_IDS = {
   BUYER_COUNTRY_LABEL: 'Country',
   BUYER_ORGANISATION: 'organisation',
   BUYER_ORGANISATION_LABEL: "Buyer's company or organisation name",
+  BUYER_ADDRESS: 'address',
+  BUYER_ADDRESS_LABEL: 'Address',
```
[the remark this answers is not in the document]
vensurag: We are having comment/ conversation every time.  These are actual; filed values declared in test and its base declaration test.Its lot of confusing. where do we need to bring the fields..

src/ui/server/controllers/insurance/your-buyer/validation/rules/address.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+      const expected = generateValidationErrors(YOUR_BUYER.BUYER_ADDRESS.ID, BUYER_ADDRESS.IS_EMPTY, mockErrors);
+
+      expect(result).toEqual(expected);
+    });
+  });
+
+  describe('when address name field exceeds max characters', () => {
+    it('should return validation error', () => {
+      const addresMockString = `1234asdfdsafsadfdsfdffdasdfdsafsadfdsfdffdasdfdsafsadfdsfdffdasdfdsafsadfdsfd
```
Zainzzkk, line 34: typo

e2e-tests/content-strings/fields/insurance/your-buyer/index.js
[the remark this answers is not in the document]
ttbarnes: We have to raise/resolve this because it wasn't addressed in the last PR.

We want to keep field IDs, separate from content strings for the field.

There is some information about this / the "why" in the README https://github.com/UK-Export-Finance/exip/tree/main-application#keeping-page-and-fields-separated

Here are some example controllers with a `pageVariables` function, that consumes field IDs and field content strings, ready to be consumed in the get/post controllers and nunjucks template:
- https://github.com/UK-Export-Finance/exip/blob/main-application/src/ui/server/controllers/insurance/policy-and-export/single-contract-policy/index.ts#L37
- https://github.com/UK-Export-Finance/exip/blob/main-application/src/ui/server/controllers/insurance/policy-and-export/about-goods-or-services/index.ts#L31

Example of field content string definitions for "policy and export": https://github.com/UK-Export-Finance/exip/blob/main-application/src/ui/server/content-strings/fields/insurance/policy-and-exports/index.ts - the first field does have an ID which is wrong, this is tech debt, All the other fields are how we want it to be.

Making this change is easier to do now rather than later, as the "your buyer" page adds more form fields and there are a lot of fields.

We want different areas of the codebase to be consistent and not deviating. This helps developers to digest the different areas and make changes, without having different approaches in different places.

src/ui/server/controllers/insurance/your-buyer/index.test.ts
```diff
@@ -27,9 +33,32 @@ describe('controllers/insurance/your-buyer/your-buyer-details', () => {
...
+      };
+
+      expect(PAGE_VARIABLES).toEqual(expected);
+    });
+  });
+
   describe('TEMPLATE', () => {
     it('should have the correct template defined', () => {
-      expect(TEMPLATE).toEqual(TEMPLATE);
+      expect(TEMPLATE).toEqual(TEMPLATES.INSURANCE.YOUR_BUYER.COMPANY_OR_ORGANISATION);
```
Zainzzkk, line 61: do we need this test?

src/ui/server/controllers/insurance/your-buyer/validation/rules/address.test.ts
```diff
@@ -0,0 +1,51 @@
...
+const {
+  YOUR_BUYER: {
+    COMPANY_OR_ORGANISATION: { ADDRESS: FIELD_ID },
+  },
+} = FIELD_IDS.INSURANCE;
+
+const {
+  INSURANCE: {
+    YOUR_BUYER: {
+      COMPANY_OR_ORGANISATION: { [FIELD_ID]: ERROR_MESSAGE },
```
Zainzzkk, line 15: is there a better way of this? it looks a bit confusing

ttbarnes: I can't think of another way to do it without breaking eslint rules - in this file we get the field ID, and get the error message for that field ID - they have the same name/ID in both files. Which is good IMO. We have this approach in other places as well

So we get the `FIELD_ID` const from field-ids, then we get the error message dynamically with the `FIELD_ID` constant but label it `ERROR_MESSAGE`. Makes the actual tests are easier to read because then we can do e.g `ERROR_MESSAGE.IS_EMPTY`.

Do you have another idea or something we could do to make it clearer?

src/ui/server/controllers/insurance/your-buyer/index.test.ts
ttbarnes: Yes, to make sure the template variable has the correct template. This is then used in the get and post controllers

src/ui/server/controllers/insurance/your-buyer/validation/rules/address.test.ts
Zainzzkk: no this seems fine then!

Zainzzkk (dismissed):
LGTM

PR #247, opened 2023-01-17
chore(EMS-613): Your business - trading address & trading name as yes/no values

ttbarnes:
This PR updates the `hasTradingAddress` and `hasTradingName` fields in the `ExporterCompany` table to be "yes or no" string values in the database, instead of a boolean.

## Why

Keystone checkbox/boolean fields, generates a nice boolean column in the SQL database - however this enforces a default value of false/0, and does not offer the ability to define a default value, or make the field nullable, without manually overiding the database and DB dump, which poses it's own risks.

This enforced default value of false, results in the Your business yes/no radio buttons being auto selected to "no", even though nothing has been provided.

In slack and some video calls, we spoke about having a "yes/no" or "answers" table and then referencing the appropriate answer in the `ExporterCompany` table. However, this would introduce some complexity in terms of getting, saving and consuming data. Keystone is not really built to be this way. We'd have to write custom resolvers to get this data which reduces the benefits we get from using Keystone.

In the interests of unblocking "Your business" tickets for the EXIP project, we're going with this approach. We can change this if required and before we (in the future) introduce more "yes or no" questions/answers.

## Summary of changes

- Create yes/no constants in the API.
- Change `hasTradingAddress` and `hasTradingName` fields in the `ExporterCompany` table to be a select keystone field with yes/no values.
- Add a `yesNoValueString` variable/condition to the `yes-no-raadio-buttons` nunujucks component. This allows this component to render yes/no radios with true or false boolean values, _or_ Yes/no values. This makes the same component work for Eligibility questions, as well as the application.
- Update company details nunjucks page to render yes/no radios component with `yesNoValueString`.
- Note: We want intend to update eligibility fields to use "yes/no" values instead of true/false booleans; To be discussed: https://ukef-dtfs.atlassian.net/browse/EMS-848


Zainzzkk (approved):
LGTM

PR #252, opened 2023-01-18
feat(EMS-605-606): Policy and exports - Check your answers - summary list

ttbarnes:
This PR adds a summary list to the Policy and export's "Check your answers" page for both a single and multiple contract policies.

## Summary of changes
- Add summary titles for every policy and export field.
- Create a new function `policyAndExportSummaryList` that generates a summary list with different fields, depending on the type of policy. Split each area up into separate functions so that it's easier to read and maintain.
- Update "check your answers" controller to get countries & currencies, call the new summary list function and render in the page.
- Created a new helper function, `getCountryByIsoCode`.
- Update the mock application used in unit tests to have a complete set of data for single/multiple contract policy types.
- Update unit tests that consume countries and currencies so that the tests for "when a currency/country is or isn't submitted, is should..." still work, since the mock application now has a complete data set.
- Created E2E tests for testing the entire summary list for both single and multiple contract policies, including creating a `checkPolicyAndExportsSummaryList` page function so that we can simplify the actual test spec, e.g `checkSummaryList[REQUESTED_START_DATE]();`

## Other improvements
- Create a new helper function `getFieldById` to simplify/make the code more readable when passing a field to `fieldGroupItem`
- Improve `generateSummaryListItemData` helper function by simplifying the "Get value" logic/function.
- Add a default `decimalPoints` param of 0 to `formatCurrency` helper function.
- Add some helper functions to the E2E test suite to avoid duplicating code in the tests.
- Re-arrrange "optional cookies" in field content strings so it's at the very top (this is a global field for every page).


src/ui/server/controllers/insurance/policy-and-export/check-your-answers/index.ts
```diff
@@ -37,14 +39,30 @@ export const get = (req: Request, res: Response) => {
...
+      ...insuranceCorePageVariables({
+        PAGE_CONTENT_STRINGS: PAGES.INSURANCE.POLICY_AND_EXPORTS.CHECK_YOUR_ANSWERS,
+        BACK_LINK: req.headers.referer,
+      }),
+      ...pageVariables(refNumber),
+      application,
+      SUMMARY_LIST: summaryList,
+    });
+  } catch (err) {
+    console.error('Error getting countries or currencie', { err });
```
Zainzzkk, line 62: typo

src/ui/server/helpers/format-currency/index.test.ts
```diff
@@ -1,10 +1,10 @@
 import formatCurrency from '.';
 
 describe('server/helpers/format-currency', () => {
-  it('shouild return a formatted currency', () => {
-    const mock = 123456;
-    const currencyCode = 'GBP';
+  const mock = 123456;
+  const currencyCode = 'GBP';
 
+  it('shouild return a formatted currency', () => {
```
Zainzzkk, line 7: typo

src/ui/server/helpers/get-country-by-iso-code/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * getCountryByIsoCode
+ * Get a country by ISO code
+ * @param {Array} Countries
+ * @param {String} Country ISO code
+ * @returns {Object} Country
+ */
+const getCountryByIsoCode = (countries: Array<Country>, isoCode: string) => {
+  const country = countries.find((c) => c.isoCode === isoCode);
```
Zainzzkk, line 11: could we use something more descriptive than c?

src/ui/server/helpers/summary-lists/generate-field-group-item/index.ts
```diff
@@ -7,17 +7,17 @@ import { SummaryListItemData, SummaryListItemDataInput } from '../../../../types
...
+ * @returns {String} Field value or default empty string dash
  */
-export const getSummaryListItemDataValue = (data: object, fieldId: string, customValue?: string): string => {
+export const getSummaryListItemDataValue = (fieldId: string, data?: object, customValue?: string): string => {
   if (customValue) {
     return customValue;
   }
 
-  if (objectHasProperty(data, fieldId)) {
+  if (data && objectHasProperty(data, fieldId)) {
```
Zainzzkk, line 20: we could add this condition to check if object exists in `objectHasProperty`?

src/ui/server/helpers/summary-lists/policy-and-export/multiple-contract-policy-fields/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const {
+  INSURANCE: {
+    POLICY_AND_EXPORTS: {
+      CONTRACT_POLICY: {
+        MULTIPLE: { TOTAL_MONTHS_OF_COVER, TOTAL_SALES_TO_BUYER, MAXIMUM_BUYER_WILL_OWE },
+      },
+    },
+  },
+} = FIELD_IDS;
```
Zainzzkk, line 16: `FIELD_IDS.INSURANCE.POLICY_AND_EXPORTS`? more readable

src/ui/server/helpers/summary-lists/policy-and-export/policy-and-date-fields/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param {Object} All submitted policy and export data
+ * @returns {Object} All policy and date fields and values in an object structure for GOVUK summary list structure
+ */
+const generatePolicyAndDateFields = (answers: ApplicationPolicyAndExport) => {
+  const fields = [
+    fieldGroupItem({
+      field: { id: POLICY_TYPE, ...FIELDS[POLICY_TYPE] },
+      data: answers,
+    }),
+    fieldGroupItem({ field: { id: REQUESTED_START_DATE, ...FIELDS.CONTRACT_POLICY[REQUESTED_START_DATE] } }, formatDate(answers[REQUESTED_START_DATE])),
```
Zainzzkk, line 28: this is quite long.  Could we assign to constants instead and then put the constant in the function call?

src/ui/server/helpers/summary-lists/policy-and-export/single-contract-policy-fields/index.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  INSURANCE: {
+    POLICY_AND_EXPORTS: {
+      CONTRACT_POLICY: {
+        SINGLE: { CONTRACT_COMPLETION_DATE, TOTAL_CONTRACT_VALUE },
+      },
+    },
+  },
+} = FIELD_IDS;
+
+describe('server/helpers/summary-lists/policy-and-export/policy-and-date-fields', () => {
```
Zainzzkk, line 19: wrong name?

src/ui/server/helpers/summary-lists/policy-and-export/single-contract-policy-fields/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const {
+  INSURANCE: {
+    POLICY_AND_EXPORTS: {
+      CONTRACT_POLICY: {
+        SINGLE: { CONTRACT_COMPLETION_DATE, TOTAL_CONTRACT_VALUE },
+      },
+    },
+  },
+} = FIELD_IDS;
```
Zainzzkk, line 16: like above, can we do FIELD_IDS.INSURANCE.POLICY_AND_EXPORTS

e2e-tests/cypress/fixtures/application.js
```diff
@@ -3,11 +3,12 @@ import {
   getMonth,
   getYear,
 } from 'date-fns';
-import { FIELD_IDS } from '../../constants';
+import { FIELD_IDS, FIELD_VALUES } from '../../constants';
```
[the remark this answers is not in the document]
ttbarnes: No, good spot

src/ui/server/helpers/summary-lists/generate-field-group-item/index.ts
ttbarnes: I don't think we should - the object property in `objectHasProperty` is a required parameter, so it won't compile if an object is _not_ provided, don't think we should change this.

Whereas in this function (`getSummaryListItemDataValue`), the data object is an optional parameter so we have the check in here

src/ui/server/helpers/summary-lists/policy-and-export/multiple-contract-policy-fields/index.ts
ttbarnes: Linting would complain about that 😄 I've updated it to be more readable - want to use this in other places as well.

Instead of importing root `FIELD_IDS`, just import the ones needed for the file

```js
import FIELD_IDS from '../../../../constants/field-ids/insurance/policy-and-exports';
```

src/ui/server/helpers/summary-lists/policy-and-export/policy-and-date-fields/index.ts
ttbarnes: I agree - linting made it like this, it's a lot clearer with new lines, but eslint doesn't like that 😞 . I am thinking of creating a new helper function `getFieldById`, to use throughout these summary list functions so we can do this:

```js
fieldGroupItem({ field: getFieldById(FIELDS, REQUESTED_START_DATE) }, formatDate(answers[REQUESTED_START_DATE])),
```

Instead of this:

```js
fieldGroupItem({ field: { id: REQUESTED_START_DATE, ...FIELDS.CONTRACT_POLICY[REQUESTED_START_DATE] } }, formatDate(answers[REQUESTED_START_DATE])),
```

What do you think?

src/ui/server/helpers/summary-lists/policy-and-export/single-contract-policy-fields/index.ts
ttbarnes: Good shout i've done the same as above - importing directly from `policy-and-exports` `FIELD_IDS`, instead of the root.

src/ui/server/helpers/summary-lists/policy-and-export/policy-and-date-fields/index.ts
Zainzzkk: I think that will be better.  Quite hard to read currently

ttbarnes: Updated all summary lists to use this new approach with new helper function, `getFieldById` which we could use in other places as well. The only summary list that doesn't use this is the quote summary list because the "titles" are structured differently to other summary lists, tech debt

Zainzzkk (approved):
LGTM

PR #256, opened 2023-01-20
fix(EMS-792-810): Policy end exports - pre-population issues

ttbarnes:
This PR fixes two issues in the single/multiple policy type page where:

- After submitting the form and revisiting the page, the "credit period with buyer" field would render an incorrect value.
- When submitting the form with a currency selection, but there are other validation errors, previously selected currency would not be auto selected.

## Changes

- Add currency mapping to the single and multiple POST controllers.
- Fix a typo in the templates for single and multiple "credit period with buyer" field.


Zainzzkk (approved):
LGTM

PR #257, opened 2023-01-20
chore(EMS-787): use constants instead of hard coded currency codes

ttbarnes:
This PR replace all hard coded instance of currency codes

## Changes

- Add some `CURRENCY_CODE` fixtures to E2E and constants to UI (only the currencies that EXIP supports).
- Update all hard coded currency codes.
- Also updates some E2E instances with `toLocaleString` functions, to use `formatCurrency` helper function.


Zainzzkk (approved):
LGTM

PR #260, opened 2023-01-24
feat(EMS-642-643): Your business - Nature of your business - Number of employees fields

Zainzzkk:
# Changes made:
* Added number of employee fields
* Added validation
* Added helper for whole number validation
* Updated and added tests

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/nature-of-business-page.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/nature-of-business-page.spec.js
@@ -13,6 +13,8 @@ const {
   NATURE_OF_YOUR_BUSINESS: {
     GOODS_OR_SERVICES,
     YEARS_EXPORTING,
+    EMPLOYEES_UK,
+    EMPLOYEES_INTERNATIONAL,
   },
 } = FIELD_IDS.INSURANCE.EXPORTER_BUSINESS;
 
@@ -103,6 +105,27 @@ context('Insurance - Your business - Nature of your business page - As an Export
 
     field.hint().contains(FIELDS.NATURE_OF_YOUR_BUSINESS[YEARS_EXPORTING].HINT);
     checkText(field.suffix(), FIELDS.NATURE_OF_YOUR_BUSINESS[YEARS_EXPORTING].SUFFIX);
+  });
+
+  it('should display the continue and save and go back button', () => {
+    checkText(submitButton(), BUTTONS.CONTINUE);
+
+    checkText(saveAndBackButton(), BUTTONS.SAVE_AND_BACK);
+  });
+
+  it(`should display ${EMPLOYEES_UK} and ${EMPLOYEES_INTERNATIONAL} input boxes`, () => {
+    const employeesUKFieldId = EMPLOYEES_UK;
+    const employeesUKField = natureOfBusiness[employeesUKFieldId];
+
+    const employeesInternationalFieldId = EMPLOYEES_INTERNATIONAL;
+    const employeesInternationalField = natureOfBusiness[employeesInternationalFieldId];
+
+    employeesUKField.input().should('exist');
+    checkText(employeesUKField.label(), FIELDS.NATURE_OF_YOUR_BUSINESS[EMPLOYEES_UK].LABEL);
+    checkText(employeesUKField.heading(), FIELDS.NATURE_OF_YOUR_BUSINESS[EMPLOYEES_UK].HEADING);
+
+    employeesInternationalField.input().should('exist');
+    checkText(employeesInternationalField.label(), FIELDS.NATURE_OF_YOUR_BUSINESS[EMPLOYEES_INTERNATIONAL].LABEL);
   });
 
   it('should display the continue and save and go back button', () => {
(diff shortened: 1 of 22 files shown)


e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/nature-of-business-page.spec.js
```diff
@@ -110,4 +112,25 @@ context('Insurance - Your business - Nature of your business page - As an Export
 
     checkText(saveAndBackButton(), BUTTONS.SAVE_AND_BACK);
   });
+
+  it(`should display ${EMPLOYEES_UK} and ${EMPLOYEES_INTERNATIONAL} input boxes`, () => {
```
ttbarnes, line 116: Can we split these up into their own "it's"?

e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/validation/nature-of-business-employees-validation.spec.js
```diff
@@ -0,0 +1,276 @@
...
+    Cypress.Cookies.preserveOnce('_csrf');
+    Cypress.Cookies.preserveOnce('connect.sid');
+  });
+
+  describe(`${EMPLOYEES_UK} validation`, () => {
+    describe(`${EMPLOYEES_UK} error`, () => {
+      describe(`when ${EMPLOYEES_UK} is left empty`, () => {
+        const errorMessage = NATURE_OF_BUSINESS_ERRORS[EMPLOYEES_UK].IS_EMPTY;
+
+        it(`should display validation errors if ${EMPLOYEES_UK} left empty`, () => {
```
ttbarnes, line 64: No need for the "if X left empty" here because we describe it above

e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/validation/nature-of-business-years-exporting-validation.spec.js
```diff
@@ -141,6 +149,8 @@ describe('Insurance - Your business - Nature of your business page - As an Expor
       const field = natureOfBusiness[fieldId];
 
       completeField(GOODS_OR_SERVICES, 'abc');
+      completeField(EMPLOYEES_UK, '5');
+      completeField(EMPLOYEES_INTERNATIONAL, '5');
```
ttbarnes, line 153: These additions seem troublesome to maintain (same in previous file). Instead of having to add new data entry to each test like this (in order to test a single error message) could we not just do e.g `partials.errorSummaryListItems().eq(3),` instead? Did this in policy and exports and worked well

src/ui/server/controllers/insurance/business/nature-of-business/validation/rules/employees-international.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * validates  number of international employees input
+ * only allows number without decimal
+ * @param {RequestBody} responseBody
+ * @param {object} errors
+ * @returns {object} errors
+ */
+const employeesInternational = (responseBody: RequestBody, errors: object) => {
+  if (!objectHasProperty(responseBody, EMPLOYEES_INTERNATIONAL)) {
+    const errorMessage = EXPORTER_BUSINESS[EMPLOYEES_INTERNATIONAL].IS_EMPTY;
```
ttbarnes, line 23: What do you think about using destructure approach with field ID and having an `ERROR_MESSAGE` const like in other places? e.g `your-buyer/validation/rules/address`

src/ui/server/helpers/whole-number-validation/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  if (!isNumber(responseBody[field]) || hasDecimal || isBelowMinimum) {
+    const errorMessage = errorField[field].INCORRECT_FORMAT;
+
+    return generateValidationErrors(field, errorMessage, errors);
+  }
+
+  return errors;
+};
+
```
ttbarnes, line 26: Nice - i'm wondering if it's worth having a new number helper `isWholeNumber` which does the checks (line 15 to 18), and then we consume the new helper in this validation file. What do you think?

Because then we'd have the ability to check for whole numbers without having to pass validation errors. Gives more flexibility

e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/validation/nature-of-business-years-exporting-validation.spec.js
Zainzzkk: should I change in this PR?

src/ui/server/helpers/whole-number-validation/index.ts
Zainzzkk: can we make a tech debt for this please?

e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/validation/nature-of-business-years-exporting-validation.spec.js
ttbarnes: I think so, don't think it'd take long? Otherwise we're introducing tech debt?

src/ui/server/helpers/whole-number-validation/index.ts
ttbarnes: https://ukef-dtfs.atlassian.net/browse/EMS-891

ttbarnes (approved):

PR #262, opened 2023-01-24
fix(EMS-793): Policy and exports - fix string fields/inputs not saving as empty with save & back functionality

ttbarnes:
This PR changes the "sanitise form data" functions and therefore fixing an issue where a text input would not be sent to the API/saved in the database, if the value has been removed after being submitted with a value. We previously stripped out empty form fields before sending to the API. Also includes some technical improvements.

## Summary of changes
- Update `sanitiseData` function so that it does not return a field that is in a set of specific number fields and has an empty value. Without this, trying to send number fields with an empty string or value of 0, results in API/DB error and the data is not saved. 
  - Created `NUMBER_FIELDS` array of specific number fields that are in the applicatino forms.
  - Created `shouldIncludeAndSanitiseField` check.
  - Improved `shouldChangeToNumber` checks so that all checks are in this function instead of being outside.
- Simplify `hasFormData` function - do not strip out empty form fields and remove unnecessary param checks (typescript handles this).
- Simplfiy `mapSubmittedData` function - remove unnecessary param checks.
- Simplfiy `getValidFields` function - remove unnecessary param checks and add an empty string check.
- Update company details controller tests to use `sanitiseValue, passing the key and value instead of just the value. Also renamed the page variables import/usage.
- Add E2E tests for wiping the "credit period with buyer" field via "save and go back", for both single and multiple contracts.

### DIFF
--- a/src/ui/server/helpers/get-valid-fields/index.ts
+++ b/src/ui/server/helpers/get-valid-fields/index.ts
@@ -6,25 +6,19 @@
  * @returns {Object} Form data with invalid fields removed
  */
 const getValidFields = (formData: object, errorList: object) => {
-  if (formData) {
-    if (errorList) {
-      const fieldsWithErrors = Object.keys(errorList);
+  const fieldsWithErrors = Object.keys(errorList);
 
-      const validFields = {};
+  const validFields = {};
 
-      Object.keys(formData).forEach((fieldName) => {
-        if (!fieldsWithErrors.includes(fieldName)) {
-          validFields[fieldName] = formData[fieldName];
-        }
-      });
+  Object.keys(formData).forEach((fieldName) => {
+    const fieldValue = formData[fieldName];
 
-      return validFields;
+    if (!fieldsWithErrors.includes(fieldName) && fieldValue !== '') {
+      validFields[fieldName] = fieldValue;
     }
+  });
 
-    return formData;
-  }
-
-  return {};
+  return validFields;
 };
 
 export default getValidFields;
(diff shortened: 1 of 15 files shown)


src/ui/server/helpers/get-valid-fields/index.ts
```diff
@@ -6,25 +6,19 @@
...
-      Object.keys(formData).forEach((fieldName) => {
-        if (!fieldsWithErrors.includes(fieldName)) {
-          validFields[fieldName] = formData[fieldName];
-        }
-      });
+  Object.keys(formData).forEach((fieldName) => {
+    const fieldValue = formData[fieldName];
 
-      return validFields;
+    if (!fieldsWithErrors.includes(fieldName) && fieldValue !== '') {
```
Zainzzkk, line 16: a comment here please for conditions.  Worth having an `isEmptyString` helper?

src/ui/server/helpers/sanitise-data/index.ts
```diff
@@ -17,9 +22,9 @@ const {
  * @param {String | Number} Field value
  * @returns {Boolean}
  */
-const shouldChangeToNumber = (value: string | number) => {
-  if (typeof value === 'string' && isNumber(value)) {
-    return true;
+const shouldChangeToNumber = (key: string, value: string | number) => {
+  if (key === COMPANY_NUMBER || key === PHONE_NUMBER || value === '') {
```
Zainzzkk, line 26: could use `isEmptyString` here

ttbarnes: Great shout, created and added new helper function

Zainzzkk (approved):
LGTM

PR #263, opened 2023-01-25
chore(tech): various technical improvements

ttbarnes:
This PR improves various parts of code base.

## Summary of changes

- Rename `MULTI` to `MULTIPLE` and "multi" to "multiple".
- Move policy and exports server routes into their own file.
- Align all insurance server routes and make sure that the unit tests are running properly (some were missing a `beforeEach`, causing the tests to not run properly).
- Update and align all insurance console logs when there is an error updating an application, so that we can more easily tell where the call was made.
- Update E2E content strings for the "total months of cover" field so that it's aligned with UI and the value is generated dynamically.
- Simplify and align companies house summary list generation functions so that it's returning and named "fields" instead of "field groups".


Zainzzkk (approved):
LGTM

PR #264, opened 2023-01-25
fix(EMS-869): Policy and exports - add missing hint text to currency code fields

ttbarnes:
This PR adds a missing hint to the policy currency code field in both the single and contract policy pages.

Also, because the same field and hint is used in multiple places, some technical improvements have been made:

- Create a reusable currency code nunjucks component.
- Create E2E helper/support function, `checkPolicyCurrencyCodeInput`.
- Update single and multiple contract pages and E2E tests to consume the above.


abhi-markan (dismissed):
LGTM

Zainzzkk (approved):
LGTM

PR #268, opened 2023-01-26
fix(EMS-852): Order countries alphabetically

ttbarnes:
Update `mapCountries` funciton to order the countries alphabetically.

This mapping function is consumed only in the full application i.e not in eligibility - eligibility uses countries from the CSI API and has a  a different mapping function (due to different data structure).


Zainzzkk (approved):
LGTM

PR #269, opened 2023-01-26
fix(EMS-884): Your business - Company Details - Fix website and phone number max lengths

Zainzzkk:
# Issue: 
* Would redirect to problem with website page if website or phone number over 191 characters

# Changes made:
* Added helper to check if above maximum number of characters
* Added check to website and phone number fields
* Added tests


e2e-tests/constants/examples/phone-number-examples.js
```diff
@@ -12,6 +12,7 @@ export const INVALID_PHONE_NUMBERS = {
   INTERNATIONAL: '0033144513100',
   INTERNATIONAL_PLUS: '+337906 208973',
   TOO_SHORT: '*44',
+  ABOVE_MAX_CHARS: '02072718010'.repeat(50),
```
ttbarnes, line 15: Do you think it'd be nicer if we did `'1'.repeat(192)`? the `repeat(50)` is a bit vague whereas having 192 shows what the maximum is?

src/ui/server/shared-validation/website-address/index.test.ts
```diff
@@ -34,6 +34,15 @@ describe('shared-validation/website', () => {
 
       expect(result).toEqual(expected);
     });
+
+    it('should display error when website is above 191 characters', () => {
+      const website = `www.google${'e'.repeat(190)}.com`;
```
ttbarnes, line 39: Could we test with specifically 192 characters? `'w'.repeat(192)`, Currently if the maximum is changed this test could have a false pass

Zainzzkk: this would trigger the invalid website check.  I wanted it to be valid but too long

ttbarnes: OK how about this? To get 192 characters

```js
`www.google${'e'.repeat(178)}.com`
```

PR #270, opened 2023-01-26
fix(EMS-851): Explicit list of string number fields to allow submission/saving if submitted as a pure number

ttbarnes:
This PR adds a an explicit list of string number fields, `STRING_NUMBER_FIELDS` - string fields that could be numbers.

When sanitising form data before sending to the API, we have a check - `shouldChangeToNumber`. This is great - ensuring that a pure number in a string, is converted to an actual number type.

However, there are scenarios where a user could enter a pure number, e.g "1234" into a free text field. The default sanitise data behaviour is to transform this into a number type, but we do not want this to happen on specific fields.

Without introducing an explicit "string number fields that we don't want to transform if they have pure numbers", the data save will fail and the user will be taken to a "problem with service" page. 

An alternative approach is to add additional validation handling in the UI, however this is more work in the long run and not an elegant design for users.

This scenario is edge case, but it could happen so we are introducing this approach.

## Summary of changes

- Add `STRING_NUMBER_FIELDS` to sanitise data file.
- Update `shouldChangeToNumber` to check the list of number fields.
- Add E2E test coverage for all scenarios in the policy and export pages.

@Zainzzkk We should add an E2E test like in this PR for the `GOODS_OR_SERVICES` field in the company section.

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/about-goods-or-services/about-goods-or-services.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/about-goods-or-services/about-goods-or-services.spec.js
@@ -46,10 +46,6 @@ const {
 
 const task = taskList.prepareApplication.tasks.policyTypeAndExports;
 
-const goToPageDirectly = (referenceNumber) => {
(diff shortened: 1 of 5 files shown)


Zainzzkk (approved):
LGTM

PR #271, opened 2023-01-26
fix(EMS-867-EMS-877): Policy and exports - add invalid date format validation, fix typos in Your buyer address field

ttbarnes:
This PR adds "is valid"  checks to the "cover start" and "contract completion" date fields in the policy and exports section. We now present an error to the user if they e.g enter an incorrect month of e.g "24", or in US style. Also fixed a small typo bug in the your buyer page.

## Summary of changes

- Use `isValid` date-fns function to check the date for invalid formatting in both date validation functions.
- Rename `IS_EMPTY` error message to `INCORRECT_FORMAT`.
- Add E2E tests for all instances (single and multiple pages).
- Align requested start date and contract completion date validation patterns (returning errors in each condition instead of assigning variable and simplify the default return param).
- EMS-877: Fix some typos for maximum characters allowed in the "address" field of Your buyer.


Zainzzkk (approved):
LGTM

Zainzzkk (approved):
LGTM

PR #273, opened 2023-01-26
fix(EMS-874): Policy and exports - save fields that have had values removed/is empty

ttbarnes:
## Why

A user might have submitted a previous value, but then want to remove it. Note - this only applies to free text fields (only two of these in the Policy and exports section).

## Changes

- Remove the false `isEmptyString` check in `getValidFields` function, so the field doesn't get stripped out and therefore it's sent to the API and updated in the database.
- Update the map/save data functions in "your business " so that we manually strips out empty form fields for this section.
- Update companies house summary list generation to return default empty dash if there is no address or sic codes in the data.
- Add E2E tests for all relevant fields in policy and exports, where
  - A value is submitted.
  - User then goes back to the page and removes the value.
  - The task status should be "in progress".
  - When reviisiting the page, the field should not have a value pre-populated.

## Other improvements
- Updated the said E2E tests to use the `checkTaskStatus` cypress command.
- Remove unnecessary parameter checks in `stripEmptyFormFields` function, typescript does this for us.


e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/about-goods-or-services/save-and-back.spec.js
```diff
@@ -64,8 +64,7 @@ context('Insurance - Policy and exports - About goods or services page - Save an
     });
 
     it('should retain the `type of policy and exports` task status as `in progress`', () => {
-      const expected = TASKS.STATUS.IN_PROGRESS;
-      cy.checkText(task.status(), expected);
+      cy.checkTaskStatus(task, TASKS.STATUS.IN_PROGRESS);
```
Zainzzkk, line 67: should we dereference IN_PROGRESS at the top? since used many times

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/multiple-contract-policy/save-and-back.spec.js
```diff
@@ -92,8 +91,7 @@ context('Insurance - Policy and exports - Multiple contract policy page - Save a
     });
 
     it('should retain the `type of policy and exports` task status as `in progress`', () => {
-      const expected = TASKS.STATUS.IN_PROGRESS;
-      cy.checkText(task.status(), expected);
+      cy.checkTaskStatus(task, TASKS.STATUS.IN_PROGRESS);
```
Zainzzkk, line 94: should we dereference IN_PROGRESS at the top?

e2e-tests/cypress/e2e/journeys/insurance/policy-and-exports/single-contract-policy/save-and-back.spec.js
```diff
@@ -74,8 +74,7 @@ context('Insurance - Policy and exports - Single contract policy page - Save and
     });
 
     it('should retain the `type of policy and exports` task status as `in progress`', () => {
-      const expected = TASKS.STATUS.IN_PROGRESS;
-      cy.checkText(task.status(), expected);
+      cy.checkTaskStatus(task, TASKS.STATUS.IN_PROGRESS);
```
Zainzzkk, line 77: should we dereference IN_PROGRESS at the top?

src/ui/server/controllers/insurance/business/map-submitted-data/index.ts
```diff
@@ -38,14 +38,11 @@ const mapSubmittedData = (formBody: RequestBody, application: Application): obje
     delete populatedData.registeredOfficeAddress;
   }
 
-  // deletes companies house input field
+  // convert and populate company number and delete the companies house input field
   if (objectHasProperty(populatedData, INPUT)) {
-    delete populatedData[INPUT];
-  }
+    populatedData[COMPANY_NUMBER] = populatedData[INPUT].toString();
```
Zainzzkk, line 43: should this not come from the API output?

ttbarnes: Good shout, reverted this change

Zainzzkk (dismissed):
LGTM

PR #275, opened 2023-01-27
fix(EMS-834-911): Policy and exports - update total contract value hint, update external links

ttbarnes:
This PR updates the hint for "total contract value" in the single contract policy page. Also fixes some incorrect external links.


Zainzzkk (approved):
lgtm

PR #276, opened 2023-01-27
fix(EMS-913): Policy and exports - fix date issue when changing contract from multiple to single

ttbarnes:
The policy and exports summary list was rendering a default date (with year 1970) for the "contract completion date" field if a date is not passed.

## Changes

- Add a condition to the summary list row value param.
- Improve test coverage for summary list row value conditions.
- Add E2E test to ensure we don't regress in the future.

### DIFF
--- a/src/ui/server/helpers/summary-lists/policy-and-export/single-contract-policy-fields/index.ts
+++ b/src/ui/server/helpers/summary-lists/policy-and-export/single-contract-policy-fields/index.ts
@@ -34,7 +34,7 @@ const generateSingleContractPolicyFields = (answers: ApplicationPolicyAndExport,
         renderChangeLink: true,
         href: `${INSURANCE_ROOT}/${referenceNumber}${SINGLE_CONTRACT_POLICY_CHANGE}#${CONTRACT_COMPLETION_DATE}-label`,
       },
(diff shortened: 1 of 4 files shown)


src/ui/server/helpers/summary-lists/policy-and-export/single-contract-policy-fields/index.ts
```diff
@@ -34,7 +34,7 @@ const generateSingleContractPolicyFields = (answers: ApplicationPolicyAndExport,
         renderChangeLink: true,
         href: `${INSURANCE_ROOT}/${referenceNumber}${SINGLE_CONTRACT_POLICY_CHANGE}#${CONTRACT_COMPLETION_DATE}-label`,
       },
-      formatDate(answers[CONTRACT_COMPLETION_DATE]),
+      answers[CONTRACT_COMPLETION_DATE] && formatDate(answers[CONTRACT_COMPLETION_DATE]),
```
Zainzzkk, line 37: is there a unit test for this?

ttbarnes: There is - strange because it's got 100% test coverage. I'll add one in anyway

PR #278, opened 2023-01-30
fix(EMS-911): policy and exports - update links to external proposal form

ttbarnes:
Update the link used in 3 policy and export fields so the link goes directly to an external form instead of a page that links to the PDF.


Zainzzkk (approved):
LGTM

PR #282, opened 2023-01-31
fix(EMS-880-938): Company Details - Handle company with no SIC code, delete previous SIC code, remove __typename from mapping

Zainzzkk:
# Issue:
* Companies house response without SIC code would cause problem with service
* Did not delete previous SIC codes
* __typename was not needed in mapping

# Changes made:
* Fixed handling of response without SIC code
* Fixed deletion of previous SIC code if there is input and api receives exporterCompany
* Removed __typename
* Added to e2e test


e2e-tests/cypress/e2e/journeys/insurance/your-business/company-details/company-house-search.spec.js
```diff
@@ -158,4 +165,23 @@ context('Insurance - Your business - Company details page - company house search
       cy.checkText(partials.yourBusinessSummaryList[COMPANY_SIC].value(), '64999');
     });
   });
+
+  describe('when the company does not have a sic code', () => {
+    it('should display your business summary list with a "-" for sic code when coming back to the company details page', () => {
```
ttbarnes, line 170: Can we use `DEFAULT.EMPTY` for all instances of "-" please

src/ui/server/helpers/summary-lists/company-house-summary-list.ts
```diff
@@ -43,7 +43,7 @@ const generateAddressHTML = (address: object): string => {
  * @returns {String} Sic codes as a single string or default empty string
  */
 const generateSicCodesValue = (sicCodes: Array<string>): string => {
-  if (sicCodes.length) {
+  if (sicCodes && sicCodes.length) {
```
ttbarnes, line 46: Can we update the unit test for `generateSicCodesValue` to check if there is no `sicCodes` provided? It currently only checks for empty array 

`sicCodes` param should now be optional

ttbarnes (approved):
Great work

ttbarnes (approved):

PR #287, opened 2023-02-01
fix(EMS-911-919): Policy and exports - fix field hint link and error message

ttbarnes:
This PR fixes 2 issues in the Policy and exports - single contract policy page.

- Incorrect link to external form in the "maximum buyer will owe" field hint.
- Incorrect/outdated error message for the "total contract value" field.
- Also updated the E2E tests for single/contract policy hint links to external form so that they assert from link content strings instead of field content strings.

### DIFF
--- a/src/ui/server/content-strings/fields/insurance/policy-and-exports/index.ts
+++ b/src/ui/server/content-strings/fields/insurance/policy-and-exports/index.ts
@@ -110,7 +110,7 @@ export const POLICY_AND_EXPORTS_FIELDS = {
           NEED_MORE_COVER: 'If you need cover for more than £499,999, ',
           FILL_IN_FORM: {
             TEXT: 'fill in this form instead.',
-            HREF: LINKS.EXTERNAL.FULL_APPLICATION,
+            HREF: LINKS.EXTERNAL.PROPOSAL_FORM,
           },
           NO_DECIMALS: 'Enter a whole number - do not enter decimals.',
         },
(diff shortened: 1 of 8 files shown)


src/ui/server/content-strings/fields/insurance/policy-and-exports/index.ts
```diff
@@ -110,7 +110,7 @@ export const POLICY_AND_EXPORTS_FIELDS = {
           NEED_MORE_COVER: 'If you need cover for more than £499,999, ',
           FILL_IN_FORM: {
             TEXT: 'fill in this form instead.',
-            HREF: LINKS.EXTERNAL.FULL_APPLICATION,
+            HREF: LINKS.EXTERNAL.PROPOSAL_FORM,
```
Zainzzkk, line 113: Is there a test which would could test the href links to the right form?

ttbarnes: There are E2E tests for this, but they test the value in field content strings (what's been changed here), which makes sense IMO.

We could update the E2E test to check the value from `LINKS` instead of field content strings. What do you think?

Zainzzkk: yes, maybe we could check the attr.href links tot he right place.  I think I did so in one of my e2e tests

Zainzzkk (approved):
LGTM

PR #288, opened 2023-02-01
feat(EMS-644): Company Details - Turnover - Financial year end date section

Zainzzkk:
# Changes made:
* Added `financialYearEndDate` from `companiesHouseResponse` and add to the DB after formatting - added during company details search
* Added turnover page
* Added helpers for dates
* Added e2e test and unit tests
* Updated `mapApplicationToFormFields` to map "financial year end date".
* Fixed an issue with mock dates in unit and E2E tests where "when date is in the past" tests would fail on the 2nd of a month.


e2e-tests/cypress/e2e/journeys/insurance/your-business/nature-of-business/nature-of-business-page.spec.js
```diff
@@ -49,6 +50,7 @@ context('Insurance - Your business - Nature of your business page - As an Export
       referenceNumber = id;
 
       const url = `${Cypress.config('baseUrl')}${ROOT}/${referenceNumber}${NATURE_OF_BUSINESS}`;
+      turnoverUrl = `${Cypress.config('baseUrl')}${ROOT}/${referenceNumber}${TURNOVER}`;
```
ttbarnes, line 53: can we name `const url` `natureOfBusinessUrl` just so it's clearer

e2e-tests/cypress/e2e/journeys/insurance/your-business/turnover/turnover-financial-year-end.spec.js
```diff
@@ -0,0 +1,98 @@
...
+  EXPORTER_BUSINESS: {
+    TURNOVER,
+  },
+} = ROUTES.INSURANCE;
+
+const { taskList } = partials.insurancePartials;
+
+const task = taskList.prepareApplication.tasks.exporterBusiness;
+
+context('Insurance - Your business - Turnover page - when financialYearEnd exists', () => {
```
ttbarnes, line 26: Can we use the field ID instead of financialYearEnd

e2e-tests/cypress/e2e/journeys/insurance/your-business/turnover/turnover-financial-year-end.spec.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  beforeEach(() => {
+    Cypress.Cookies.preserveOnce('_csrf');
+    Cypress.Cookies.preserveOnce('connect.sid');
+  });
+
+  it(`should not display ${FINANCIAL_YEAR_END_DATE} section`, () => {
+    const fieldId = FINANCIAL_YEAR_END_DATE;
+    const field = turnover[fieldId];
```
ttbarnes, line 92: Can we move `fieldId` and `field` to just before the first `context(...`? Declared twice in this file

src/api/helpers/create-full-timestamp-from-day-month.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      // @ts-ignore
+      const result = createTimestampFromNumbers(undefined, month);
+
+      expect(result).toEqual(null);
+    });
+  });
+
+  describe('when month is not provided', () => {
+    it('should return null', () => {
+      // @ts-ignore
```
ttbarnes, line 18: We should make the day and month params option fields (`day?: string, month?: string`), then we don't need to add TS ignores

src/api/helpers/create-full-timestamp-from-day-month.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * createTimestampFromNumbers
+ * Create a date from day, month
+ * Adds a default year of this year
+ * @param {Number} day
+ * @param {Number} month
```
ttbarnes, line 6: String 😛

src/ui/server/controllers/insurance/business/nature-of-business/index.ts
```diff
@@ -101,6 +101,8 @@ const post = async (req: Request, res: Response) => {
       return res.redirect(ROUTES.PROBLEM_WITH_SERVICE);
     }
 
+    const { referenceNumber } = application;
```
ttbarnes, line 104: Can we get this from params? To be consistent

src/ui/server/controllers/insurance/business/turnover/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const pageVariables = (referenceNumber: number) => ({
+  FIELDS: {
+    FINANCIAL_YEAR_END_DATE: {
+      ID: FINANCIAL_YEAR_END_DATE,
+      ...TURNOVER_FIELDS[FINANCIAL_YEAR_END_DATE],
+    },
+  },
+  POST_ROUTES: {
+    NATURE_OF_BUSINESS: `${INSURANCE_ROOT}/${referenceNumber}${TURNOVER_ROOT}`,
```
ttbarnes, line 30: I don't think we need this? Only `SAVE_AND_BACK_URL`? if it's just save and back (which I think is the case) we don't need `POST_ROUTES` either

src/ui/server/controllers/insurance/business/turnover/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      return res.redirect(ROUTES.PROBLEM_WITH_SERVICE);
+    }
+
+    const { exporterCompany } = application;
+
+    // values from application if they exist
+    const submittedValues = {
+      // if FINANCIAL_YEAR_END_DATE exists, then will display in correct format or will not render section if null
+      [FINANCIAL_YEAR_END_DATE]: exporterCompany[FINANCIAL_YEAR_END_DATE] ? chooseDateFormat(exporterCompany[FINANCIAL_YEAR_END_DATE], 'd MMMM') : null,
+    };
```
ttbarnes, line 55: Can we move this into a `mapSubmittedData` function?

src/ui/server/controllers/insurance/business/turnover/index.ts
Zainzzkk: I have followed above for the all the other pages in your-business? Forgot to rename NATURE_OF_BUSINESS to TURNOVER

src/ui/server/controllers/insurance/business/turnover/index.ts
Zainzzkk: this is the only field that needs mapping though?

src/ui/server/helpers/date/choose-date-format/index.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * chooseDateFormat
+ * Format a date with whichever pattern chosen
+ * @param {Date} timestamp
+ * @param {String} pattern pattern of format
+ * @returns {String} Formatted date
+ */
+const chooseDateFormat = (timestamp: Date, pattern: string): string => format(new Date(timestamp), pattern);
+
+export default chooseDateFormat;
```
ttbarnes, line 12: We already have a `formatDate` helper function - `helpers/date/format-date`

I think we should update that so it has an optional `format` param, and if none is provided it defaults to whats currently in there `'d MMMM yyyy'`

src/ui/server/test-mocks/mock-application.ts
```diff
@@ -74,6 +74,7 @@ const mockApplication = {
       premises: '',
       __typename: 'ExporterCompanyAddress',
     },
+    financialYearEndDate: '2023-07-31 00:00:00.000',
```
ttbarnes, line 77: Should we generate dynamically?

src/ui/server/controllers/insurance/business/turnover/index.ts
ttbarnes: We only need it for companies house page because there are 2 forms - also noticed there is this in the "continue" submit button:

```js
'formaction': POST_ROUTES.NATURE_OF_BUSINESS
```

We don't need this, native HTML will pick up the same route without this, simpler 😄

src/ui/server/controllers/insurance/business/turnover/index.ts
ttbarnes: I think it's cleaner to have mapping seperate, keeps the controller and controller test cleaner? We could do this as tech improvement after PR is merged if you'd like

ttbarnes (approved):

PR #291, opened 2023-02-03
feat(EMS-646): Company Details - Turnover - Percentage of Turnover Field and new shared cypress validation command

Zainzzkk:
# Changes made:
* Added percentage of turnover field
* Added validation
* Added helper for maximum value
* Added tests
* Added cypress command for checking validation `submitAndAssertErrors`

### DIFF
--- a/e2e-tests/constants/field-ids/insurance/exporter-business/index.js
+++ b/e2e-tests/constants/field-ids/insurance/exporter-business/index.js
@@ -25,6 +25,6 @@ export const EXPORTER_BUSINESS = {
   TURNOVER: {
     FINANCIAL_YEAR_END_DATE: 'financialYearEndDate',
     ESTIMATED_ANNUAL_TURNOVER: 'estimatedAnnualTurnover',
-    PERCENTAGE_TURNOVER: 'percentageTurnover',
+    PERCENTAGE_TURNOVER: 'exportsTurnoverPercentage,
   },
 };
--- a/e2e-tests/constants/field-ids/insurance/exporter-business/index.js
+++ b/e2e-tests/constants/field-ids/insurance/exporter-business/index.js
@@ -25,6 +25,6 @@ export const EXPORTER_BUSINESS = {
   TURNOVER: {
     FINANCIAL_YEAR_END_DATE: 'financialYearEndDate',
     ESTIMATED_ANNUAL_TURNOVER: 'estimatedAnnualTurnover',
-    PERCENTAGE_TURNOVER: 'exportsTurnoverPercentage,
+    PERCENTAGE_TURNOVER: 'exportsTurnoverPercentage',
   },
 };
--- a/e2e-tests/constants/field-ids/insurance/exporter-business/index.js
+++ b/e2e-tests/constants/field-ids/insurance/exporter-business/index.js
@@ -25,5 +25,6 @@ export const EXPORTER_BUSINESS = {
   TURNOVER: {
     FINANCIAL_YEAR_END_DATE: 'financialYearEndDate',
     ESTIMATED_ANNUAL_TURNOVER: 'estimatedAnnualTurnover',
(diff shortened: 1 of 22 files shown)


e2e-tests/constants/field-ids/insurance/exporter-business/index.js
```diff
@@ -25,5 +25,6 @@ export const EXPORTER_BUSINESS = {
   TURNOVER: {
     FINANCIAL_YEAR_END_DATE: 'financialYearEndDate',
     ESTIMATED_ANNUAL_TURNOVER: 'estimatedAnnualTurnover',
+    PERCENTAGE_TURNOVER: 'percentageTurnover',
```
ttbarnes, line 28: Since the description of this field is "Percentage of turnover from exports", shall we call this

`exportsTurnoverPercentage`?

e2e-tests/cypress/e2e/journeys/insurance/your-business/turnover/validation/turnover-percentage-turnover-validation.spec.js
```diff
@@ -0,0 +1,202 @@
...
+      cy.navigateToUrl(url);
+
+      const fieldId = PERCENTAGE_TURNOVER;
+      const field = turnover[fieldId];
+
+      field.input().clear().type('0');
+      submitButton().click();
+      partials.errorSummaryListItems().should('have.length', 1);
+    });
+  });
```
ttbarnes, line 201: What do you think about using a shared assertion function? Each scenario/test is very repetitive, had the same with password validation: https://github.com/UK-Export-Finance/exip/blob/main-application/e2e-tests/cypress/e2e/journeys/insurance/account/create/your-details/account-create-your-details-validation-password.spec.js#L27

Could do this for kinds of test for "your business" after this PR as a tech improvement

src/ui/server/controllers/insurance/business/turnover/validation/rules/percentage-of-turnover.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      expect(response).toEqual(expected);
+    });
+  });
+
+  describe(`when the ${PERCENTAGE_TURNOVER} input is above 100`, () => {
+    it('should return a validation error', () => {
+      mockBody[PERCENTAGE_TURNOVER] = '125';
+      const response = percentageTurnover(mockBody, mockErrors);
+
+      const errorMessage = EXPORTER_BUSINESS[PERCENTAGE_TURNOVER].ABOVE_MAXIMUM;
```
ttbarnes, line 99: Can we get the error message like in the actual validation, shorter lines then, so it can be

```js
const errorMessage = ERROR_MESSAGE.ABOVE_MAXIMUM
```

src/ui/server/helpers/number/index.test.ts
```diff
@@ -78,4 +78,33 @@ describe('server/helpers/number', () => {
       expect(result).toEqual(true);
     });
   });
+
+  describe('isNumberAboveMaximum', () => {
+    it('should return true if value is above maximum', () => {
+      const mockValue = 120;
```
ttbarnes, line 84: `101`

e2e-tests/cypress/support/submit-and-assert-errors.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * @param {String} field
+ * @param {String} fieldValue - the value to input - can be null
+ * @param {Number} errorIndex - index of error in errorSummary
+ * @param {Number} errorSummaryLength - the number of expected errors in errorSummary
+ * @param {String} errorMessage
+ */
+export default (field, fieldValue, errorIndex, errorSummaryLength, errorMessage) => {
+  // if fieldValue (null if want to test empty input)
```
ttbarnes, line 12: Could we say "only type if a field value is provided"? Found this a bit confusing

PR #292, opened 2023-02-03
feat(EMS-794): Account creation - save in the database

ttbarnes:
This PR introduces the ability to save account data in the database.

## Changes
- Create new `Exporter` keystone schema definition.
- Introduce API password constants to define salt/hash requirements, aligned with DTFS.
- Create a new custom mutation resolver, `createAccount`. This mutation:
  - Creates a salt and hash for the provided password.
  - Saves the account with salt and hash to the database via keystone context API.
  - Keystone context API then triggers the `resolveInput` hook, where we add some timestamps and send a "confirm email" to the provided email, via Notify API.
- In the UI:
  - Add a new GraphQL call - `keystone.account.create(...)`.
  - Update "your details" POST controller to trigger the new GraphQL call via new "save data" function.
  - When submitting the "your details" form add the submitted email to `req.session.emailAddressToConfirm`.
  - When redirected to the next page "confirm email", remove `emailAddressToConfirm` from the session, store it in a local variable and display in the template.
  - Add new mock account to test-mocks with the email coming from new environment variable `GOV_NOTIFY_EMAIL_RECIPIENT`.
  - Add `GOV_NOTIFY_EMAIL_RECIPIENT` environment variable to GHA workflow file, cypress environment config and E2E .env.sample (for running locally).

## Notes

- We have a placeholder, passed to the Notify call - "mockConfirmToken". A real token will be generated in future PR.


e2e-tests/cypress/fixtures/account.js
```diff
@@ -11,7 +11,7 @@ const {
 const account = {
   [FIRST_NAME]: 'Joe',
   [LAST_NAME]: 'Bloggs',
-  [EMAIL]: 'mock@email.com',
+  [EMAIL]: 'Tony.Barnes@ukexportfinance.gov.uk',
```
Zainzzkk, line 14: can we use mock emails in tests? @abhi-markan do you agree?

abhi-markan: Yes please, no PII only mock emails, however we can use `ukexportfinance.gov.uk` domain if you wish.

src/api/schema.ts
```diff
@@ -38,6 +40,8 @@ export const lists = {
       resolveInput: async ({ operation, resolvedData, context }) => {
         if (operation === 'create') {
           try {
+            // const emailResponse = await notify.sendEmail(EMAIL_TEMPLATE_IDS.ACCOUNT.CONFIRM_EMAIL, 'Tony.Barnes@ukexportfinance.gov.uk');
```
Zainzzkk, line 43: can this be removed or TODO added for removal/change

src/ui/server/controllers/insurance/account/create/confirm-email/index.ts
```diff
@@ -14,10 +14,16 @@ export const PAGE_CONTENT_STRINGS = PAGES.INSURANCE.ACCOUNT.CREATE.CONFIRM_EMAIL
  * @param {Express.Response} Express response
  * @returns {Express.Response.render} Confirm email page
  */
-export const get = (req: Request, res: Response) =>
-  res.render(TEMPLATE, {
+export const get = (req: Request, res: Response) => {
+  const email = req.session.emailAddressToConfirm;
```
Zainzzkk, line 18: can we destructure session since being used more than once

e2e-tests/cypress/fixtures/account.js
ttbarnes: We're waiting for `exporter1@ukexport...` emails to be setup from Risual which will take a while. In the meantime we need to use an email that's in the Notify whitelist - gone with this for now. Happy to change it just _needs_ to be in the Notify whitelist.

src/api/schema.ts
ttbarnes: Removed sorry, typo during a POC.

src/api/.keystone/config.js
```diff
@@ -60,6 +60,45 @@ var APPLICATION = {
...
+  STRING_TYPE: "hex",
+  PBKDF2: {
+    ITERATIONS: 1e4,
+    KEY_LENGTH: 64,
+    DIGEST_ALGORITHM: "sha512"
+  }
+};
+var EMAIL_TEMPLATE_IDS = {
+  ACCOUNT: {
+    CONFIRM_EMAIL: "24022e94-171c-4044-b0ee-d22418116575"
```
abhi-markan, line 74: Should we not have centralised EMAIL template IDs please?

src/ui/server/controllers/insurance/account/create/confirm-email/index.ts
ttbarnes: I've added destructuring for this part, but we can't destructure in the second line because it's a delete call, needs to reference it without destructuring.

abhi-markan (changes requested):
Few minor comments to address please.

src/api/.keystone/config.js
ttbarnes: We have centralised `api/constants.ts` -   this file, `.keystone/config.js` is auto generated by keystone from build script, have no control over this file.

e2e-tests/cypress/fixtures/account.js
IgnatG: @ttbarnes  can we pass this as an environment variable instead? - so that the username is not here?

ttbarnes: Thanks guys great stuff - created a new environment variable `GOV_NOTIFY_EMAIL_RECIPIENT`. 

We have to use an email that is on the GOV notify whitelist, which will be a specific ukexportfinance.gov.uk domain. But its anonymised now via environment variable 😄

abhi-markan (dismissed):

Zainzzkk (approved):
LGTM

PR #298, opened 2023-02-06
fix(EMS-946-947): Nature of Business - Mapping fields which are 0

Zainzzkk:
# Issue: 
* 0s were not displayed on the page when saved to the database

# Changes made:
* Added mapping function to convert 0 to strings
* Added tests


src/ui/server/helpers/display-number-field-value/index.ts
```diff
@@ -0,0 +1,15 @@
...
+ * if number is 0, then will make 0 a string so can be displayed
+ * @param {Number} value
+ * @returns value - either number or string or null
+ */
+const displayNumberFieldValue = (value?: number) => {
+  if (value === 0) {
+    return String(value);
+  }
+
+  return value;
```
ttbarnes, line 12: Should we call this `transformNumberToString`, and always return a string? Can also then add a return type of `:string` to the function on line 7

src/ui/server/helpers/display-number-field-value/index.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  describe('when provided value is null', () => {
+    it('should return null', () => {
+      const response = displayNumberFieldValue();
+
+      expect(response).toBeUndefined();
+    });
+  });
+
+  describe('when provided value is a number', () => {
+    it('should return null', () => {
```
abhi-markan, line 13: Incorrect description.

src/ui/server/helpers/display-number-field-value/index.test.ts
```diff
@@ -0,0 +1,27 @@
...
+      const response = displayNumberFieldValue(5);
+
+      expect(response).toEqual(5);
+    });
+  });
+
+  describe('when provided value is 0', () => {
+    it('should return null', () => {
```
abhi-markan, line 21: Incorrect description.

src/ui/server/helpers/mappings/map-application-to-form-fields/index.test.ts
```diff
@@ -52,6 +54,52 @@ describe('server/helpers/mappings/map-application-to-form-fields', () => {
     });
   });
 
+  describe('when an application has exporterBusiness fields', () => {
+    it('should return the relevant exporterBusiness fields as timestamps', () => {
```
ttbarnes, line 58: There's no timestamps in `exporterBusiness`? Can just remove the "as timestamps" part

src/ui/server/helpers/display-number-field-value/index.ts
Zainzzkk: the return otherwise is null or number btw

src/ui/server/helpers/mappings/map-application-to-form-fields/index.test.ts
```diff
@@ -52,6 +54,52 @@ describe('server/helpers/mappings/map-application-to-form-fields', () => {
...
+          [EMPLOYEES_UK]: displayNumberFieldValue(mockApplication.exporterBusiness[EMPLOYEES_UK]),
+          [EMPLOYEES_INTERNATIONAL]: displayNumberFieldValue(mockApplication.exporterBusiness[EMPLOYEES_INTERNATIONAL]),
+        },
+      };
+
+      expect(result).toEqual(expected);
+    });
+  });
+
+  describe('when an application has exporterBusiness fields set to 0', () => {
```
ttbarnes, line 76: We don't need this test, in here we only care about calling `displayNumberFieldValue`/`transformNumberToString` - those functions have their own tests for the 0 scenario

src/ui/server/helpers/display-number-field-value/index.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * helper to return number field value for mapping
+ * if number is 0, then will make 0 a string so can be displayed
+ * @param {Number} value
+ * @returns value - either number or string or null
+ */
+const displayNumberFieldValue = (value?: number) => {
+  if (value === 0) {
```
abhi-markan, line 8: Can we simplify this please.

```
value === 0 ? String(value) : value

src/ui/server/helpers/display-number-field-value/index.test.ts
ttbarnes, line 21: typo

abhi-markan (changes requested):
Few comments to address please.

src/ui/server/helpers/display-number-field-value/index.ts
Zainzzkk: fails on eslint

ttbarnes (approved):

abhi-markan (approved):

PR #306, opened 2023-02-08
feat(EMS-861): Account sign in - page setup, field validation

ttbarnes:
This PR sets up the "sign in" page, complete with field validation.

Note: there is no actual sign in functionality, this is purely setting up the page and core form validation.

## Changes
- Setup routes, controller and nunjucks template for the new page.
- Extract the content strings for email field into the top level of account, this is the same for account creation and sign in.
- Extract cypress selectors for email and password inputs into a shared partial - this is used for account creation and sign in.
- E2E test coverage.
- Add new cypress commands for:
  - Submitting eligibility and starting the account sign in process.
  - Completing and submitting the sign in form.

## Other improvements
- Add failsafe empty field check to shared email and password validation.
- Fixed some typos: duplicate "govuk-link" classes, align taskList cypress partial file name with other partials.


e2e-tests/cypress/e2e/journeys/insurance/account/sign-in/account-sign-in.spec.js
```diff
@@ -0,0 +1,159 @@
...
+
+  beforeEach(() => {
+    Cypress.Cookies.preserveOnce('_csrf');
+    Cypress.Cookies.preserveOnce('connect.sid');
+  });
+
+  it('passes the audits', () => {
+    cy.lighthouse({
+      accessibility: 100,
+      performance: 70,
```
abhi-markan, line 51: Is there a reason why `performance` has been set to `70`?

abhi-markan (changes requested):
LGTM, one minor comment to address please.

e2e-tests/cypress/e2e/journeys/insurance/account/sign-in/account-sign-in.spec.js
ttbarnes: Good shout @abhi-markan just bumped it to 75 like we have in the majority of places - 70 was a copy/paste from account creation which is required.

It's an interesting one because the performance is higher when running locally, but in GHA apparently page load speed takes a little bit longer, so it affects the performance audit check.

75 seems to be fail safe option for GHA - tried having it higher before, but it can be flaky and alter in different PRs.

ttbarnes: Actually sorry - it needs to be 70 like the "your details" page. This is because some client side MOJ javascript is run  for the password reveal functionality.

e2e-tests/cypress/e2e/journeys/insurance/account/create/your-details/account-create-your-details-validation.spec.js
```diff
@@ -97,7 +98,7 @@ context('Insurance - Account - Create - Your details page - form validation', ()
       yourDetailsPage[LAST_NAME].input().should('have.focus');
 
       partials.errorSummaryListItemLinks().eq(2).click();
-      yourDetailsPage[EMAIL].input().should('have.focus');
+      accountFormFields[EMAIL].input().should('have.focus');
```
Zainzzkk, line 101: can we use my cy validation command here?

src/ui/server/content-strings/error-messages.ts
```diff
@@ -209,6 +209,14 @@ export const ERROR_MESSAGES = {
           },
         },
       },
+      SIGN_IN: {
+        [FIELD_IDS.INSURANCE.ACCOUNT.EMAIL]: {
+          INCORRECT: "Email address - either the email address or password you've entered is incorrect",
+        },
+        [FIELD_IDS.INSURANCE.ACCOUNT.PASSWORD]: {
+          INCORRECT: "Password - either the email address or password you've entered is incorrect",
```
Zainzzkk, line 217: is there a reason for the "Password - " bit?

ttbarnes: Design. I presume it's because in the summary, without this, it doesn't have the context of what field as both fields have the same error message

src/ui/server/controllers/insurance/account/sign-in/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * post
+ * Check Sign in validation errors and if successful, redirect to the next part of the flow.
+ * @param {Express.Request} Express request
+ * @param {Express.Response} Express response
+ * @returns {Express.Response.redirect} Next part of the flow or error page
+ */
+export const post = (req: Request, res: Response) => {
+  const validationErrors = generateValidationErrors(req.body);
```
Zainzzkk, line 66: are we sanitising the body anywhere?

e2e-tests/cypress/e2e/journeys/insurance/account/create/your-details/account-create-your-details-validation.spec.js
ttbarnes: Great shout - updated

PR #308, opened 2023-02-10
feat(EMS-660): Company Details - Broker - Using a broker contact details

Zainzzkk:
# Changes made:
* Added conditional fields for broker yes option
* Added shared address input nunjucks file
* Added conditional HTML rendering in yes-no-radios shared nunjucks file
* Added validation for fields
* Added extra step in validation to check for broker input being yes
* Added unit and e2e tests


e2e-tests/constants/field-ids/insurance/exporter-business/index.js
```diff
@@ -28,6 +28,14 @@ export const EXPORTER_BUSINESS = {
     PERCENTAGE_TURNOVER: 'exportsTurnoverPercentage',
   },
   BROKER: {
+    BROKER_HEADING: 'broker',
```
ttbarnes, line 31: I don't think should have `_HEADING` here as this is the actual radio input right? `isUsingBroker`?

Maybe the parent object should be `USING_BROKER` so we don't have `BROKER.BROKER`

Thoughts?

e2e-tests/constants/field-ids/insurance/exporter-business/index.js
```diff
@@ -28,6 +28,14 @@ export const EXPORTER_BUSINESS = {
...
     USING_BROKER: 'isUsingBroker',
+    BROKER_NAME: 'brokerName',
+    BROKER_ADDRESS_LINE_1: 'brokerAddressLineOne',
+    BROKER_ADDRESS_LINE_2: 'brokerAddressLineTwo',
+    BROKER_TOWN: 'brokerTown',
+    BROKER_COUNTY: 'brokerCounty',
+    BROKER_POSTCODE: 'brokerPostcode',
+    BROKER_EMAIL: 'brokerEmail',
```
ttbarnes, line 39: Can we have the fields without broker prefix? `addressLineOne, town, email`?

src/ui/server/controllers/insurance/business/broker/index.test.ts
```diff
@@ -7,8 +7,14 @@ import insuranceCorePageVariables from '../../../../helpers/page-variables/core/
...
 import { mockReq, mockRes, mockApplication } from '../../../../test-mocks';
 import generateValidationErrors from './validation';
+import { FIELDS } from '../../../../content-strings/fields/insurance/your-business';
 
-const { USING_BROKER } = FIELD_IDS.BROKER;
+const { BROKER: BROKER_FIELDS } = FIELDS;
+
+// below will be removed on next PR, linting not working for this line
+// eslint-disable-next-line operator-linebreak
+const { USING_BROKER, BROKER_HEADING, BROKER_NAME, BROKER_ADDRESS_LINE_1, BROKER_ADDRESS_LINE_2, BROKER_COUNTY, BROKER_TOWN, BROKER_POSTCODE } =
```
ttbarnes, line 16: Can just add new line break to each destructure?

src/ui/templates/components/address-input.njk
```diff
@@ -0,0 +1,128 @@
...
+
+  {% set ADDRESS_LINE_ONE_FIELD = params.addressLineOne %}
+  {% set ADDRESS_LINE_TWO_FIELD = params.addressLineTwo %}
+  {% set COUNTY_FIELD = params.country %}
+  {% set TOWN_FIELD = params.town %}
+  {% set POSTCODE_FIELD = params.postcode %}
+  {% set validationErrors = params.validationErrors %}
+  {% set application = params.application %}
+  {% set submittedValues = params.submittedValues %}
+
```
ttbarnes, line 13: Could we call this component `address-fields`?

Then i'm thinking we could be more DRY with each field, they all look the same with just a different ID?f

Could have a new component/macro `address-field-input` which renders a `govukInput` like below, but passing an ID, then below could just have e.g

```js
{{ addressFieldInput.render({
  id: ADDRESS_LINE_ONE_FIELD
}) }}
```

src/ui/server/controllers/insurance/business/broker/index.test.ts
Zainzzkk: eslint complains about that too

src/ui/templates/components/yes-no-radio-buttons.njk
```diff
@@ -29,9 +31,12 @@
     {% set noValue = 'No' %}
   {% endif %}
 
+  {% if not params.notInlineRadios %}
```
ttbarnes, line 34: Can we call this `horizontalRadios` so it's inline with GOV design descriptions?

src/ui/server/controllers/insurance/business/broker/index.test.ts
ttbarnes: Interesting - I think if we rename like above comment so we don't have `BROKER_` prefix, eslint will be OK?

src/ui/templates/components/address-field-input.njk
```diff
@@ -0,0 +1,33 @@
+{% from "govuk/components/input/macro.njk" import govukInput %}
+
+{% macro render(params) %}
+
+  {% set FIELD = params.field %}
```
ttbarnes, line 5: Can we pass in `id` and `label` specifically instead?

ttbarnes (approved):

PR #310, opened 2023-02-10
feat(EMS-801): Account creation - verification

ttbarnes:
This PR adds the ability for an Exporter to verify their account by clicking an link (or manually navigating to a URL).

## User flow

1) User submits the "create an account" form.
2) We generate a verification token (valid for 24 hours).
3) User receives an email with a "verify your email" link and with token param.
4) User goes to this "verify your email" URL.
5) We then check if there is an Exporter with this matching verification token, check if it's still valid (less than 24 hours old) and if so, mark the Exporter as verified.
6) User is then directed to the "sign in" page. Otherwise, "link expired" page, or if everything fails, "problem with service".

### UI and API
- Update `createAccount` GraphQL resolver so that we addd a `verificationHash` and `verificationExpiry` to new accounts.
- Add custom resolver to the API, `verifyAccountEmailAddress`. This resolver:
  - Checks if an exporter exists with the provided token.
  - Checks if the token has not gone past the expiry date (24 hours).
  - If the above checks are OK, mark the exporter as verified and nullify the verification hash and expiry field.
- Update the "create" `resolveInput` keystone hook for Account/Exporter so that we pass the actual verification hash to the Notify verification email, instead of the previous mock token.
- Add "verify email" UI route, and controller that calls a new GQL call, `verifyEmailAddress`. Passing the verification token from the URL.
- Change API port from default 3000 to 5001 for when running outside of docker.

### E2E tests
- Add API support to E2E test suite:
  - `apiUrl` is now in the cypress config.
  - Create an API support file with `getExporterByEmail`.
  - This allows us to get an exporter from the API and therefore grab the `verificationHash`, so we can mimic "navigating to a URL with token param" - like an exporter would via clicking an email link.
- Add E2E test for creating an account, navigating to verify URL and then being redirect to the sign in page.
- Add `VERIFICATION_EXPIRY` to API constants. Move password constants into new `ACCOUNT` constant.

## Other improvements
- Add `express-rate-limit` to UI server config.
- Rename `isActive` exporter field to `isVerified`.
- Add unit test coverage for `createAccount` GraphQL resolver.
- Update the "update" `resolveInput` keystone hook for Account/Exporter so that anytime an update is made, the `updatedAt` timestamp is updated.


e2e-tests/cypress/e2e/journeys/insurance/account/create/create-account-and-verify-email/create-account-and-verify-email.spec.js
```diff
@@ -0,0 +1,55 @@
...
+       * so that we can mimic "clicking email verification link" and proceed with the user flow.
+       */
+      api.getExporterByEmail(exporterEmail).then((response) => {
+        exporter = response;
+      });
+    });
+
+    it(`should validate the exporter and redirect to ${SIGN_IN.ROOT}`, () => {
+      /**
+       * Mimic "clicking email verification link"
```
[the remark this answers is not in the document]
ttbarnes: Makes sense, i've updated - will wait to push whilst waiting for reviews 😊

src/api/constants.ts
```diff
@@ -20,13 +20,22 @@ export const APPLICATION = {
...
-  PBKDF2: {
-    ITERATIONS: 10000,
-    KEY_LENGTH: 64,
-    DIGEST_ALGORITHM: 'sha512',
+export const ACCOUNT = {
+  EMAIL: {
+    VERIFICATION_EXPIRY: () => {
+      // 24 hours
+      const date = new Date();
+      return new Date(date.setDate(date.getDate() + 1));
```
Zainzzkk, line 28: this is quite hard to follow, is it worth doing setDate in another const and then passing it?

ttbarnes: I've split this up more should be easy to read now

src/ui/server/index.ts
```diff
@@ -35,6 +36,15 @@ app.use(seo);
 app.use(security);
 app.use(compression());
 
+const limiter = rateLimit({
+  windowMs: 1 * 60 * 1000, // 1 minute
+  max: 1000, // 1K requests / 1 window
```
Zainzzkk, line 41: can we put this and above in constants?

src/ui/server/routes/insurance/account/index.ts
```diff
@@ -12,6 +13,8 @@
 
 insuranceAccountRouter.get(INSURANCE_ROUTES.ACCOUNT.CREATE.CONFIRM_EMAIL, confirmEmailGet);
 
+insuranceAccountRouter.get(INSURANCE_ROUTES.ACCOUNT.CREATE.VERIFY_EMAIL, verifyEmailGet);
```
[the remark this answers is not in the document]
Zainzzkk: @abhi-markan does this need to be actioned?

src/ui/server/controllers/insurance/account/create/verify-email/index.ts
```diff
@@ -0,0 +1,39 @@
...
+      const response = await api.keystone.account.verifyEmailAddress(token);
+
+      if (response.success) {
+        return res.redirect(SIGN_IN.ROOT);
+      }
+
+      return res.redirect(CREATE.VERIFY_EMAIL_LINK_EXPIRED);
+    }
+
+    return res.redirect(CREATE.VERIFY_EMAIL_LINK_EXPIRED);
```
Zainzzkk, line 33: if there's no token, would it still be expired or problem with service?

ttbarnes: Good spot, I had a good old think about this. If no token is provided, ideally I think from a user/design perspective we would redirect to a "no token provided" or "invalid/broken URL, try ...." kind of page.

Redirecting to "problem with service" would be misleading because there isn't actually a problem, just the token/URL sent by the user is invalid (or doesn't exist).

I think this is an ok solution for MVP unless anyone else thinks otherwise. i'll raise with design for post MVP improvement.

src/ui/server/routes/insurance/account/index.ts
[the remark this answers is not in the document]
ttbarnes: @Zainzzkk @abhi-markan This is now outdated and fully resolved after a new push, by introducing `express-rate-limit` package 🎉 marking as resolved

src/ui/server/controllers/insurance/account/create/verify-email/index.ts
ttbarnes: Discussed with design, this is all fine for MVP, design will look into better messaging/pages for these scenarios for post MVP 💪

src/ui/server/index.ts
ttbarnes: Could do - I think we should do everything in the UI server in one go though - can be future improvement? There's other instances (probably the same in DTFS), e.g for cookies, csrf etc

Zainzzkk (dismissed):
LGTM

src/api/constants.ts
```diff
@@ -20,13 +20,26 @@ export const APPLICATION = {
...
-    ITERATIONS: 10000,
-    KEY_LENGTH: 64,
-    DIGEST_ALGORITHM: 'sha512',
+export const ACCOUNT = {
+  EMAIL: {
+    VERIFICATION_EXPIRY: () => {
+      // 24 hours from now
+      const now = new Date();
+      const day = now.getDate();
+
```
abhi-markan, line 29: Do we need extra carriage return?

abhi-markan (changes requested):
Few minor changes to address please.

e2e-tests/cypress/support/api/index.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+const getExporterByEmail = async (email) => {
+  try {
+    const baseUrl = Cypress.config('apiUrl');
+    const url = `${baseUrl}?query=${queryStrings.getExporterByEmail(email)}`;
+
+    const response = await cy.request(url);
+
+    const { data } = response.body;
+
```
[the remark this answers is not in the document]
ttbarnes: The try/catch should be enough, but i've inside the `try { ...`, i've added an additional check for `response.body` and `.data` - if one of these doesn't exist it'll throw an error.

src/api/custom-resolvers/verify-account-email-address.ts
```diff
@@ -0,0 +1,72 @@
...
+        verificationHash: '',
+        verificationExpiry: null,
+      },
+    });
+
+    return {
+      success: true,
+    };
+  } catch (err) {
+    throw new Error(`Verifying exporter email address ${err}`);
```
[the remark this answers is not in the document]
ttbarnes: Good spot - added

src/api/constants.ts
ttbarnes: As discussed, yes - this is "constants function" for generating a date (of 24 hours from now) so we return the date so we can consume it e.g `const date = CONSTANTS.EMAIL.VERIFICATION_EXPIRY();`

abhi-markan (approved):

PR #311, opened 2023-02-10
feat(EMS-871): Account sign in - password reveal

ttbarnes:
This PR ensures that password reveal functionality is working for the Account sign in page.

## Changes
- Create a cypress command for checking password reveal functionality. This is now a feature used in 2 places and there will be more in the future.
- Call this new command in the "create account" and "sign in" E2E page tests.

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/account/sign-in/account-sign-in.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/account/sign-in/account-sign-in.spec.js
@@ -111,6 +111,12 @@ context('Insurance - Account - Sign in - I want to sign in into my UKEF digital 
 
       field.input().should('exist');
     });
+
+    describe('reveal button', () => {
+      it('should be rendered and show/reveal the password input', () => {
+        cy.assertPasswordRevealButton();
+      });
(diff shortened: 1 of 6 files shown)


e2e-tests/cypress/e2e/journeys/insurance/account/sign-in/account-sign-in.spec.js
```diff
@@ -111,6 +111,10 @@ context('Insurance - Account - Sign in - I want to sign in into my UKEF digital
 
       field.input().should('exist');
     });
+
+    it('should render a revreal button that shows/reveals the password input', () => {
```
IgnatG, line 115: 'revreal'?

IgnatG (approved):
Lgtm
I've added a small comment. Not a blocker though

e2e-tests/cypress/e2e/journeys/insurance/account/sign-in/account-sign-in.spec.js
ttbarnes: Whoops! Good spot - i'll fix this immediately in another branch after merge.

PR #317, opened 2023-02-14
feat(EMS-799): Account creation - resend verification email 

ttbarnes:
This PR adds the ability to resend an account verification email.

## Changes 

In summary - add a new route, controller and template for the "resend/resent confirmation email" page. This controller picks up the exporter ID from query params and calls a new GQL API call, `sendEmailConfirmEmailAddress` with the exporter ID. The API then fetches the exporter's verification hash and calls GOV Notify with the hash.

### API

- Add new custom resolver mutation, `sendEmailConfirmEmailAddress`.
- Add a new `emails` directory where we can store all "send email" functions moving forwards.
- Update GOV Notify "send email" function to always return `emailRecipient` as well as `success`.
- Rename `VerifyAccountEmailAddressResponse` custom GQL type to be a generic `EmailResponse`
- Create test mocks directory with a mock account used for unit tests.

### UI 

- Update the "request a new link" link, to point to "email resent" URL.
- Add new GQL API calls to get an account by ID and send the "confirm email address" email.
- Update "confirm email" controller to get an exporter's ID from the provided email. Pass the ID to the template so that it can be rendered in the "resend email" link.
- Create a new nunjucks partial `confirm-email-page-content`. This is used in both the "confirm email" and "confirm email resent" pages (both have exactly the same content).

### E2E 

- Create a new cypress command, `assertConfirmEmailPageContent`. This is used in both the "confirm email" and "confirm email resent" pages (both have exactly the same content).
- Add E2E test coverage for triggering "resend confirm email".

### Also

- Simplify the returns in UI API calls.
- Fixed `.gitignore` typo.
- Fixed some typos.
- Aligned some error logs.
- Updated some mocks.


src/api/emails/index.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const result = await sendEmail.confirmEmailAddress(email, firstName, verificationHash);
+
+    expect(sendEmailSpy).toHaveBeenCalledTimes(1);
+    expect(sendEmailSpy).toHaveBeenCalledWith(EMAIL_TEMPLATE_IDS.ACCOUNT.CONFIRM_EMAIL, email, firstName, verificationHash);
+
+    const expected = sendEmailResponse;
+
+    expect(result).toEqual(expected);
+  });
```
Zainzzkk, line 31: are there any tests for if it fails?

src/api/emails/index.ts
```diff
@@ -0,0 +1,32 @@
+import { EMAIL_TEMPLATE_IDS } from '../constants';
+import notify from '../integrations/notify';
+import { EmailResponse } from '../types';
+
+/**
+ * confirmEmailAddress
+ * @param {Object} Exporter account
```
Zainzzkk, line 7: incorrect params here I think

src/api/emails/index.test.ts
ttbarnes: I tried to add this - had a struggle testing "throw error" assertion/stubs etc. Will have another look, might want to do this separately

src/ui/server/controllers/insurance/account/create/confirm-email/index.ts
```diff
@@ -14,16 +15,31 @@ export const PAGE_CONTENT_STRINGS = PAGES.INSURANCE.ACCOUNT.CREATE.CONFIRM_EMAIL
...
-  return res.render(TEMPLATE, {
-    ...insuranceCorePageVariables({
-      PAGE_CONTENT_STRINGS,
-      BACK_LINK: req.headers.referer,
-    }),
-    email,
-  });
+export const get = async (req: Request, res: Response) => {
+  try {
+    const { accountIdToConfirm: exporterId } = req.session;
```
Zainzzkk, line 20: if there is no id, should it still show the template?

If not, then should we not add a if (!exporterId) then go to problem with service

src/ui/templates/partials/confirm-email-page-content.njk
```diff
@@ -0,0 +1,34 @@
+{% if exporterEmail %}
+    <p data-cy="we-sent-link-to">{{ CONTENT_STRINGS.WE_SENT_LINK_TO }} {{ exporterEmail }}</p>
```
Zainzzkk, line 2: indentation in this file seems to be off a little.  I think all indented when not in the if.  And line 3 should remove indentation

src/ui/server/controllers/insurance/account/create/confirm-email/index.ts
ttbarnes: It should still show the template - bit of a strange one. We temporarily store the ID in the session specifically for this page - then we remove it from the session before rendering the page. For security.

If a user refreshes this page, there will be no ID and therefore we can't display the email address. 

But we still want to say "sent you an email", just without the email address, instead of redirecting to an error page.

src/ui/templates/partials/confirm-email-page-content.njk
ttbarnes: Whoops, fixed

src/api/emails/index.test.ts
ttbarnes: Not sure what I was doing before but got it working 😄

src/api/.keystone/config.js
```diff
@@ -102,14 +102,36 @@ var notify = {
...
+    if (emailResponse.success) {
+      return emailResponse;
+    }
+    throw new Error(`Sending email verification for account creation ${emailResponse}`);
+  } catch (err) {
+    console.error(err);
+    throw new Error(`Sending email verification for account creation ${err}`);
+  }
+};
+var sendEmail = {
```
IgnatG, line 130: should this be `let`/`const`?

src/api/.keystone/config.js
```diff
@@ -548,6 +566,32 @@ var verifyAccountEmailAddress = async (root, variables, context) => {
 };
 var verify_account_email_address_default = verifyAccountEmailAddress;
 
+// custom-resolvers/send-email-confirm-email-address.ts
+var sendEmailConfirmEmailAddress = async (root, variables, context) => {
```
IgnatG, line 570: const/let?

src/api/.keystone/config.js
ttbarnes: This is auto generated by keystone from our typescript file

src/api/.keystone/config.js
ttbarnes: Same as above, anything insdie of `.keystone` directory is auto generated

src/ui/server/api/keystone/account/index.ts
```diff
@@ -33,6 +35,35 @@ const account = {
...
+      if (response.errors) {
+        console.error('GraphQL error getting exporter account ', response.errors);
+      }
+
+      if (response?.networkError?.result?.errors) {
+        console.error('GraphQL network error getting exporter account ', response.networkError.result.errors);
+      }
+
+      if (response?.data?.exporter) {
+        const { data } = response;
```
IgnatG, line 55:
```
const { data: { exporter} } = response;
return exporter;
```

or we return straight away the exporter?
```
return response.data.exporter;
```

src/ui/server/api/keystone/account/index.ts
```diff
@@ -62,6 +93,35 @@ const account = {
...
+      }
+
+      if (response?.networkError?.result?.errors) {
+        console.error('GraphQL network error sending email verification for account creation ', response.networkError.result.errors);
+      }
+
+      if (response?.data?.sendEmailConfirmEmailAddress) {
+        const { data } = response;
+
+        return data.sendEmailConfirmEmailAddress;
```
IgnatG, line 115: same as above

src/ui/server/api/keystone/account/index.ts
ttbarnes: Great shout @IgnatG i've updated all instances of this in the UI API calls 👍

e2e-tests/cypress/e2e/journeys/insurance/account/create/resend-confirm-email/resend-confirm-email.spec.js
```diff
@@ -0,0 +1,59 @@
...
+      /**
+       * Get the exporter ID directly from the API,
+       * so that we can assert that the URL and `request a new link` has the correct ID.
+       */
+      const exporterEmail = Cypress.env('GOV_NOTIFY_EMAIL_RECIPIENT');
+
+      api.getExporterByEmail(exporterEmail).then((response) => {
+        exporter = response;
+
+        expectedUrl = `${CONFIRM_EMAIL_RESENT}?id=${exporter.id}`;
```
abhi-markan, line 42: Can we please add void exporter check, in case api call fails.

src/api/emails/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ * confirmEmailAddress
+ * @param {String} Email address
+ * @param {String} First name
+ * @param {String} Verification hash
+ * @returns {Object} Object with success flag and emailRecipient
+ */
+const confirmEmailAddress = async (email: string, firstName: string, verificationHash: string): Promise<EmailResponse> => {
+  try {
+    console.info('Sending email verification for account creation');
+
```
abhi-markan, line 15: Should we add argument void check?

abhi-markan (changes requested):

e2e-tests/cypress/e2e/journeys/insurance/account/create/resend-confirm-email/resend-confirm-email.spec.js
ttbarnes: I don't think we need to, we handle error in `api.getExporterByEmail` - if there is API error/error calling the API, this function will throw an error and the E2E test will not continue.

As we progress the full application/accounts, want to avoid adding checks like this in actual tests as they'd be a lot of them

src/api/emails/index.ts
ttbarnes: Typescript does this for us 😊

abhi-markan (approved):

src/api/.keystone/config.js
ttbarnes: Resolving this so we can merge - I did try adding this to `.gitignore`, but turns out this is required for the build unfortunately.

PR #319, opened 2023-02-15
feat(EMS-802): account verification - sign in page success banner

ttbarnes:
This PR adds a success banner to the sign in page after an account has been verified.

## Changes
- Save a value in `req.flash` after account verification
- Update sign in controller to consume `req.flash` and pass a flag/variable to the template.
- Render success banner on the sign in page template if the flag is true.
- E2E test coverage


src/ui/server/controllers/insurance/account/sign-in/index.ts
```diff
@@ -46,14 +46,24 @@ export const PAGE_CONTENT_STRINGS = PAGES.INSURANCE.ACCOUNT.SIGN_IN.ROOT;
...
  * @returns {Express.Response.render} Sign in page
  */
-export const get = (req: Request, res: Response) =>
-  res.render(TEMPLATE, {
+export const get = (req: Request, res: Response) => {
+  const flash = req.flash('successBanner');
+
+  let renderSuccessBanner = false;
+
+  if (flash.includes('newAccountVerified')) {
```
IgnatG, line 54: very small comment:
this could be simplified into something like
```
let renderSuccessBanner = flash.includes('newAccountVerified') || false;
```

IgnatG (dismissed):
LGTM
I've added a comment, but it's not a blocker

src/ui/server/controllers/insurance/account/sign-in/index.ts
ttbarnes: Thanks @IgnatG - surprised I didn't think of that 🤦 will update this.

abhi-markan (dismissed):

Zainzzkk (approved):
LGTM

abhi-markan (approved):

PR #321, opened 2023-02-15
feat(EMS-870): account sign in - security code page setup

ttbarnes:
This PR sets up the "security code" page/form for the Account sign in flow.

Note: there is no form validation or actual security code functionality, this PR just sets up the GET route and renders the form.

## Changes

- Setup route, controller and nunjucks template.
- Add temporary POST route that redirects to dashboard route (this will be updated in a future PR for validation API response handling).
- Add E2E test coverage.

### DIFF
--- a/src/ui/server/controllers/insurance/account/sign-in/enter-code/index.ts
+++ b/src/ui/server/controllers/insurance/account/sign-in/enter-code/index.ts
@@ -1,0 +1,56 @@
+import { PAGES } from '../../../../../content-strings';
+import { FIELD_IDS, ROUTES, TEMPLATES } from '../../../../../constants';
+import { ACCOUNT_FIELDS as FIELDS } from '../../../../../content-strings/fields/insurance/account';
+import insuranceCorePageVariables from '../../../../../helpers/page-variables/core/insurance';
+import { Request, Response } from '../../../../../../types';
+
+const {
+  ACCOUNT: { SECURITY_CODE },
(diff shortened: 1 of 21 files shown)


src/ui/server/controllers/insurance/account/sign-in/enter-code/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  INSURANCE: { DASHBOARD },
+} = ROUTES;
+
+/**
+ * PAGE_VARIABLES
+ * Page fields
+ * @returns {Object} Page variables
+ */
+export const PAGE_VARIABLES = {
+  FIELDS: {
```
Zainzzkk, line 21: do we need to do this if there's only one field on the page?

ttbarnes: Good spot, I did have a go with single input variables function, had some troubles would have entailed a lot of changes due to the way the function gets field strings with the current strings structure (not all insurance fields are available at the root).

Will have another look, have another idea of just using core page variables or tweaking the field string imports

ttbarnes: Updated this so there is no longer `FIELDS` and just `FIELD`. Tried to use single input variables function, but it entails a lot of work. We will probably have same/similar requirement when it comes to declarations, can look at the function more then.

Zainzzkk (approved):
LGTM

PR #324, opened 2023-02-17
feat(EMS-870): Account sign in - API validation, OTP/security code generation

ttbarnes:
This PR adds the ability to verify exporter account sign in credentials against the database and then generate a OTP/security code.

## Changes

- New custom resolver/mutation: `accountSignIn`. This mutation:
  - Gets the account from DB.
  - Checks if the password is valid.
  - Generates a OTP/security code with `otplib` package.
  - Sends an email with OTP/security code.
- Add new function to send "here is your security code" email.
- UI: Add new GQL api call for sign in.
- UI: Update the sign in controller to call sign in GQL call. Add a temporary redirect to `invalid-credentials` (this will be replaced with a re-render in a future PR).
- E2E: 
  - Create cypress commands `getExporterByEmail`, and `verifyAccountEmail`.
  - Update some tests so that sign in functionality is using a verified account (happy path).

## Other improvements
- Extract `createAccount` mutation into it's own file instead of being inside custom-schemaa
- Extract "get an account by X field" function into it's own helper file.
- Update `verifyAccountEmailAddress` resolver/mutation to consume new "get an account by X field" function, simplifying the function, DRY.
- Rename `PASSWORD` constants to `ENCRYPTION` since it's now used in multiple places and it's not exclusive to "account password".


src/api/.keystone/config.js
```diff
@@ -69,19 +69,29 @@ var ACCOUNT = {
...
+  ENCRYPTION: {
     RANDOM_BYTES_SIZE: 32,
     STRING_TYPE: "hex",
     PBKDF2: {
       ITERATIONS: 1e4,
       KEY_LENGTH: 64,
       DIGEST_ALGORITHM: "sha512"
     }
+  },
+  OTP: {
```
ttbarnes, line 81: ⚠️ NOTE: this file is auto generated/compiled by keystone.

e2e-tests/cypress/support/insurance/account/verify-account-email.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * verifyAccountEmail
+ * Get the exporter (with verification hash) directly from the API,
+ * Mimic "cliking email verification link" in an email inbox by manually navigating to the URL
+ */
+const verifyAccountEmail = () => {
+  // get exporter
+  cy.getExporterByEmail(exporterEmail).then((response) => {
+    const { data } = response.body;
```
abhi-markan, line 17: Should we consider `response` and `data` void check, otherwise please wrap the in try-catch.

e2e-tests/cypress/support/insurance/account/get-exporter-by-email.js
```diff
@@ -0,0 +1,23 @@
+import api from '../../api';
+
+/**
+ * getExporterByEmail
+ * @param {String} Exporter account email
+ * Get an exporter directly from the API,
+ * @returns {Object} Exporter account
+ */
+const getExporterByEmail = (email) => {
+  api.getExporterByEmail(email).then((response) => {
```
abhi-markan, line 10: Wrap them in try-catch please.

src/api/helpers/generate-otp.ts
```diff
@@ -0,0 +1,50 @@
...
+    const expiry = OTP.VERIFICATION_EXPIRY();
+
+    return {
+      securityCode,
+      salt,
+      hash,
+      expiry,
+    };
+  } catch (err) {
+    throw new Error(`Error generating OTP ${err}`);
```
abhi-markan, line 42: Please add console.error.

src/api/helpers/is-valid-account-password.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      const result = isValidAccountPassword(mockPassword, salt, hash);
+
+      expect(result).toEqual(true);
+    });
+  });
+
+  describe('when password is invalid', () => {
+    it('should return false', () => {
+      const result = isValidAccountPassword('IncorrectPassword123!', salt, hash);
+
```
abhi-markan, line 37: Please add a test case where password provided in empty or will this be prevented by TS and UI validation?

src/ui/server/controllers/insurance/account/sign-in/index.ts
```diff
@@ -83,6 +84,23 @@ export const post = (req: Request, res: Response) => {
     });
   }
 
-  // TODO: check sign in details are correct and generate a code.
-  return res.redirect(ENTER_CODE);
+  try {
+    // validate credentials
```
abhi-markan, line 88: Can we please add `req.body` void check.

abhi-markan (changes requested):
Great work, few comments to address please.

src/api/helpers/is-valid-account-password.test.ts
ttbarnes: This will be prevented by TS on the API and UI level - the functions cannot be called without all parameters being provided.

src/ui/server/controllers/insurance/account/sign-in/index.ts
ttbarnes: This is done by TS, a `req` will always have a body

e2e-tests/cypress/support/insurance/account/verify-account-email.js
ttbarnes: response/data check is handled by `cy.getExporterByEmail` so no need for it here as well - i've added try/catch to both.

abhi-markan (approved):

PR #338, opened 2023-02-24
feat(EMS-900): Account sign in - verify OTP and create user session

ttbarnes:
This PR adds the ability for a user to verify their OTP sign in code and create a new user session.

### API
- Add new fields to user/exporter schema: `sessionExpiry`, `sessionIdentifer`.
- Add constants for `JWT` generation/encoding and session expiry.
- Update account sign in GQL resolver to return the `accountId`. This is passed to the UI in order to call the API with a OTP/security code and an account ID.
- Create a new helper function, `is-valid-otp`.
- Create a new helper function, `create-jwt`.
- Creata a new helper function, `generateOTPAndUpdateAccount` to be DRY. Update resolvers to use this function.
- Create new GQL resolver, `verifyAccountSignInCode`. This gets an exporter by account ID and then checks if the provided OTP/security code is correct. If so, create a JWT/session and update in the DB.

### UI 
- Update "enter security code" page/route/form to:
  - Call the API to validate and log the user in, `verifyAccountSignInCode`. if all is OK, redirect to the dashboard.
  - Throw a validation error if the provided code is invalid, invalid format, empty input etc).
- Create initial dashboard route, controller and page (currently just renders a simple page title).

### E2E
- Added E2E test coverage for:
  - Validating the  OTP/security code when signing in.
- Add apollo client setup to E2E tests so that we can make GQL queries in cypress tests.
- Create new GQL resolver, `addAndGetOtp`, this is required purely for E2E tests:
  - This resolver generates a OTP, but returns it in the response.
  - This way, we can test a real world scenario in E2E tests, of entering a valid or invalid security code.
  - The alternative to this is to have mailbox intergation inside E2E tests (risky/flaky), or we stub the GET/POST calls - which is then not reflective of a real world scenario.

### Also 
- Fix some typos.
- Bumped keystone deps.

## Note

This does not complete account sign. Next PRs will be:
- Checking an account session is valid and not expired to all appropriate routes, updating all tests.
- Checking the given account session has access to the given application.


e2e-tests/cypress/e2e/journeys/insurance/account/sign-in/enter-code/validation/account-sign-in-enter-code-validation.spec.js
```diff
@@ -0,0 +1,96 @@
...
+         * This is to ensure that we are testing a real world scenario.
+         *
+         * The alternative approach is to either intercept the UI requests and fake the security code validation,
+         * or have email inbox testing capabilities which can be risky/flaky.
+         * This approach practically mimics "get my security code from my email inbox".
+         */
+        const exporterEmail = Cypress.env('GOV_NOTIFY_EMAIL_RECIPIENT');
+
+        api.addAndGetOtp(exporterEmail).then((validSecurityCode) => {
+          enterCodePage[SECURITY_CODE].input().clear().type(validSecurityCode);
```
Zainzzkk, line 85: cy.keyboardInput() please

src/api/helpers/create-jwt.ts
```diff
@@ -0,0 +1,39 @@
...
+  ENCRYPTION: { RANDOM_BYTES_SIZE, STRING_TYPE },
+  JWT: {
+    KEY: { SIGNATURE, ENCODING, STRING_ENCODING },
+    TOKEN: { EXPIRY },
+  },
+} = ACCOUNT;
+
+const PRIV_KEY = Buffer.from(SIGNATURE, ENCODING).toString(STRING_ENCODING);
+
+const createJWT = (accountId: string) => {
```
Zainzzkk, line 15: documentation please

src/api/helpers/create-jwt.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const expiresIn = EXPIRY;
+
+  const payload = {
+    sub: accountId,
+    iat: Date.now(),
+    sessionIdentifier,
+  };
+
+  const signedToken = jsonwebtoken.sign(payload, PRIV_KEY, { expiresIn, algorithm: 'RS256' });
```
Zainzzkk, line 26: RS256 in constants?

src/ui/server/controllers/insurance/account/sign-in/enter-code/index.ts
```diff
@@ -35,20 +43,80 @@ export const PAGE_CONTENT_STRINGS = PAGES.INSURANCE.ACCOUNT.SIGN_IN.ENTER_CODE;
...
+        lastName: response.lastName,
+        token: response.token,
+        expires: response.expires,
+      };
+
+      return res.redirect(DASHBOARD);
+    }
+
+    // incorrect sign in code - force validation errors by mimicking empty form submission)
+    validationErrors = securityCodeValidationErrors({}, {});
```
Zainzzkk, line 107: rather than passing empty objects, could we not have a check in `securityCodeValidationErrors` to see if they are passed and if not, return an error

ttbarnes: I can't see a better way to do this - in this scenario, the code has been passed in the `req.body`, but the API says it's invalid (no `response.success`). So we need to render the validation error - but the validation rules only have "if a code is not submitted" checks - which is all the UI should care about.

I think the only other way to do this would be to pass a new flag to the validation rule e.g `forceError=true`, but I think passing an empty object is cleaner. What do you think?

src/api/.keystone/config.js
```diff
@@ -95,6 +97,23 @@ var ACCOUNT = {
...
+      STRING_ENCODING: "ascii"
+    },
+    TOKEN: {
+      EXPIRY: "8h",
+      ALGORITHM: "RS256"
+    },
+    SESSION_EXPIRY: () => {
+      const now = new Date();
+      const hours = 8;
+      const future = new Date(now.getTime() + hours * 60 * 60 * 1e3);
```
abhi-markan, line 114: Can the number of second be loaded from a constant please?

src/api/.keystone/config.js
```diff
@@ -796,6 +830,32 @@ var generate = {
 };
 var generate_otp_default = generate;
 
+// helpers/generate-otp-and-update-account.ts
+var generateOTPAndUpdateAccount = async (context, accountId) => {
+  try {
+    console.info("Adding OTP to exporter account");
```
abhi-markan, line 836: Do we need `accountId` validation or is it already validated?

src/api/constants.ts
```diff
@@ -61,6 +67,27 @@ export const ACCOUNT = {
...
+  },
+  // JSON web token
+  JWT: {
+    KEY: {
+      SIGNATURE: String(process.env.JWT_SIGNING_KEY),
+      ENCODING: 'base64' as BufferEncoding,
+      STRING_ENCODING: 'ascii' as BufferEncoding,
+    },
+    TOKEN: {
+      EXPIRY: '8h',
```
abhi-markan, line 81: `8h` Constant?

src/api/constants.ts
```diff
@@ -61,6 +67,27 @@ export const ACCOUNT = {
...
+      ALGORITHM: 'RS256' as Algorithm,
+    },
+    SESSION_EXPIRY: () => {
+      // 8 hours from now
+      const now = new Date();
+
+      const hours = 8;
+      const future = new Date(now.getTime() + hours * 60 * 60 * 1000);
```
abhi-markan, line 89: Constant?

abhi-markan (changes requested):
Great work, only few minor comments to address.

src/api/constants.ts
ttbarnes: This is in the constants file already?

src/api/.keystone/config.js
ttbarnes: `accountId` and checking that the account exists is already in place before this function is called e.g

```js
const exporter = await getAccountByField(context, 'email', email);

if (!exporter) {
  console.info('Unable to validate exporter account - no account found');

  return { success: false };
}
```

src/api/constants.ts
ttbarnes: Since this is already in constants file i've extracted the seconds into it's own variable like `hours`

src/api/.keystone/config.js
ttbarnes: This file is auto generated by keystone, the actual code we write is in `api/constants.ts` - same comment there, have addressed

abhi-markan (approved):

Zainzzkk (approved):
LGTM

PR #341, opened 2023-02-27
fix(EMS-554-1026-1036): various UI fixes

ttbarnes:
This PR fixes some UI issues:

## Changes

- Fixed an issue where if a user is on the "confirm email resent" and they navigate back to the "confirm email" page via back button, "problem with a service" page would appear. We now pass account ID as a query param back to the page, exactly the same as the "confirm email resent" paage.
- Extracted autocomplete UI client side scripts from nunjucks template, into the autocomplete script.
- Added E2E test coverage to check that autocomplete inputs are working with client side JS, by asserting a CSS class that is added by the client side JS.
- Add some missing copy for an exit page reason.
- Add `application/json` header to E2E API calls, this was causing some issues.
- Updated API unit tests for account sign in - was being flaky in GHA.


src/ui/server/controllers/insurance/account/create/confirm-email/index.test.ts
```diff
@@ -66,6 +83,19 @@ describe('controllers/insurance/account/create/confirm-email', () => {
       });
     });
 
+    describe('when there is no accountIdToConfirm in the session and no ID query param', () => {
+      beforeEach(() => {
```
IgnatG, line 87: very very small, if it's only one `it` statement, then this can be `beforeAll`?

ttbarnes: Great shout @IgnatG , beforeAll was causing some issues so i've just moved the before's entirely and have them inside the `it` statement.

Zainzzkk (approved):
lgtm

Zainzzkk (approved):
LGTM

PR #342, opened 2023-02-27
fix(EMS-991): Your business - Turnover - Percentage Turnover allowed commas

Zainzzkk:
# Issue:
* percentageTurnover field allowed commas as valid input

# Changes made:
* Added new helper to handle all validation for percentage numbers
* Removed commas being valid for percentages
* Added tests

### DIFF
--- a/src/ui/server/controllers/insurance/business/turnover/validation/rules/percentage-of-turnover.test.ts
+++ b/src/ui/server/controllers/insurance/business/turnover/validation/rules/percentage-of-turnover.test.ts
@@ -2,7 +2,6 @@ import percentageTurnover from './percentage-of-turnover';
 import { ERROR_MESSAGES } from '../../../../../../content-strings';
 import FIELD_IDS from '../../../../../../constants/field-ids/insurance/exporter-business';
 import { RequestBody, NumberErrorMessage } from '../../../../../../../types';
-import generateValidationErrors from '../../../../../../helpers/validation';
 import percentageNumberValidation from '../../../../../../helpers/percentage-number-validation';
 
 const {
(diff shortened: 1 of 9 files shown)


src/ui/server/controllers/insurance/business/turnover/validation/rules/percentage-of-turnover.test.ts
```diff
@@ -32,80 +39,19 @@ describe('controllers/insurance/business/turnover/validation/rules/percentage-of
     });
   });
 
-  describe(`when the ${PERCENTAGE_TURNOVER} input contains a letter`, () => {
-    it('should return a validation error', () => {
-      mockBody[PERCENTAGE_TURNOVER] = 'I5';
+  describe(`when the ${PERCENTAGE_TURNOVER} input is invaluid`, () => {
```
ttbarnes, line 42: typo

src/ui/server/controllers/insurance/business/turnover/validation/rules/percentage-of-turnover.ts
```diff
@@ -15,55 +13,28 @@ const {
   EXPORTER_BUSINESS: { [FIELD_ID]: ERROR_MESSAGE },
 } = ERROR_MESSAGES.INSURANCE;
 
-const MINIMUM = 0;
-const MAXIMUM = 100;
-
 /**
  * validates percentage turnover
- * only numbers without decimals or special characters
+ * only numbers without decimals or special characters or commas
```
ttbarnes, line 18: Very minor but could we reduce the "ors" here 

```js
only numbers without decimals, special characters or commas
```

src/ui/types/errors.ts
```diff
@@ -0,0 +1,7 @@
+interface Errors {
+  INCORRECT_FORMAT: string;
+  BELOW_MINIMUM: string;
+  ABOVE_MAXIMUM: string;
+}
```
ttbarnes, line 5: Can we call this `NumberErrorMessage`? More explicit and applies to both non-percentage and percentage type numbers.

src/ui/server/helpers/percentage-number-validation/index.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  const FIELD = 'testField';
+
+  const errorMessages = {
+    INCORRECT_FORMAT: 'Incorrect format',
+    BELOW_MINIMUM: 'Below minimum',
+    ABOVE_MAXIMUM: 'Above maximum',
+  } as Errors;
+
+  describe('percentage is a letter', () => {
+    it('should return a validation error', () => {
```
abhi-markan, line 24: Do we need test cases like percentage is `null`, `undefined` or `empty`?

abhi-markan (changes requested):
Good works, one minor change pleas.e

src/ui/server/helpers/percentage-number-validation/index.test.ts
Zainzzkk: no, as that is done in the main validation file, this exclusively checks the inputted value if it exists

ttbarnes: Good point @Zainzzkk seems we should add `objectHasProperty` validation helper function to the actual validation? Then in the test we can just add "if field is empty" because the helper function handles the rest.

Zainzzkk: it would mean that this could not be reused on any future optional percentage fields, is that fine?

ttbarnes (approved):

abhi-markan (approved):

PR #348, opened 2023-03-01
feat(EMS-1013): Account creation - only create an account if the email is unique

ttbarnes:
This PR updates account creation to ensure that an account can only be created if the email is unique and has not been used before.

## Changes

- Update `createAccount` resolver to reject if an account has already been created with the provided email address.
- Render a new validation error in the UI if an account already exists.
- Improve/add test coverage for the `createAccount` resolver.

### DIFF
--- a/src/api/custom-resolvers/create-account.ts
+++ b/src/api/custom-resolvers/create-account.ts
@@ -1,6 +1,7 @@
 import { Context } from '.keystone/types'; // eslint-disable-line
 import crypto from 'crypto';
 import { ACCOUNT } from '../constants';
+import getAccountByField from '../helpers/get-account-by-field';
 import { AccountCreationVariables } from '../types';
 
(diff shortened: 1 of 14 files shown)


src/api/custom-resolvers/create-account.ts
```diff
@@ -20,6 +21,15 @@ const createAccount = async (root: any, variables: AccountCreationVariables, con
   try {
     const { firstName, lastName, email, password } = variables;
 
+    // check if an account with the email already exists
+    const exporter = await getAccountByField(context, 'email', email);
```
Zainzzkk, line 25: can 'email' be a constant please

Zainzzkk (approved):
LGTM

src/api/custom-resolvers/create-account.ts
ttbarnes: Agreed will do this as part of next PR as there are a few instances of this

PR #355, opened 2023-03-05
chore(EMS-976): E2E tests - move backLink selector into pages/shared, clickBackLink cypress command

ttbarnes:
This PR simplifies our tests where the back link is involved.

- Move the back link selector into pages/shared, removing the need to import an additional file.
- Add a new cypress command `clickBackLink` so we can just call `cy.clickBackLink` instead of having to import the selector.


Zainzzkk (approved):
LGTM

PR #357, opened 2023-03-07
feat(EMS-626-1060): Application access - only allow an exporter to access an application if their ID matches

ttbarnes:
## Changes
- Create route/controller/page for "you do not have access" page.
- Create a middleware function (runs on all relevant insurance/application routes), to check if the account ID in session matches the application's exporter ID. If it does not match, redirect to the "no access" page. Otherwise, continue.
- Update account related E2E commands to have a default value, but accept a parameter.
- Add E2E test coverage.
- Add new environment variable for second account email testing, `GOV_NOTIFY_EMAIL_RECIPIENT_2`.

## E2E test changes

- Due to the E2E tests being updated so that login is required - this causes cypress to wipe a session when running multiple specs in a test. This is unfortunately not configurable and so we have had to: 
  - Update all E2E tests to use new cypress command `cy.saveSession`.
  - Go to the page/URL that is under test before running assertions.
  - Tweak/simplify/restructure various tests so there are less "describes", longer "its" and usage of `cy.navigateToUrl(...)`.


src/ui/server/middleware/insurance/application-access/index.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        DASHBOARD,
+        NO_ACCESS_TO_APPLICATION,
+      ];
+
+      expect(IRRELEVANT_ROUTES).toEqual(expected);
+    });
+  });
+
+  describe('when the route is not relevant', () => {
+    beforeEach(() => {
```
Zainzzkk, line 40: do we need the beforeEach as there is only 1 test in each describe

ttbarnes: We do because there's a spy for this test, the spy won't work if it's not in a `beforeEach`

e2e-tests/cypress/support/core-page-checks.js
```diff
@@ -37,8 +37,6 @@ const checkBackLink = (currentHref, expectedHref) => {
   cy.url().should('eq', expectedUrl);
 
   // go back to the current page
-  // cy.navigateToUrl(`${Cypress.config('baseUrl')}${currentHref}`);
```
Zainzzkk, line 40: is there a reason this is commented out and not removed?

Zainzzkk (approved):
LGTM

e2e-tests/cypress/support/core-page-checks.js
ttbarnes: This is strange - it's not commented out, but github PR is showing that it is. I've even pushed some commits changing this file and it's still showing this commented line. Viewing the file by itself inside this PR shows the latest 🤷‍♂️  https://github.com/UK-Export-Finance/exip/blob/540b66f8f1aab43743025cc760f6c50bdb09a6b9/e2e-tests/cypress/support/core-page-checks.js

abhi-markan (approved):

PR #360, opened 2023-03-08
feat(EMS-1081-1082): Account sign in - session expiry checks, signed out page

ttbarnes:
This PR adds user session expiry checks and if the current time is after the expired time, delete the session and redirect user to "we have signed you out" page.

## Changes
- Create user session middleware to check the current time vs session expiry time.
- Create route/controller/page for "we have signed you out" page.
- Update `verifyAccountSignInCode` GQL resolver to return the session expiry date.
- Note: unfortunately it seems impossible to add a true E2E test where a session has expired. I.e, after X hours, session should be deleted. Tried various approaches with `cy.clock()`  and setting cypress system time, but it does not translate into the UI, where the actual check is made.


e2e-tests/cypress/e2e/journeys/insurance/account/account-signed-out.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+const {
+  START,
+  ACCOUNT: {
+    SIGNED_OUT,
+    SIGN_IN: { ROOT: SIGN_IN_ROOT },
+  },
+} = ROUTES;
+
+context('TODO', () => {
```
Zainzzkk, line 16: should this be changed?

ttbarnes: yes 😂  missed this

Zainzzkk (approved):
LGTM

PR #364, opened 2023-03-14
feat(EMS-1145-1164): Declarations - Confidentiality

ttbarnes:
This PR sets up the first page/form for declarations - confidentiality.

## Changes
- Setup page, route, controller and field ID for confidentiality.
- Add confidentiality content to API schema complete with versioning.
- Create a new GQL call in the UI to get the latest confidentiality content/version and render in the confidentiality page.
- Update the task list to pass the application reference number to the "submit application" group, so that the correct link can be rendered for declarations.
- Simplify the `createSubmitApplicationTasks` dependencies array.
- Update and create new nunjucks components that render a list from a keystone document data structure. This consumes the data and renders a list appropriately as per design, with up to 3 levels supported.
- Add E2E test coverage.


e2e-tests/cypress/e2e/pages/insurance/declarations/confidentiality.js
```diff
@@ -0,0 +1,35 @@
...
+  INSURANCE: {
+    DECLARATIONS: { AGREE_CONFIDENTIALITY },
+  },
+} = FIELD_IDS;
+
+const confidentialityPage = {
+  listItems: {
+    intro: () => cy.get("[data-cy='paragraph-1']"),
+    level1: {
+      item1: () => cy.get("[data-cy='list-level-1-item-1']"),
```
Zainzzkk, line 13: could we simplify this to be `item: () => cy.get("[data-cy='list-level-${level}-item-${item}']"),`

src/ui/server/api/keystone/application/declarations.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (response.errors) {
+        console.error('GraphQL error getting latest declaration - confidentiality ', response.errors);
+      }
+
+      if (response?.networkError?.result?.errors) {
+        console.error('GraphQL network error getting latest declaration - confidentiality ', response.networkError.result.errors);
+      }
+
+      if (response?.data?.declarationConfidentialities && isPopulatedArray(response.data.declarationConfidentialities)) {
+        return response.data.declarationConfidentialities[0];
```
Zainzzkk, line 22: is there a reason we need to return index 0?

e2e-tests/cypress/e2e/pages/insurance/declarations/confidentiality.js
ttbarnes: I did think about this, I think it would end up too vague/ambiguous as we'd end up with tests like this:

```js
cy.checkText(item(1, 2), content[1].text);
cy.checkText(item(3, 1), content[2].text);
cy.checkText(item(2, 4), content[3].text);
```

Extracting the numbers would add a lot of lines of code as well.

What do you think?

src/ui/server/api/keystone/application/declarations.ts
ttbarnes: Yes, it's an array, returns only one object so we return only this in the response

Zainzzkk (approved):
LGTM

PR #365, opened 2023-03-14
feat(EMS-1164) Declarations - Confidentiality - data saving, save and go back

ttbarnes:
This PR updates the Confidentiality page/flow so that the declaration answer is saved via the regular submit button, or the "save and back" button. Also update the task list to change to "in progress" after submitting an answer.

## Changes

- Update the API schema to have a new table, `Declaration` that has a relationship with an application.
- Add a new GQL call from the API to update an application's declarations.
- Update the "get application" GQL call to return an application's declarations answers.
- Update the Declarations - Confidentiality POST endpoint to call the API via `save-data` function.
- Add new endpoint Declarations - Confidentiality "save and back" route and controller.
- Create a new helper function `getTrueProperties` that returns only properties in an object that have a true boolean. This is required to strip out false answers to declarations when the task list logic is run.
- Update the task list for the first Declarations field, therefore making the Declarations task change to "in progress" once the first field is answered correctly.
- E2E test coverage.
- Update DB dump.

### DIFF
--- a/src/ui/server/controllers/insurance/declarations/confidentiality/save-and-back/index.test.ts
+++ b/src/ui/server/controllers/insurance/declarations/confidentiality/save-and-back/index.test.ts
@@ -1,0 +1,105 @@
+import { post } from '.';
+import { ROUTES } from '../../../../../constants';
+import { Request, Response } from '../../../../../../types';
+import save from '../save-data';
+import { mockApplication, mockReq, mockRes } from '../../../../../test-mocks';
+
+const {
+  INSURANCE: { INSURANCE_ROOT, ALL_SECTIONS },
+} = ROUTES;
+
+describe('controllers/insurance/declarations/confidentiality/save-and-back', () => {
+  let req: Request;
+  let res: Response;
+
+  jest.mock('../save-data');
+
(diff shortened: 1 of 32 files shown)


src/ui/server/controllers/insurance/declarations/confidentiality/save-and-back/index.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    res = mockRes();
+
+    res.locals.application = mockApplication;
+    req.params.referenceNumber = String(mockApplication.referenceNumber);
+
+    req.body = mockFormBody;
+  });
+
+  describe('when the form has data', () => {
+    it('should call mapAndSave.policyAndExport with application and req.body', async () => {
```
Zainzzkk, line 38: typo

Zainzzkk (approved):
LGTM

src/ui/server/controllers/insurance/declarations/confidentiality/save-and-back/index.test.ts
ttbarnes: As discussed will fix this in the next PR

PR #367, opened 2023-03-15
fix(EMS-1263): application access - add a check for exporter ID

ttbarnes:
Currently in dev if an application does not have an exporter ID, the application access middleware fails because it cannot read the ID and then the server crashes.

The middleware might be running somewhere were it doesn't need to - need to look into this (not a quick thing). In the meantime, need this changes merged so we can unblock QA / dev environment.

## Changes

- Add an additional check to make sure exporter object exists.

### DIFF
--- a/src/ui/server/middleware/insurance/application-access/index.test.ts
+++ b/src/ui/server/middleware/insurance/application-access/index.test.ts
@@ -84,6 +84,24 @@ describe('middleware/insurance/application-access', () => {
       });
     });
 
+    describe('when res.locals.application does not have an exporter ID', () => {
+      beforeEach(() => {
+        res.locals.application = {
+          ...mockApplication,
+          // @ts-ignore
+          exporter: {},
+        };
+
+        next = nextSpy;
+      });
+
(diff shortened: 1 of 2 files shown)


src/ui/server/middleware/insurance/application-access/index.test.ts
```diff
@@ -84,6 +84,24 @@ describe('middleware/insurance/application-access', () => {
       });
     });
 
+    describe('when res.locals.application does not have an exporter ID', () => {
+      beforeEach(() => {
+        res.locals.application = {
+          ...mockApplication,
+          // @ts-ignore
```
ttbarnes, line 91: Note - this is required. Exporter ID is a required type in the Application type. So we cannot have an empty exporter object.

However in the dev environment, this has been the case, where an application has an empty exporter object. This might be from old applications before the exporter relationship was introduced.

These changes makes sure that this won't happen and we need to add a unit test for this.

Zainzzkk (approved):
LGTM

PR #368, opened 2023-03-15
feat(EMS-912-922-929): Dashboard, default application status

ttbarnes:
This PR updates the dashboard to map applications and render a table as per design. Also updated applications to have an initial status of "draft" .

## Changes

- Add status to an application. When created, it's defaulted to "draft".
- Update dashboard page/controller to map applications via a new mapping function and render in a table.
- E2E test coverage for dashboard page, table and user journeys / population as the application is updated.
- Extracted some cypress commands, this will help us in the next dashboard PR.
- Also fixed a typo.


Zainzzkk (approved):
LGTM

PR #369, opened 2023-03-15
feat(EMS-912): Dashboard - update all E2E tests to delete application as well as account

ttbarnes:
Now that the insurance dashboard is in place and this will soon be updated to only return an exporter's applications (instead of all applications), we need to have a clean DB state for these future tests.

## Changes
- Update all E2E tests that have account sign in and application creation to:
  - Decalare `referenceNumber` in the same way.
  - Use a new cypress command `deleteAccountAndApplication`, instead of `deleteAccount`.
- Create new GQL/API calls in the E2E test suite to get an application ID's by reference number and then delete it by ID.
- Also fixed some typos.
- ⚠️ Temporarily disable "get a quote" and some generic static page E2E tests so we can run the tests more quckly and clear the PR backlog.


Zainzzkk (approved):
LGTM

Zainzzkk (approved):
LGTM

Zainzzkk (approved):
LGTM

PR #370, opened 2023-03-16
fix(EMS-849): Account verification - handle invalid tokens UI redirection

ttbarnes:
This PR fixes an issue where if an account verification token is invalid, or the account does not exist - it would redirect to the "problem with service" page instead of the "link expired" page.

## Changes

- Update `VerifyAccountEmailAddressResponse` type to have an optional  `accountId` instead of optional, therefore the an error will not be thrown if an account is not found.
- Add E2E test coverage.


Zainzzkk (dismissed):
lgtm

Zainzzkk (approved):
lgtm

PR #395, opened 2023-03-24
feat(EMS-1144): Check your answers - completed status tag

ttbarnes:
This PR updates all "check your answers" controllers/pages to get all fields required for the section and generate a "completed" status if the section is completed.

Note: for MVP, we are only be displaying "completed", as per design. The logic is setup to handle other statuses e.g "in progress", but it is very edge case to be in this scenario where a section is incomplete. We could, but we are intentionally not testing for this - after MVP/in the very near future, we will be restricting these pages so a user cannot navigate to these pages if the section is incomplete.

## Changes

- Extract some logic and definitions from task list creation functions:
  - New `getSubmittedFields` helper function to get certain fields from submitted application data.
  - New `requiredFields` helper functions for all sections of an applcation. Each function gets all required fields for the section. Depending on the section, there are conditionally required fields depending on what has been submitted.
  - Generic `sectionStatus` function that checks required and submitted fields and returns status for the section.
  - All of these functions are now used in Check your answers and task list generation and simplifies the task list functions.
- Update the check your answers controllers and generic template to pass the section status and render a status tag in the UI.
- Fix/update the task list ordering and wording for declarations and check your answers.
- Create new cypress command to check a task status is "completed" - `checkTaskStatusCompleted`.
- Update E2E tests to check that each check your answers page renders a completed status tag, and that this is retained when changing any answer in any section.


ttbarnes:
Moved to https://github.com/UK-Export-Finance/exip/pull/396

PR #404, opened 2023-03-29
feat(EMS-1254-1260): Application submission - exporter document emails

ttbarnes:
This PR adds the second and final email that gets sent to an exporter when an application is submitted.

The email template that is sent depends on certain answers to a couple of questions RE the buyer and declarations.

## Changes
- Add new email template IDs.
- Rename `sendEmailApplicationSubmitted` to `sendApplicationSubmittedEmails`.
- Add new `documentsEmail` function to the emails object.
- Rename/restructure `sendApplicationSubmittedEmails` into `applicationSubmittedEmails.send`, makes for simpler unit tests/stubbing.
- Add logic to `sendApplicationSubmittedEmails` so that it gets the application's declaration and depending on some answers, send the document email with the appropriate template.

## Other improvements
- Improved test coverage for `sendEmail` functions.
- Move `mockBuyer` into test mocks.
- Update mock declarations.
- Update `applicationSubmittedEmail` function to accept an object instead of string params - DRY.
- Update/align UI mock application.


src/api/emails/send-application-submitted-emails/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (!declaration) {
+      console.error('Sending application submitted emails to exporter - no declarations exist with the provided ID');
+
+      return {
+        success: false,
+      };
+    }
+
+    // generate email variables
+    const { email, firstName } = exporter;
```
abhi-markan, line 62: I guess `email` and `firstName` will always exist?

abhi-markan (commented):
Good work, only minor comments to address please.

src/api/.keystone/config.js
```diff
@@ -1353,24 +1373,52 @@ var sendEmailApplicationSubmitted = async (context, accountId, buyerId, referenc
...
+      buyerName: buyer.companyOrOrganisationName
+    };
+    const submittedResponse = await emails_default.applicationSubmittedEmail(sendEmailVars);
+    if (!submittedResponse.success) {
+      throw new Error("Sending application submitted emails to exporter");
     }
-    throw new Error(`Sending email to exporter - application submitted ${emailResponse}`);
+    let documentsResponse;
+    let useAntiBriberyAndTradingHistoryTemplate = false;
+    if (buyer.exporterIsConnectedWithBuyer && buyer.exporterIsConnectedWithBuyer === "Yes") {
```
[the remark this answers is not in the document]
ttbarnes: They are loaded from constants but in this file (keystone compilation), it's compiled without the constant which makes sense

I.e where we actually define it, we have:

```js
const isConectedWithBuyer = buyer.exporterIsConnectedWithBuyer && buyer.exporterIsConnectedWithBuyer === ANSWERS.YES;
```

src/api/emails/send-application-submitted-emails/index.ts
ttbarnes: Yes, they will always exist - an account/exporter cannot be created without these, amongst other fields

abhi-markan (approved):

PR #412, opened 2023-04-05
feat(EMS-1179): improved docker image build

abhi-markan:
## Introduction
Docker image build process has not been following best practices, thus costing image build process in terms of time and file size.

## Miscellaneous
* Dependencies update


src/ui/.dockerignore
```diff
@@ -1,6 +1,10 @@
 .env
+.env.sample
+.gitignore
 .prettierrc.json
+.prettierignore
+.eslintrc.json
 .eslintignore
-node_modules
```
IgnatG, line 4: we should keep the node_modules in the `.dockerignore` file as well to avoid copying it to the container

docker-compose.yml
```diff
@@ -1,37 +1,47 @@
...
     image: mysql
+    container_name: exip_db
     restart: always
     command: --default-authentication-plugin=mysql_native_password
     ports:
-      - "3306:3306"
+      - "${DATABASE_PORT}:${DATABASE_PORT}"
     volumes:
-      - ./database/exip.sql:/docker-entrypoint-initdb.d/database.sql
+      - ./database/exip.sql:/docker-entrypoint-nitdb.d/database.sql:rw
```
IgnatG, line 13: is it not called `initdb` anymore?

docker-compose.yml
```diff
@@ -45,20 +55,25 @@ services:
...
-      - "5000:5000"
+      - "${UI_PORT}:${UI_PORT}"
     volumes:
       - ./src/ui/server:/app/server:ro
       - ./src/ui/public:/app/public:ro
       - ./src/ui/templates:/app/templates:ro
     environment:
-      PORT: 5000
+      PORT: ${UI_PORT}
+      NODE_ENV:
```
IgnatG, line 76: we could probably populate these `NODE_ENV` with `development` value by default since we only use it locally for development

src/api/.dockerignore
```diff
@@ -1,6 +1,9 @@
 .env
+.env.sample
+.gitignore
 .prettierrc.json
+.prettierignore
+.eslintrc.json
 .eslintignore
-node_modules
```
IgnatG, line 4: we should keep the `node_modules` folder to avoid copying it to the Docker image manually

docker-compose.yml
abhi-markan: A good spot, must have been an accidental delete.

src/api/Dockerfile
```diff
@@ -1,30 +1,75 @@
-# Node Alpine Docker Image
-FROM node:18.9-alpine3.16
+###################
+# 1. BUILD
+###################
 
-# Variables
+# NPM 8.19.1 Alpine linux image
+FROM node:18.15-alpine3.16 AS build
```
IgnatG, line 6: keep an eye in Azure to ensure that this version of Node is supported

src/api/Dockerfile
```diff
@@ -1,30 +1,75 @@
...
+
+# Build variables
 ARG GITHUB_SHA
 ARG SESSION_SECRET
 ARG GOV_NOTIFY_API_KEY
-ENV GITHUB_SHA $GITHUB_SHA
-ENV SESSION_SECRET $SESSION_SECRET
-ENV GOV_NOTIFY_API_KEY $GOV_NOTIFY_API_KEY
+ARG PORT
+ENV GITHUB_SHA=$GITHUB_SHA
```
IgnatG, line 13: do we need `=`? I thought we can use a space between the variable and value

src/ui/.dockerignore
abhi-markan: Good shout have added.

src/api/.dockerignore
abhi-markan: Have added.

src/api/Dockerfile
```diff
@@ -1,30 +1,75 @@
...
+COPY --chown=node:node --from=build /app/helpers ./helpers
+COPY --chown=node:node --from=build /app/emails ./emails
+COPY --chown=node:node --from=build /app/custom-resolvers ./custom-resolvers
+COPY --chown=node:node --from=build /app/test-mocks ./test-mocks
+
+# Non-root user
+USER node
+
+# Container healthcheck
+HEALTHCHECK \
```
IgnatG, line 65: just thinking about it now, we probably don't need the healthcheck here because Azure can ping the container using one of their settings. And for local development, we have docker-compose.yml file which has the healthcheck property

docker-compose.yml
abhi-markan: We use `NODE_ENV` with in the source code whether to populate HTTP based authentication. This is effects healthcheck if set to `development` `ui` healtcheck will fail as it will return 404 instead of 2022.

src/api/Dockerfile
abhi-markan: Will do.

src/api/Dockerfile
abhi-markan: I have moved healthcheck from docker-compose to Dockerfile.

src/api/Dockerfile
```diff
@@ -1,30 +1,75 @@
...
+
+# Non-root user
+USER node
+
+###################
+# 2. PRODUCTION
+###################
+
+# NPM 9.5.1 Alpine linux image
+FROM node:19.8-alpine3.16 AS production
```
IgnatG, line 46: this version of node is different than the one from above. shouldn't both of them be the same?

src/api/Dockerfile
abhi-markan: Either is acceptable, I referred to official docker build guideline https://docs.docker.com/engine/reference/builder/.

src/api/Dockerfile
abhi-markan: Build image has quite number of vulnerabilities, where as the `production` base image has only `1`.

src/ui/Dockerfile
```diff
@@ -1,22 +1,64 @@
...
+COPY --chown=node:node --from=build /app/scripts ./scripts
+COPY --chown=node:node --from=build /app/styles ./styles
+COPY --chown=node:node --from=build /app/server ./server
+COPY --chown=node:node --from=build /app/templates ./templates
+COPY --chown=node:node --from=build /app/types ./types
+
+# Non-root user
+USER node
+
+# Container healthcheck
```
IgnatG, line 53: same as above, we shouldn't need this as Azure allows us to do it from settings and for local development we have it in a setting for docker-compose.yml file

src/ui/Dockerfile
```diff
@@ -1,22 +1,64 @@
...
+# `devDependencies` will still be resolved inside `package-lock.json`,
+# however they will not be installed inside `node_modules` directory.
+RUN npm ci --legacy-peer-deps --omit=dev --ignore-scripts
+
+###################
+# 2. PRODUCTION
+###################
+
+# NPM 9.5.1 Alpine linux image
+FROM node:19.8-alpine3.16 AS production
```
IgnatG, line 34: same as above - shouldn't we align the 2 versions of Node?

src/api/Dockerfile
IgnatG: cool. but shouldn't we use the same version from prod to build? i.e. both versions should be 19.8?

src/api/Dockerfile
IgnatG: ok cool

package.json
```diff
@@ -8,12 +8,8 @@
     "npm": ">=9.5.0"
   },
   "scripts": {
-    "build": "tsc -p tsconfig.json && tsc -p tsconfig-cjs.json",
-    "postinstall": "husky install",
-    "prepublishOnly": "npm run build",
```
ttbarnes, line 13: Why are we removing these and `prepare` script please? These stem from the template-typescript-package package

src/ui/tsconfig.json
```diff
@@ -33,6 +33,6 @@
     "allowJs": true,
     "importHelpers": true
   },
-  "include": ["**/*.ts"],
-  "exclude": ["./node_modules/*", "node_modules"]
+  "include": ["**/*.ts", "./server", "./jest.config.js", "./.eslintrc.js"],
+  "exclude": ["node_modules", "generated_reports", "logs"]
```
ttbarnes, line 37: I don't think we need logs, I don't believe this directory can exist? Unless when in docker/deployment environment this is added?

IgnatG (dismissed):
LGTM

package.json
ttbarnes: bump, I think we should keep `postinstall`

ttbarnes (approved):

PR #413, opened 2023-04-05
feat(EMS-1268-1332-1333): Authenticated header

ttbarnes:
This PR updates all controllers to pass `req.session.user` to all page tempates. The header can therefore check if there is a user session and if so, render some header navigation links.

## Changes
- Update all UI controllers to call a new helper function `getUserNameFromSession` in order to pass a user's first and alst name to the page template.
- Update the header nunjucks component to conditionally render navigation items if there is a user session.
- Create new cypress command:`checkAuthenticatedHeader`.
- Update E2E core page checks command to run `checkAuthenticatedHeader` by default.
- Update E2E page tests that do not require a user session/account sign in, to _not_ check an authenticated header in the core page checks command.
- Add E2E test coverage for:
  - Authenticated header elements when sign in.
  - Authenticated header persistance when signed in and starting a new application via the eligibility flow.
  - Checking the authenticated header elements do not render by default.

## Other improvements
- Address some todo comments in E2E tests - in various tests, we can now mimic a real world user journey by clicking on the header's dashboard link, instead of manually navigating to the dashboard URL.
- Align types and test-mocks import order in various cnotrollers.


src/ui/server/controllers/insurance/eligibility/uk-goods-or-services/index.ts
```diff
@@ -21,6 +21,7 @@ export const TEMPLATE = TEMPLATES.INSURANCE.ELIGIBILITY.UK_GOODS_OR_SERVICES;
 export const get = (req: Request, res: Response) =>
   res.render(TEMPLATE, {
     ...singleInputPageVariables({ ...PAGE_VARIABLES, BACK_LINK: req.headers.referer }),
+    user: req.session.user,
```
IgnatG, line 24: just a sanity check:
a) do we need all properties from the req.session.user?
b) do we expect all properties to be available? or do we need to check `req.session?.user`?

ttbarnes: Good shout - we only really need the first and last name - will create a function to only return these properties if they exist.

BTW this is currently using the same approach as DTFS so might want to update it there as well

IgnatG (dismissed):
LGTM - I've added a small comment, but not a blocker

src/ui/server/controllers/insurance/eligibility/uk-goods-or-services/index.ts
IgnatG: cool. happy for this to be a separate PR if you want

ttbarnes:
Thanks @IgnatG let's do it as part of this PR, do it properly 😊

abhi-markan (approved):

PR #415, opened 2023-04-05
feat(EMS-1334): Account - sign out

ttbarnes:
This PR adds the ability to sign out of an account.

## Changes

- Add sign out GET route/controller to the UI. This controler wipes the user session and adds a flag to `req.flash`.
- Update the sign in page/controller to consume `req.flash('importantBanner')` and render a banner in the template.
- E2E test coverage.

### DIFF
--- a/src/api/.keystone/config.js
+++ b/src/api/.keystone/config.js
@@ -18,6 +18,10 @@ var __copyProps = (to, from, except, desc) => {
   return to;
 };
 var __toESM = (mod, isNodeMode, target) => (target = mod != null ? __create(__getProtoOf(mod)) : {}, __copyProps(
+  // If the importer is in node compatibility mode or this is not an ESM
+  // file that has been converted to a CommonJS file using a Babel-
+  // compatible transform (i.e. "__esModule" has not been set), then set
+  // "default" to the CommonJS "module.exports" for node compatibility.
   isNodeMode || !mod || !mod.__esModule ? __defProp(target, "default", { value: mod, enumerable: true }) : target,
   mod
 ));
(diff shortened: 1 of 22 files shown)


src/ui/server/controllers/insurance/account/sign-out/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * get
+ * Account sign out / wipe user session
+ * @param {Express.Request} Express request
+ * @param {Express.Response} Express response
+ * @returns {Express.Response.redirect} Redirect to sign in page
+ */
+export const get = (req: Request, res: Response) => {
+  delete req.session.user;
```
abhi-markan, line 18: Do we need to check whether `session.use` exists before we delete it?

abhi-markan (commented):
Looks good, just a comment to address please.

src/ui/server/controllers/insurance/account/sign-out/index.ts
ttbarnes: No, it will work even if it's undefined. If it was more than 1 level we would have to check

src/api/.keystone/config.js
```diff
@@ -1415,6 +1415,10 @@ var send = async (context, referenceNumber, accountId, buyerId, declarationId, e
       referenceNumber,
       buyerName: buyer.companyOrOrganisationName,
       exporterCompanyName: exporterCompany.companyName,
+<<<<<<< Updated upstream
```
Zainzzkk, line 1418: dont think this is meant to be here

Zainzzkk (approved):
LGTM

PR #419, opened 2023-04-07
fix(EMS-1352): Application relationships - update application's updatedAt timestamp

ttbarnes:
This PR updates all application relationships so that when an update is made, the application's `updatedAt` timestamp is updated to the latest time.

## Changes
- New helper function `updateApplication.timestamp`.
- Update each relationship in the keystone schema  to utilise the `afterOperation` hook and call the new helper function.


IgnatG (approved):
LGTM

PR #423, opened 2023-04-11
fix(EMS-1328): Declarations list item styles

ttbarnes:
This PR fixes an issue where some declarations list would display incorrect list item numbers, cross browser, due to useage of CSS property `counter-increment` which gives us custom list item numbers formatting.

## Changes

- Remove `counter-increment` usage.
- Add `govuk-list--number` class to ordered lists rendered in the keystone document component.
- Remove now unused styles and `counter-list` class.


Zainzzkk (approved):
LGTM

PR #427, opened 2023-04-14
feat(EMS-1281): Feedback - Phase banner link

Zainzzkk:
# Changes made:
* Added correct route from phase banner for insurance and quote
* Added tests and e2e tests
* Changed to content strings for phase-banner


src/ui/server/controllers/root/contact-us/index.test.ts
```diff
@@ -36,6 +36,7 @@ describe('controllers/root/contact-us', () => {
           BACK_LINK: req.headers.referer,
           PRODUCT: { DESCRIPTION: PRODUCT.DESCRIPTION.GENERIC },
           START_ROUTE: startRoute,
+          FEEDBACK_ROUTE: LINKS.EXTERNAL.FEEDBACK,
```
ttbarnes, line 39: Could this be a default in `corePageVariables` function, to save adding to all the root controllers?

src/ui/templates/partials/phase-banner.njk
```diff
@@ -1,5 +1,9 @@
 {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
 
+{% set bannerHtml %}
+  <span data-cy="phase-banner-text">This is a new service - your <a class="govuk-link" href="{{ FEEDBACK_ROUTE }}" data-cy="phase-banner-feedback-link">feedback</a> will help us to improve it.</span>
```
ttbarnes, line 4: Whilst we're in here could we have this text coming from `CONTENT_STRINGS`? I think it's the only thing that doesn't come from content strings 😢

e2e-tests/cypress/support/check-phase-banner.js
```diff
@@ -1,13 +1,20 @@
...
 
-  cy.checkText(partials.phaseBanner.feedbackLink(), 'feedback');
+  let route = LINKS.EXTERNAL.FEEDBACK;
 
-  // TODO: EMS-1281
-  // partials.phaseBanner.feedbackLink().should('have.attr', 'href', LINKS.EXTERNAL.FEEDBACK);
+  if (isInsurancePage) {
+    route = INSURANCE_ROUTES.FEEDBACK;
+  }
+  partials.phaseBanner.feedbackLink().should('have.attr', 'href', route);
```
ttbarnes, line 19: Not a blocker - could use `checkLink` here and remove line 12

src/ui/server/controllers/root/cookies/index.ts
```diff
@@ -54,7 +54,7 @@ export const post = (req: Request, res: Response) => {
 
   return res.render(TEMPLATE, {
     userName: getUserNameFromSession(req.session.user),
-    ...singleInputPageVariables({ ...PAGE_VARIABLES, BACK_LINK: backLink, START_ROUTE: startRoute }),
+    ...singleInputPageVariables({ ...PAGE_VARIABLES, BACK_LINK: backLink, START_ROUTE: startRoute, FEEDBACK_ROUTE: LINKS.EXTERNAL.FEEDBACK }),
```
ttbarnes, line 57: Looks like we don't need the `FEEDBACK_ROUTE` here because `singleInputPageVariables` has it?

ttbarnes (approved):

PR #428, opened 2023-04-14
feat(EMS-1381): Application submission - safety checks

ttbarnes:
This PR adds some additional checks to application submission to ensure that:

- The number of required fields equals the number of submitted fields.
- The date of submission does not exceed the application submission deadline.
- The application has a "draft" status.

## Changes
- Add all declaration fields to the `updateApplicationDeclarationMutation` query response.
- Update `how-your-data-will-be-used` POST controller to add the final declaration answer into an application object structure - saves making another API call. Then, call new helper functions to check if the application can be submitted.
- Create new helper function - `applicationIsComplete`. This helper utilises existing `required-fields` definitions.
- Create new `required-fields` definitions for declarations, section review (check your answers) and an entire application,
- Create new helper function - `canSubmitApplication`. This helper checks that the application is complete, has a `DRAFT` status and does not exceed the application's submission deadline.
- Update `flattenApplicationData` helper function to specifically consume the top level application properties.
- Rename `getTrueProperties` helper function to `getYesAndTrueAnswerProperties`, so it now returns fields with a true boolean or "Yes" value.
- Update `application-submitted` GET controller test to use an application with a status of `SUBMITTED`.

## Other improvements
- Fixed some typos.
- Add `submissionDate` to the UI's mock application.


src/api/custom-resolvers/submit-application.ts
```diff
@@ -21,14 +22,17 @@ const submitApplication = async (root: any, variables: SubmitApplicationVariable
     })) as Application;
 
     if (application) {
-      const canSubmit = application.status === APPLICATION.STATUS.DRAFT;
+      const hasDraftStatus = application.status === APPLICATION.STATUS.DRAFT;
```
Zainzzkk, line 25: will this be needed anywhere else? could it go in a helper

ttbarnes: It won't be used anywhere other than in here - we could extract, but then we'd have duplicated unit tests, I don't think it's worth doing

Zainzzkk (approved):
LGTM

PR #430, opened 2023-04-18
feat(EMS-1459): Feedback - Add feedback to database

Zainzzkk:
# Changes made
* Added ability to add feedback to database
* Added to schema
* Added unit and api test

### DIFF
--- a/src/api/schema.ts
+++ b/src/api/schema.ts
@@ -674,4 +674,14 @@ export const lists = {
     },
     access: allowAll,
   }),
+  Feedback: list({
+    fields: {
+      type: text(),
+      satisfaction: text(),
+      improvement: text(),
+      otherComments: text(),
+      referralUrl: text(),
+    },
+    access: allowAll,
+  }),
 } as Lists;
--- a/src/ui/server/controllers/insurance/feedback/feedback-form/index.test.ts
+++ b/src/ui/server/controllers/insurance/feedback/feedback-form/index.test.ts
@@ -2,12 +2,22 @@ import { TEMPLATE, pageVariables, MAXIMUM, get, post } from '.';
 import { PAGES, FIELDS } from '../../../../content-strings';
 import { TEMPLATES, ROUTES, FIELD_IDS } from '../../../../constants';
 import insuranceCorePageVariables from '../../../../helpers/page-variables/core/insurance';
-import { mockReq, mockRes } from '../../../../test-mocks';
+import { mockReq, mockRes, mockInsuranceFeedback } from '../../../../test-mocks';
 import { Request, Response } from '../../../../../types';
 import api from '../../../../api';
 import generateValidationErrors from './validation';
 
-const { SATISFACTION, IMPROVEMENT, OTHER_COMMENTS, VERY_SATISFIED, SATISFIED, NEITHER, DISSATISFIED, VERY_DISSATISIFED } = FIELD_IDS.FEEDBACK;
+const {
+  SATISFACTION,
+  IMPROVEMENT,
+  OTHER_COMMENTS,
+  VERY_SATISFIED,
+  SATISFIED, NEITHER,
+  DISSATISFIED,
+  VERY_DISSATISIFED,
+  REFERRAL_URL,
+  TYPE
+} = FIELD_IDS.FEEDBACK;
 
 const { FEEDBACK_PAGE } = PAGES;
 const { FEEDBACK: FEEDBACK_TEMPLATE } = TEMPLATES.INSURANCE;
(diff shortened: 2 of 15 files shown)


src/api/schema.ts
```diff
@@ -674,4 +674,14 @@ export const lists = {
     },
     access: allowAll,
   }),
+  Feedback: list({
+    fields: {
+      type: text(),
+      satisfaction: text(),
```
ttbarnes, line 680: Could we use the `select` keystone field with options coming from constants?

src/api/schema.ts
```diff
@@ -674,4 +674,14 @@ export const lists = {
...
+      improvement: text(),
+      otherComments: text(),
+      referralUrl: text(),
```
ttbarnes, line 683: Should this be required? If so could it come from `select` field with options?

src/ui/server/controllers/insurance/feedback/feedback-form/index.test.ts
```diff
@@ -109,16 +125,28 @@ describe('controllers/insurance/feedback/feedback-confirmation', () => {
...
+
+          await post(req, res);
+
+          const saveVariables = {
+            ...mockInsuranceFeedback,
+            [TYPE]: 'Insurance',
+            [REFERRAL_URL]: req.headers.referer,
+          };
+
+          expect(api.keystone.feedback.createInsuranceFeedback).toHaveBeenCalledTimes(1);
```
ttbarnes, line 146: Can we rename the structure so it's `feedback.create()`?

src/ui/server/controllers/insurance/feedback/feedback-form/index.ts
```diff
@@ -95,9 +111,20 @@ const post = async (req: Request, res: Response) => {
         [OTHER_COMMENTS]: feedback[OTHER_COMMENTS],
       } as InsuranceFeedbackVariables;
 
+      // variables for adding feedback to the database
+      const databaseVariables = {
```
ttbarnes, line 115: `feedbackVariables` please, we shouldn't be mentioned DB in the UI, the UI only interacts with the API

src/ui/server/controllers/insurance/feedback/feedback-form/index.ts
```diff
@@ -87,6 +101,8 @@ const post = async (req: Request, res: Response) => {
       });
     }
 
+    const referralUrl = req.flash('feedbackOriginUrlDB');
```
ttbarnes, line 104: Seems we don't need `feedbackOriginUrlDB` as it's got the same value as `req.flash('feedbackOriginUrl')`?

src/ui/server/controllers/insurance/feedback/feedback-form/index.ts
```diff
@@ -95,9 +111,20 @@ const post = async (req: Request, res: Response) => {
         [OTHER_COMMENTS]: feedback[OTHER_COMMENTS],
       } as InsuranceFeedbackVariables;
 
+      // variables for adding feedback to the database
+      const databaseVariables = {
+        [SATISFACTION]: feedback[SATISFACTION] ?? '',
+        [IMPROVEMENT]: feedback[IMPROVEMENT],
+        [OTHER_COMMENTS]: feedback[OTHER_COMMENTS],
+        [TYPE]: 'Insurance',
```
ttbarnes, line 119: Can we move this into a constant - also "type" field name feels confusing, how about `service` `SERVICE`?

I think it'd be a good idea to add default `serviceName` or `product` as well with a value of "EXIP", for future data consumption outside of our DB. Not sure about the namings, what do you think?

src/api/schema.ts
Zainzzkk: but it could come from any page in the service?

src/ui/server/controllers/insurance/feedback/feedback-form/index.ts
```diff
@@ -87,6 +92,8 @@ const post = async (req: Request, res: Response) => {
       });
     }
 
+    const referralUrl = req.flash('feedbackOriginUrlAPI');
```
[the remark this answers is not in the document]
Zainzzkk: had done it incase the page is refreshed, since this is meant to be an important one to store in the db, can change it if required?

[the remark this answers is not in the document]
Zainzzkk: does not work without the flash.  For the post it adds the give-feedback route instead from the get request

src/ui/server/controllers/insurance/feedback/feedback-form/index.ts
```diff
@@ -95,9 +102,20 @@ const post = async (req: Request, res: Response) => {
...
+      const feedbackVariables = {
+        [SATISFACTION]: feedback[SATISFACTION] ?? '',
+        [IMPROVEMENT]: feedback[IMPROVEMENT],
+        [OTHER_COMMENTS]: feedback[OTHER_COMMENTS],
+        [SERVICE]: INSURANCE,
+        [REFERRAL_URL]: referralUrl?.toString(),
+        [PRODUCT]: SERVICE_NAME,
+      };
+
+      const saveResponse = await api.keystone.feedback.create(feedbackVariables);
```
ttbarnes, line 114: Not a blocker, but we should just have 1 API call "create feedback" that creates the feedback and sends the email. Can improve after merge

Zainzzkk: will improve this after merge

ttbarnes (approved):

PR #436, opened 2023-04-20
feat(EMS-968): Account - Password reset - link sent page

ttbarnes:
This PR adds a "link sent" page for the password reset page/journey.

## Changes

- Add route, controller and page.
- Update the "password reset" controller to add the submitted email to `req.session`. The new "link sent" controller then deletes this from the session.
- Add E2E test coverage.
- Also fixed some typos.


src/ui/server/controllers/insurance/account/password-reset/index.ts
```diff
@@ -60,7 +60,7 @@ export const get = (req: Request, res: Response) =>
  */
 export const post = async (req: Request, res: Response) => {
   try {
-    const email = sanitiseValue(FIELD_ID, req.body[FIELD_ID]);
+    const email = String(sanitiseValue(FIELD_ID, req.body[FIELD_ID]));
```
Zainzzkk, line 63: could this move below the validation error check, incase for any reason the email is null?

ttbarnes: Good shout, have moved

Zainzzkk (approved):

PR #445, opened 2023-04-26
feat(EMS-1083 965): Account - Password reset - set new password

ttbarnes:
This PR adds the ability to create/encrypt a new account password after visiting a URL with password reset token.

## Changes

- Create UI route, controler and page for 2 new pages - "new password" and "success".
- Add E2E test coverage for:
  - Successfully submitting a new password.
  - Trying to navigate to the "new password" form/page without a token.
  - Page/element tests for "new password" and "success" pages.
- Create a new GQL query mutation, `accountPasswordReset`. This checks if an account exists and has a password reset hash and expiry. If so, check that we are not passed the expiry date. If all the checks are valid we generate an encrypted password provided by the UI and update the account. This mutation is called in the UI on successful "new password" form submission.
- Create a new GQL query resolver for E2E tests, `getAccountPasswordResetToken`. This allows us to get a password reset token by email address, therefore in the tests we can mimic "clicking URL with reset token".

## Other improvements
- Create new API helper, `encryptPassword` to be DRY, now that we're encrypting passwords in multiple places.
- Update content-strings to have a shared `PASSWORD_HINT`, to be DRY.
- Create a new cypress commands to be DRY:
  - `cy.completeAndSubmitPasswordResetForm`
  - `cy.completeAndSubmitNewPasswordAccountForm`
  - `cy.assertPasswordLabelHintAndInput`
- Simplify some account password related E2E test's `before()` commands - we don't need to go through a full UX journey of eligibility in all the account password tests.


src/ui/server/controllers/insurance/account/password-reset/success/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Add the success flag to req.flash
+   * This prevents an issue where the success flag will be undefined if:
+   * - The page is refreshed
+   * - A user navigates away and comes back to the page
+   */
+  req.flash('passwordResetSuccess', String(passwordResetSuccess));
+
+  const success = passwordResetSuccess || req.flash('passwordResetSuccess');
```
Zainzzkk, line 35: will the flash always be consumed here? or does it need to be consumed so it is cleared from flash?

ttbarnes: It won't always be consumed, but solves a scenario where if a user refreshes the page or submits the analytics banner on this page, the success flag will still be defined - so the page will still render.

If however say this URL is visited randomly and there is no success flag, we redirect to the password reset route

src/ui/templates/insurance/account/password-reset/new-password.njk
```diff
@@ -0,0 +1,96 @@
...
+            {% for rule in FIELD.HINT.RULES %}
+              <li data-cy="{{ FIELD.ID }}-hint-list-item-{{ loop.index }}">{{ rule }}</li>
+            {% endfor %}
+          </ul>
+
+        </div>
+      </div>
+    {% endset %}
+
+    {{ govukInput({
```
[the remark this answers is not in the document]
ttbarnes: Great shout, made a note to do this

Zainzzkk (approved):

ttbarnes:
Merging this into to progress the sprint / PRs - will have a session with Abhi running through all password reset scenarios, can make any changes after that if required

PR #447, opened 2023-04-27
feat(EMS-1092-966): Account - Password reset - sign in, link/token expiry UX

ttbarnes:
This PR enhances/improves the Password reset journey to ensure a user can login after password reset and that there is a nice UX with clear messaging, if a password reset token/link is invalid.

## Changes
- Create a new GQL query resolver, `verifyAccountPasswordResetToken`. This accepts a token, checks if an account exists with the provided token and does not surpass the expiry time.
- Update "new password" UI controller to call the new GQL query resolver. Only render the page/form if the token is valid, otherwise go to "link expired" page.
- Create a new UI route and controller for "link expired" page. **Note** the "send new link" button intentionally does not do anything at the moment.
- Add E2E test coverage for:
  - Signing into an account after resetting password.
  - Navigating to the "new password" page with an expired token, redirects to a new "link expired" page.
  - Navigating to the "new password" page with an invalid token, redirects to a new "link expired" page.

## Other improvements
- Simplify various E2E calls to the API for `getAccountPasswordResetToken` and account OTP so that comments and used email address are minimal and not repeated.


e2e-tests/cypress/e2e/journeys/insurance/account/password-reset/link-expired/account-password-reset-link-expired.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      const [firstExporter] = data.exporters;
+      const exporter = firstExporter;
+
+      /**
+       * Update the exporter's password reset verification expiry date via the API,
+       * so that we can mimic missing the verification period.
+       */
+      const now = new Date();
+
+      const milliseconds = 300000;
```
Zainzzkk, line 61: could we make this a constant?

ttbarnes: Good shout - can we do this after this PR, there are 4 instances of this so can do all in one go. There's a few tech improvements need to do, can do all in one PR

src/ui/templates/insurance/account/password-reset/link-expired.njk
```diff
@@ -0,0 +1,23 @@
+{% extends 'index.njk' %}
+
+{% block pageTitle %}
```
Zainzzkk, line 3: do we need a back button on these pages?

ttbarnes: No, if a link (from email) has expired or is invalid we redirect to this page. A back link would not have a HREF unless the page is refreshed. Have discussed with design

Zainzzkk (approved):

PR #450, opened 2023-04-28
fix(EMS-1506-1528-1536): Fix application submission and page specific header/back link issues

ttbarnes:
This PR fixes several issues/bugs:

- Application submission would fail when answering "no" to a specific declaration question.
- Accessibility statement displayed "[Object Object]" in the header after branch merges.
- After account sign out, the sign in page would render a back link which would never work because the session has been deleted.

## Changes
- Update `getTrueAnswerProperties` helper function to check for "No" field values as well as "Yes", therefore fixing the application submission issue. This helper is used in the "can submit application" helper function.
- Add E2E test coverage to ensure application submission works when submitting "no" to a specific declaration question. This involves updating various cypress commands so we can submit the declaration question as "yes" or "no", depending on the new `exportingWithCodeOfConduct` flag.
- Update account sign in controller/page to conditionally render the back link, i.e it now longer renders immediately after a succesful sign out.
- Update accessibility statement controller to use `corePageVariables` and therefore fixing the service header. Also added missing E2E test coverage.


e2e-tests/cypress/support/insurance/complete-declarations.js
```diff
@@ -1,17 +1,33 @@
...
   task.link().click();
 
   cy.completeAndSubmitDeclarationConfidentiality();
+
   cy.completeAndSubmitDeclarationAntiBribery();
+
   cy.completeAndSubmitDeclarationAntiBriberyCodeOfConduct();
-  cy.completeAndSubmitDeclarationAntiBriberyExportingWithCodeOfConduct();
+
+  if (exportingWithCodeOfConduct === true) {
```
Zainzzkk, line 23: could this just be if `(exportingWithCodeOfConduct) {`

ttbarnes: Will do this in the next PR if that's OK, so we can get this merged and deployed?

Zainzzkk (approved):

PR #454, opened 2023-05-02
feat(EMS-1474-1572): Account - Password reset - display error if account does not exist

ttbarnes:
This PR updates the password reset form submission to render a validation error if an email is valid, but an account does not exist.

## Changes
- Create new validation rule/function: `accountDoesNotExistValidation`.
- Update password reset POST controller to render the new validation error message if the API does not return `{ success: true}`.
- Add E2E test coverage, remove previous test - now has the same coverage, but simpler.

## Other improvements

- Fix incorrect error message reference in the E2E test.
- Fixed some typos (including EMS-1572).

### DIFF
--- a/e2e-tests/content-strings/error-messages.js
+++ b/e2e-tests/content-strings/error-messages.js
@@ -279,7 +279,7 @@ export const ERROR_MESSAGES = {
       },
       PASSWORD_RESET: {
         [FIELD_IDS.INSURANCE.ACCOUNT.EMAIL]: {
-          INCORRECT_FORMAT: 'Enter your email address in the correct format, like name@example.com',
+          INCORRECT_FORMAT: 'Enter an email address in the correct format, like name@example.com',
           ACCOUNT_DOES_NOT_EXIST: 'Enter the email address used when creating your account',
         },
       },


abhi-markan (approved):

src/ui/server/controllers/insurance/account/password-reset/index.ts
```diff
@@ -85,7 +86,18 @@ export const post = async (req: Request, res: Response) => {
       return res.redirect(LINK_SENT);
     }
 
-    return res.redirect(ROUTES.PROBLEM_WITH_SERVICE);
+    // no account found
+    validationErrors = accountDoesNotExistValidation();
```
Zainzzkk, line 90: I don't fully understand what this is for. Some more documentation?

Zainzzkk (commented):
Just one comment

src/ui/server/controllers/insurance/account/password-reset/index.ts
ttbarnes: OK, will add documentation soon - have some other tech improvements todo, added to the list.

This is to render a validation error on the page if no account is found

PR #456, opened 2023-05-03
fix(EMS-1347): Account sign in - OTP form submission without completing eligibility

ttbarnes:
This PR fixes an issue where submitting the "enter code" form without completing eligibility would result in a redirect to the "problem with service" page as it was trying to create an application without the required eligibility data.

## Changes

- Update the "enter code" controller to only call `createApplication` if the eligibility data is has values.
- Add E2E test coverage.
- Create new E2E command, `completeAndSubmitEnterCodeAccountForm`.


src/ui/server/controllers/insurance/account/sign-in/enter-code/index.test.ts
```diff
@@ -261,6 +261,16 @@ describe('controllers/insurance/account/sign-in/enter-code', () => {
         });
       });
 
+      describe('when req.session.submittedData.insuranceEligibility is an empty object', () => {
```
Zainzzkk, line 264: is there a test to check it goes to the right page (problem with service) if empty object?

PR #457, opened 2023-05-03
fix(EMS-1186): Your buyer -  Check your answers - Address multiple lines

Zainzzkk:
# Issue:
* Address was not showing on multiple lines on check your answers for your buyer

# Changes made: 
* Replaced `\r\n` with `<br/>` in ui to display address on multiple lines
* Enabled hot reloading for docker

### DIFF
--- a/src/ui/server/constants/index.ts
+++ b/src/ui/server/constants/index.ts
@@ -17,3 +17,4 @@ export * from './ukef-contact-details';
 export const INSURANCE = 'Insurance';
 export const QUOTE = 'Quote';
 export const SERVICE_NAME = 'EXIP';
+export const SECURE_OPTION_COOKIE = '__Secure-optionalCookies';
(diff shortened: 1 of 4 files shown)


src/ui/server/helpers/summary-lists/your-buyer/company-or-organisation/index.ts
```diff
@@ -26,8 +26,11 @@ const {
  * @returns {Object}
  */
 export const generateAddressObject = (answers: ApplicationBuyer) => {
+  // replace new lines with line breaks to display in summary list
+  const address = answers[ADDRESS].replaceAll('\r\n', '<br />');
```
ttbarnes, line 30: Could this be a helper function so we can easily re-use in other places if required?

Also is there a possible scenario where there is only `\r` and no `\n`, and vice versa?

Zainzzkk: from the component, no.  All line breaks entered into the textarea seem to be` \r\n` if a new line is added

src/ui/server/test-mocks/mock-buyer.ts
```diff
@@ -4,7 +4,7 @@ dotenv.config();
 
 const mockBuyer = {
   companyOrOrganisationName: 'Test name',
-  address: 'Test address',
+  address: 'Test \r\n address',
```
ttbarnes, line 7: Not a blocker - a more complete/populated mock address e.g 3 lines country etc would be great

ttbarnes (approved):

src/ui/server/test-mocks/mock-buyer.ts
Zainzzkk: will do in next PR

PR #458, opened 2023-05-03
fix(EMS-1514-1339-1374-1516): Cookies - added Secure-optionalCookies to cookies check

Zainzzkk:
# Issue: 
* Dev would say cookies rejected when accepted and did not populate with correct cookie choice in cookies page

# Changes made:
* Added check for Secure-optionalCookies or optionalCookies in middleware
* Added check in controller for  Secure-optionalCookies or optionalCookies
* Added tests

### DIFF
--- a/src/ui/server/controllers/root/cookies/index.ts
+++ b/src/ui/server/controllers/root/cookies/index.ts
@@ -21,15 +21,11 @@ export const get = (req: Request, res: Response) => {
   // store the previous URL so that we can use this in the POST res.render.
   req.flash('previousUrl', req.headers.referer);
 
-  // TODO: Remove after debug
-  // eslint-disable-next-line no-console
-  console.log('cookies page', req.cookies);
-
   return res.render(TEMPLATE, {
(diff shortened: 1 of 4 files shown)


src/ui/server/controllers/root/cookies/index.ts
```diff
@@ -60,7 +56,7 @@ export const post = (req: Request, res: Response) => {
     userName: getUserNameFromSession(req.session.user),
     ...singleInputPageVariables({ ...PAGE_VARIABLES, BACK_LINK: backLink, START_ROUTE: startRoute }),
     FIELD: FIELDS[FIELD_IDS.OPTIONAL_COOKIES],
-    submittedValue: req.cookies.optionalCookies,
+    submittedValue: req.cookies.optionalCookies || req.cookies['__Secure-optionalCookies'],
```
ttbarnes, line 59: Great work - not a blocker can be done afterwards - how about a constant for this cookie name?

ttbarnes (approved):

ttbarnes (approved):

PR #459, opened 2023-05-03
feat(EMS-1348): Application status - disallow access to "check answers" pages if previous sections are incomplete

ttbarnes:
This PR prevents access to any "check your answers" pages, if the previous required sections are not completed. Now, if a user manually navigates to one of these pages in this scenario, they are redirected to a new page, "complete other sections".

### New page
 - Setup a new route and controller for "complete other sections" page.

### Application status checks
 - Update `section-review` `requiredFields` helper function to consume all fields from previous sections. This was previously  incorrectly referencing section names instead of section fields.
 - Create new helper function `canAccessCheckYourAnswersRoutes`. This checks uses previously created helpers to check if the fields required to access "check your answers" have been completed.
 - Update `applicationStatus` middleware to run an additional check for the check your answers routes/application completion. If "check your answers" cannot be accessed, a user is redirected to the new "complete other sections" page.
 - Create a new middleware helper function, `isCheckYourAnswersRoute`. This maps all appropriate "check your answers" routes into an array, prefixed with the application's reference number. This allows us to check a current URL without stripping anything.
 - Update `getApplication` middleware to list the new "complete other sections" route as a relevant route for application data fetching.
 - Add E2E test coverage for the new page and a user trying to access a "check your answers" page with an brand new application.

### Necessary "policy and exports" changes

Previously, if a user proceeds to the "insurance - check your answers" and changes a policy type from single to multiple (or vice versa), on form submission, the user would be taken back to the check your answers page they came from. 

However, with the these changes - this will be prevented, because if an application's policy type is changed without completing single/multiple specific required fields, the application's "prepare your application" data is incomplete and therefore the new middleware/helpers will pick this up.

Therefore, the policy and exports "type of policy" POST controller has been updated so that it will redirect to the next part of the flow instead of going back to the "check your answers" page, which would otherwise (correctly) result in a redirection to the "complete the rest of the application".

- Update/simplify "type of policy" POST controller redirect.
- Update and improve E2E tests that change a single contract to a multiple and vice versa, from the "insurance - check your answers - policy and exports" page.

## Other improvements

 - Create a new cypress command, `assertUrl`.
 - Improved some unit test descriptions.


e2e-tests/constants/routes/insurance/check-your-answers.js
```diff
@@ -1,4 +1,4 @@
-const ROOT = '/check-your-answers';
+export const ROOT = '/check-your-answers';
```
Zainzzkk, line 1: is ROOT not exported below?

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #462, opened 2023-05-04
feat(EMS-1348): Application status - disallow access to "declarations" pages if previous sections are incomplete

ttbarnes:
This PR prevents access to any "declarations" pages, if the previous required sections are not completed. Now, if a user manually navigates to one of these pages in this scenario, they are redirected to the "complete other sections".

## Changes
- Update/simplify `applicationIsComplete` and `required-fields` helper functions to consume `requiredPrepareApplicationFields`.
- Rename/repurpose `canAccessCheckYourAnswersRoutes` to be `canAccessSubmitYourApplicationRoutes`, as we need to do the same checks for different routes in the "submit your application" section/routes.
- Rename/repurpose `isCheckYourAnswersRoute` to be `isSubmitYourApplicationRoute`, update application status middleware.
- Add E2E test coverage.
- Remove now irrelevant E2E test where we test skipping the flow and trying to submit the application (this PR prevents a user from getting to the final page/form before submission).
- Add E2E test coverage for completing the "prepare your application" group and that all task statuses in this group and the "submit application" groups are correct at this stage in application journey.


src/ui/server/helpers/can-access-submit-your-application-routes/index.ts
```diff
@@ -1,17 +1,18 @@
...
 import sectionStatus from '../section-status';
 import { TASKS } from '../../content-strings';
 import { Application } from '../../../types';
 
 /**
- * canAccessCheckYourAnswersRoutes
- * Check if an application has submitted previous fields/sections required before accessing the "check your answers" section/routes.
+ * canAccessSubmitYourApplicationRoutes
+ * Check if an application has submitted previous fields/sections required before accessing the "submit your application" group/sections/routes.
+ * The "submit your application" group includes all routes for "check your answers" and "declarations".
```
Zainzzkk, line 10: do we not still need one for declarations and one for check your answers?

ttbarnes: No, this function is only used to check the "prepare application" group so it now consumes both check your answers/section review routes as well as declarations.

Both of these sections for the "submit application" group have the same requirement - both of them can only be unlocked/visited when the "prepare application" group is completed.

Zainzzkk (approved):

PR #463, opened 2023-05-04
fix(EMS-1254): Application submission - fix incorrect flag used for sending emails

ttbarnes:
This PR fixes an issue where an incorrect flag was being used in a condition to trigger a particular email template from being sent as part of application submission.

## Changes
- Rename usage of `buyer.exporterIsConnectedWithBuyer` with `buyer.exporterHasTradedWithBuyer` in the `getApplicationSubmittedEmailTemplateIds` function.
- Update `exporterHasTradedWithBuyer` mock data/answer to reflect what is actually stored in the DB.

### DIFF
--- a/src/api/.keystone/config.js
+++ b/src/api/.keystone/config.js
@@ -2351,7 +2351,7 @@ var getApplicationSubmittedEmailTemplateIds = (application) => {
     templateIds.exporter = EXPORTER.SEND_DOCUMENTS.ANTI_BRIBERY;
     templateIds.underwritingTeam = UNDERWRITING_TEAM.NOTIFICATION_ANTI_BRIBERY;
   }
-  const isConnectedWithBuyer = buyer.exporterIsConnectedWithBuyer && buyer.exporterIsConnectedWithBuyer === ANSWERS.YES;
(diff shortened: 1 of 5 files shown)


src/api/.keystone/config.js
```diff
@@ -2351,7 +2351,7 @@ var getApplicationSubmittedEmailTemplateIds = (application) => {
     templateIds.exporter = EXPORTER.SEND_DOCUMENTS.ANTI_BRIBERY;
     templateIds.underwritingTeam = UNDERWRITING_TEAM.NOTIFICATION_ANTI_BRIBERY;
   }
-  const isConnectedWithBuyer = buyer.exporterIsConnectedWithBuyer && buyer.exporterIsConnectedWithBuyer === ANSWERS.YES;
+  const isConnectedWithBuyer = buyer.exporterHasTradedWithBuyer && buyer.exporterHasTradedWithBuyer === ANSWERS.YES;
```
Zainzzkk, line 2354: does this happen anywhere else? we could have a helper for answerIsYes and answerIsNo

ttbarnes: Great idea, it happens in only one other place - CSV mapping for underwriters. We have some tickets to fix some CSV things, could create a helper then if that's OK? So we can progress this ticket, has been floating around for a while

Zainzzkk (approved):

PR #475, opened 2023-05-11
feat(EMS-1087): Account sign in - Reset verification expiry if account is unverified

ttbarnes:
This PR improves a the account sign in process for unverified users. If a user has not verified their email during the account creation process, when they attempt to sign in, we now do the following:

- If an account has a verification hash/token and the expiry period has not passed:
  - Reset the verification period.
  - Resend the "verify your email address" email (with the token).
  - In the UI, redirect to the user to the "new link has been sent" page.

## Changes
- Update account sign in resolver to execute the above checks and return a new flag, `resentVerificationEmail` for the UI. Also improved documentation and logging.
- Extract the email sending logic in `sendEmailConfirmEmailAddress` resolver, into a new helper function to be DRY as it is now executed in multiple places.
- Update the UI to redirect to `CONFIRM_EMAIL_RESENT` route with account ID param, if a `resentVerificationEmail` flag is returned from the API.
- Update "form submission with unverified account" E2E test.
- Update the `completeAndSubmitSignInAccountForm` cypress command so it accepts params as an object. We can therefore easily add a new param `assertSuccessUrl` and this allows us to _not_ assert the sucess URL in the unverified account E2E test, as it does not redirec to the success URL.

### DIFF
--- a/src/api/custom-resolvers/mutations/account-sign-in.test.ts
+++ b/src/api/custom-resolvers/mutations/account-sign-in.test.ts
@@ -62,9 +62,6 @@ describe('custom-resolvers/account-sign-in', () => {
 
     jest.resetAllMocks();
 
-    // sendConfirmEmailAddressEmailSpy = jest.fn(() => Promise.resolve(mockSendEmailResponse));
-    // confirmEmailAddressEmail.send = sendConfirmEmailAddressEmailSpy;
-
     securityCodeEmailSpy = jest.fn(() => Promise.resolve(mockSendEmailResponse));
     sendEmail.securityCodeEmail = securityCodeEmailSpy;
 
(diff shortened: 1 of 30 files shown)


src/api/custom-resolvers/mutations/account-sign-in.test.ts
```diff
@@ -93,26 +100,98 @@ describe('custom-resolvers/account-sign-in', () => {
...
+          isVerified: false,
+          verificationHash: mockAccount.verificationHash,
+        },
+      });
 
       result = await accountSignIn({}, variables, context);
+    });
 
-      const expected = { success: false };
+    test('it should reset the account`s verificationExpiry', async () => {
```
Zainzzkk, line 121: would this be better being split into 2 test()?

ttbarnes: Great shout, have split into 2 tests and simplified the variables.

Zainzzkk (approved):

Zainzzkk (approved):

PR #477, opened 2023-05-12
fix(EMS-1577-1590): Feedback - Refactor and additional email variables

Zainzzkk:
# Changes made:
* Added additional fields to email - date, time, referralUrl, service and product
* Refactored ui feedback call to call one create endpoint
* Refactored api feedback to have custom resolver which creates feedback and sends email
* Updated tests


src/api/custom-resolvers/index.ts
```diff
@@ -11,7 +11,7 @@ import {
   deleteApplicationByReferenceNumber,
   updateCompanyAndCompanyAddress,
   submitApplication,
-  sendEmailInsuranceFeedback,
+  createInsuranceFeedbackAndEmail,
```
ttbarnes, line 14: How about just `createFeedback` or `createInsuranceFeedback`?

Will the quote feedback data and email be any different?

src/api/custom-resolvers/mutations/create-feedback.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    it('should generate and return the created feedback', () => {
+      expect(feedbackResponse.service).toEqual(variables.service);
+      expect(feedbackResponse.satisfaction).toEqual(variables.satisfaction);
+      expect(feedbackResponse.improvement).toEqual(variables.improvement);
+      expect(feedbackResponse.otherComments).toEqual(variables.otherComments);
+      expect(feedbackResponse.referralUrl).toEqual(variables.referralUrl);
+      expect(feedbackResponse.product).toEqual(variables.product);
+      expect(feedbackResponse.success).toEqual(true);
```
ttbarnes, line 50: Can we destructure `feedbackResponse`?

src/api/custom-resolvers/mutations/create-feedback.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    test('should throw an error', async () => {
+      try {
+        feedbackResponse = (await createInsuranceFeedbackAndEmail({}, variables, context)) as Feedback;
+      } catch (err) {
+        const expected = new Error(`Creating feedback: ${mockErrorMessage}`);
+        expect(err).toEqual(expected);
+      }
+    });
+  });
```
ttbarnes, line 98: Can we align the error assertion descriptions with other tests?

```js
describe('when the account ...', () => {
  // ..
  test('it should return success=false', async () => {
    // ..

```

src/api/custom-resolvers/index.ts
Zainzzkk: I am not sure yet, it may have extra fields from the design

src/api/custom-resolvers/mutations/create-feedback.ts
```diff
@@ -0,0 +1,45 @@
...
+import sendEmail from '../../emails';
+
+/**
+ * creates feedback in database and sends email
+ * adds createdAt timestamp to feedback
+ * @param root
+ * @param {InsuranceFeedbackVariables} variables
+ * @param context
+ * @returns {Object} with success true or false and response
+ */
```
ttbarnes, line 12: In other places we're describing like this:

```js
* @param {Object} GraphQL root variables
* @param {Object} GraphQL variables for the AccountSignInNewCode mutation
* @param {Object} KeystoneJS context API
```

src/api/custom-resolvers/mutations/create-feedback.ts
```diff
@@ -0,0 +1,45 @@
...
+const createInsuranceFeedbackAndEmail = async (root: any, variables: InsuranceFeedbackVariables, context: Context) => {
+  console.info('Creating feedback');
```
ttbarnes, line 14: And in the catch handler, should it be "insurance feedback"?

src/api/custom-schema/type-defs.ts
```diff
@@ -205,10 +205,13 @@ const typeDefs = `
     ): SuccessResponse
 
     """ send email for insurance feedback """
-    sendEmailInsuranceFeedback(
+    createInsuranceFeedbackAndEmail(
```
ttbarnes, line 208: Can we add "create" to the description here just so it's clear

src/api/types.ts
```diff
@@ -319,6 +323,8 @@ interface Feedback {
   otherComments: string;
   referralUrl: string;
   product: string;
+  createdAt: string;
+  success: boolean;
```
ttbarnes, line 327: I don't think we need `success` here, could the resolver just use this as a return type? Unless we need to return the whole feedback object

```js
Promise<SuccessResponse>
```

Zainzzkk: its useful for the test to check if feedback has been generated? won't be able to test it otherwise in API

ttbarnes: I see, in the custom resolver can you add the return type?

`Promise<Feedback>`

src/api/custom-resolvers/mutations/create-feedback.ts
Zainzzkk: keeping it generic for all feedback

src/api/custom-resolvers/index.ts
Zainzzkk: it cannot be createFeedback as that clashes with the autogenerated resolver by keystone.  Set to `createFeedbackAndEmail` so is generic

ttbarnes: OK great, `createFeedbackAndSendEmail` would be ideal, not a blocker

src/api/types.ts
ttbarnes: Actually don't need to add `success: boolean;` here as it already gets it from `SuccessResponse` - not a blocker if we remove this in next PR, so we can get this merged and progress the sprint

ttbarnes (approved):

PR #481, opened 2023-05-16
fix(EMS-1615): Account verification - resend "confirm email address" email

ttbarnes:
This PR fixes a couple of issues caused by email updates made in https://github.com/UK-Export-Finance/exip/pull/480

## Changes
- Update `accountSignIn` GQL mutation to accept `urlOrigin`.
- Crate a new UI route and controller, `/resend-confirm-email`, this calls the API to resend the email and if successful, redirect to "confirm email resent".
- Update "confirm email resent" UI controller to get an account (to get the email), instead of sending the "confirm email address" email (the email is now sent prior to getting to this controller).
- Re-enable temporarily disabled E2E tests.


src/ui/server/controllers/insurance/account/create/confirm-email-resent/index.ts
```diff
@@ -27,18 +27,14 @@ export const get = async (req: Request, res: Response) => {
...
+    const accountResponse = await api.keystone.account.get(id);
 
-    const account = await api.keystone.account.sendEmailConfirmEmailAddress(urlOrigin, id);
+    let accountEmail;
 
-    if (!account.success) {
-      console.error("Error sending new email verification for account creation and rendering 'new link sent' page");
-
-      return res.redirect(ROUTES.PROBLEM_WITH_SERVICE);
+    if (accountResponse && accountResponse.email) {
```
Zainzzkk, line 34: if there's no account response, should it not immediately go to problem with service?

ttbarnes: Great shout, i've tweaked the controller to account for this and if there is no account response or email in the returned response, redirect to problem with service

Zainzzkk (approved):

PR #482, opened 2023-05-16
feat(EMS-1513): Your business - Contact - Save

Zainzzkk:
# Changes made:
* Added tables for contact and linked to business table
* Added ability to save contact details to db
* Added save and go back functionality
* Added prepopulation of fields from req.session if not in database
* Added helper to get data from session or database for specific field


src/api/schema.ts
```diff
@@ -528,6 +542,16 @@ export const lists = {
     },
     access: allowAll,
   }),
+  BusinessContactDetail: list({
```
ttbarnes, line 545: What about `BusinessContact` instead? What's the "Detail" for?

src/api/types.ts
```diff
@@ -125,7 +129,7 @@ interface Application {
   company: ApplicationCompany;
   companySicCodes: Array<ApplicationCompanySicCode>;
   companyAddress: ApplicationCompanyAddress;
-  business: ApplicationRelationship;
+  business: Business;
```
ttbarnes, line 132: `ApplicationBusiness` so it's aligned with other types?

src/ui/server/api/keystone/application/index.ts
```diff
@@ -200,6 +201,36 @@ const application = {
         throw new Error('Updating application business');
       }
     },
+    businessContact: async (id: string, update: object) => {
+      try {
+        console.info('Updating application business contact details');
```
ttbarnes, line 206: Can remove "details" from this log? The other logs don't have it

src/ui/server/controllers/insurance/business/contact/index.test.ts
```diff
@@ -138,19 +144,31 @@ describe('controllers/insurance/business/contact', () => {
     });
 
     describe('when there are no validation errors', () => {
+      const body = {
+        [FIRST_NAME]: 'firstName',
+        [LAST_NAME]: 'lastName',
+        [EMAIL]: 'test@test.com',
+        [POSITION]: 'CEO',
+      };
```
ttbarnes, line 152: Could we use test mocks for this?

src/ui/server/controllers/insurance/business/contact/index.ts
```diff
@@ -74,6 +77,7 @@ const get = (req: Request, res: Response) => {
       }),
       userName: getUserNameFromSession(req.session.user),
       application: mapApplicationToFormFields(application),
+      applicationAnswers: getFromSessionOrApplication(application, BUSINESS, BUSINESS_CONTACT_DETAIL, req.session.user),
```
ttbarnes, line 80: What do you think about calling this `submittedValues`, same as other places? Also `getFromSessionOrApplication` doesn't feel descriptive enough to me, how about `getSubmittedValues()`?

src/ui/server/controllers/insurance/business/contact/save-and-back/index.ts
```diff
@@ -0,0 +1,44 @@
+import { Request, Response } from '../../../../../../types';
+import { ROUTES } from '../../../../../constants';
+import generateValidationErrors from '../validation';
+import mapAndSave from '../../map-and-save';
+
+const { INSURANCE_ROOT, ALL_SECTIONS } = ROUTES.INSURANCE;
+
+/**
+ * saves and goes back to all sections from contact page unless there are database errors
```
ttbarnes, line 9: `API` errors 😛 no DB things here

src/ui/server/controllers/insurance/business/save-data/index.ts
```diff
@@ -83,8 +83,27 @@ const broker = async (application: Application, formBody: RequestBody, errorList
...
+  const sanitisedData = sanitiseData(dataToSave);
+
+  const brokerId = application.business?.businessContactDetail?.id;
+
+  try {
+    // send the form data to the API for database update.
+    const saveResponse = await api.keystone.application.update.businessContact(brokerId, sanitisedData);
+    return saveResponse;
+  } catch (err) {
+    throw new Error("Updating business' contact");
```
ttbarnes, line 100: typo

src/ui/server/helpers/object/index.ts
```diff
@@ -9,6 +9,10 @@ type ObjectType = {
  * @returns {Boolean}
  */
 export const objectHasKeysAndValues = (obj: ObjectType) => {
+  if (!obj) {
+    return false;
+  }
```
ttbarnes, line 14: I don't think this is needed, `obj` is a required param so if it's not provided it won't compile

src/ui/server/helpers/get-from-session-or-application/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param {String} field eg. businessContactDetails
+ * @param {RequestSessionUser} session
+ * @returns {Object} session or application section
+ */
+const getFromSessionOrApplication = (application: Application, section: string, field: string, session?: RequestSessionUser) => {
+  if (!application[section] || !application[section][field]) {
+    return session;
+  }
+
+  const { id, __typename, ...sectionObject } = application[section][field];
```
ttbarnes, line 19: Can we add a condition and test to make sure `application[section]` exists? Currently if it doesn't this will error

src/ui/server/helpers/object/index.ts
Zainzzkk: I can remove this.  I had it crash on me once when for some reason when I passed the wrong variable and it got undefined and went to problem with service.  What do you say?

src/ui/server/helpers/get-from-session-or-application/index.ts
Zainzzkk: line 15.  It should return the session if application[section] does not exist.

And ts complains about the `submittedData `in `RequestSession` which is missing when I pass it req.session.user

src/ui/server/helpers/object/index.ts
ttbarnes: I think it's safe to remove, I don't believe it's possible for it to error if object is not passed - if the function is called without an object it will not compile. Unless you can replicate it erroring seems ok to remove

src/ui/server/controllers/insurance/business/contact/index.ts
ttbarnes: I think `submittedValues` makes sense even if that name is primarily used in post's - `applicationAnswers`doesn't feel right because it implies it will be from an application when it's not. I can't think of a more appropriate name, what do you think?

src/ui/server/helpers/get-from-session-or-application/index.ts
ttbarnes: OK makes sense, sorry didn't see that.

RE the session in this case can we just rename `session` to `userSession`? and the file name as `getFromUserSessionOrApplication` just so it's clearer

src/api/schema.ts
Zainzzkk: discussed.  Initially had agreed on `BusinessContactDetail`

src/ui/server/api/keystone/application/index.ts
```diff
@@ -200,6 +201,36 @@ const application = {
...
+          return response.data.updateBusinessContactDetail;
+        }
+
+        console.error(response);
+        throw new Error('Updating application business');
+      } catch (err) {
+        console.error(err);
+        throw new Error('Updating application business');
+      }
+    },
```
ttbarnes, line 233: Not a blocker can be done afterwards, but the logs need to match - the first was says `business contact` but the others just say `business`, would be confusing if debugging

src/ui/server/test-mocks/mock-business-contact.ts
```diff
@@ -0,0 +1,17 @@
...
+import ACCOUNT_FIELD_IDS from '../constants/field-ids/insurance/account';
+
+dotenv.config();
+
+const { POSITION } = FIELD_IDS.CONTACT;
+const { FIRST_NAME, LAST_NAME, EMAIL } = ACCOUNT_FIELD_IDS;
+
+const mockBusinessContact = {
+  [FIRST_NAME]: 'firstName',
+  [LAST_NAME]: 'lastName',
```
ttbarnes, line 12: Not a blocker but can we call these anything other than the actual IDs themselves 🙃

ttbarnes (approved):
Couple of minor comments can be addressed after merge so we can progress this and the sprint

PR #483, opened 2023-05-16
feat(EMS-628): Eligibility - Save all buyer country answers in the session

ttbarnes:
This PR updates the "buyer country" form in both the quote and eligibility flows. Previously, if a user submitted a country that is not supported for an online quote or application - after being taken to an exit page, if the user goes back to the country page, the submitted country would not display.

Now, we save the country in the session even if it's not supported for an online quote or application and we ensure the required data / data shape is valid if the country in session has `canApplyOnline` with a true boolean.

### UI
- Update insurance eligibility and quote eligibility "buyer country" controllers to save the submitted buyer country to the session.
- Create new helper function `mapSubmittedEligibilityCountry` to map the country and a `canApplyOnline` flag into an object with BUYER_COUNTRY field.
- Update `requiredDataProvided` middleware function to specifically check for the `BUYER_COUNTRY` field and if the field is passefd, check if it has a `canApplyOnline` flag to ensure that the required data state is valid.

### E2E 
- Create new fixtures to use for countries that are supported/unsupported/supported via email, to avoid hard coding in tests.
- Update all buyer country tests to assert that the search input has a prepopulated value and search results when going back to the page via a back link.
- Rename buyer country page selectors `searchInput` to `input` to be aligned with all other inputs.

### Other improvements
- Simplify a condition in "confirm email resent" GET controller.


src/ui/server/controllers/insurance/eligibility/buyer-country/index.ts
```diff
@@ -115,10 +112,24 @@ export const post = async (req: Request, res: Response) => {
     }
 
     if (canApplyOffline(country)) {
```
Zainzzkk, line 114: not a blocker.  Can this use `if (applyOnline)`

ttbarnes: It can't unfortunately, it needs to specifically check if it can apply offline which is different to online. It checks for this field from CIS data: `shortTermCover`:

It will only return true if the field has one of these values:
```js
ILC
CILC
REFER
```

Zainzzkk (approved):

PR #487, opened 2023-05-18
fix(EMS-1568-1632-1633): Feedback - Max characters, no satisfaction and email variables

Zainzzkk:
# Issue:
* Redirected to problem with service if satisfaction not selected
* Redirected to problem with service if over 1000 characters provided in text fields in feedback
* email variable for satisfaction had db variable without spaces

# Changes made:
* Added mapping for satisfaction value provided to provide formatted version - eg. verySatisfied to Very satisfied
* Added db change to 1200 characters for improvement and otherComments fields
* Added option with empty string if no selection for satisfied


e2e-tests/cypress/e2e/journeys/insurance/feedback/submit-feedback.spec.js
```diff
@@ -72,4 +72,58 @@ context('Insurance - Feedback - Submit feedback form', () => {
...
+      it(`should redirect to ${FEEDBACK_SENT}`, () => {
+        cy.url().should('eq', feedbackConfirmationUrl);
+      });
+    });
+
+    describe('when clicking the "back to service button"', () => {
+      it(`should redirect to ${startUrl}`, () => {
+        submitButton().click();
+
+        cy.url().should('eq', startUrl);
```
ttbarnes, line 125: We can use new cypress command `cy.assertUrl(startUrl)` for everything now

e2e-tests/cypress/e2e/journeys/insurance/feedback/submit-feedback.spec.js
```diff
@@ -72,4 +72,58 @@ context('Insurance - Feedback - Submit feedback form', () => {
...
+  describe(`when submitting a populated form with maximum characters for ${IMPROVEMENT} and ${OTHER_COMMENTS}`, () => {
+    beforeEach(() => {
+      cy.saveSession();
+
+      cy.navigateToUrl(startUrl);
+      partials.phaseBanner.feedbackLink().click();
+
+      feedbackPage.field(SATISFIED).input().click();
+      cy.keyboardInput(feedbackPage.field(IMPROVEMENT).input(), 'a'.repeat(1200));
+      cy.keyboardInput(feedbackPage.field(OTHER_COMMENTS).input(), 'a'.repeat(1200));
```
ttbarnes, line 85: variable for `'a'.repeat(1200)` to be DRY?

src/api/helpers/map-feedback-satisfaction/index.ts
```diff
@@ -0,0 +1,12 @@
+import { FEEDBACK } from '../../constants';
+
+/**
+ * mapFeedbackSatisfaction
+ * maps satisfaction value from user selection
+ * returns formatted text value for email
+ * @param {String} satisfaction
+ * @returns {String} mapped satisfaction email text
+ */
+const mapFeedbackSatisfaction = (satisfaction: string) => FEEDBACK.EMAIL_TEXT[satisfaction];
```
ttbarnes, line 10: If a satisfaction is not found, will things break?

Zainzzkk: it should never not be found as the schema won't allow it to be submitted to the api.  And if null, then this won't be called

ttbarnes (approved):

PR #488, opened 2023-05-18
feat(EMS-1638): Account - Password reset - block account after a certain amount of requests in a period of time

ttbarnes:
This PR updates the password reset functionality to mark an account as blocked if the "request reset password link" form is submitted with a valid email/account that exists and is submitted more than a certain amount of times over a certain time period, as per security requirements.

### API / DB

- Create new tables, `AuthenticationRetry` and `Authencation`. Note: `Authentication` is currently not used but will be very soon.
- Add account constants for password reset retries threshold and the timeframe/period that we will check when the form/GQL resolver is called.
- Update `sendEmailPasswordResetLink` GQL resolver to:
  - Generate a retry entry via new helper function, `createAuthenticationRetryEntry`.
  - Check if an account has authentication retry entries.
  - If so, check if the amount of retries in a certain time period breach the threshold.
  - If the threshold if breached, update the account so that the `isBlocked` flag is true.
- Create a new helper function to check if an account should be blocked, `shouldBlockAccount`.

###  UI 

- Create a new "account suspended" page.
- Update the initial password reset controller to redirect to the new "account suspended" route, if the API response contains `isBlocked`.

### E2E

- Create new cypress command,  `completeAndSubmitPasswordResetFormMaximumRetries`.
- Update `completeAndSubmitPasswordResetForm` cypress command to optionall check for success URL.
- Change `deleteAccountsById` API call to `deleteAnAccount`. This also simplifies API consumption in the delete account cypress command, as we only need to make one call now.


e2e-tests/cypress/support/insurance/account/complete-and-submit-password-reset-form-maximum-retries.js
```diff
@@ -0,0 +1,24 @@
...
+
+const attemptsToMake = [...Array(ACCOUNT.MAX_PASSWORD_RESET_TRIES)];
+
+/**
+ * completeAndSubmitPasswordResetFormMaximumRetries
+ * Complete and submit the password reset form multiple times,
+ * Mathcing the maximum retries allowed before an account becomes blocked.
+ */
+const completeAndSubmitPasswordResetFormMaximumRetries = () => {
+  attemptsToMake.forEach((item, index) => {
```
Zainzzkk, line 13: would a for loop make more sense here?

e2e-tests/cypress/support/insurance/account/complete-and-submit-password-reset-form.js
```diff
@@ -16,12 +16,18 @@ const {
...
+ * @param {Object} Object with flags on how to complete password reset form.
+ */
+const completeAndSubmitPasswordResetForm = ({ assertRedirectUrl = true }) => {
   cy.keyboardInput(accountFormFields[EMAIL].input(), mockAccount[EMAIL]);
 
   submitButton().click();
 
-  cy.url().should('eq', linkSentUrl);
+  if (assertRedirectUrl) {
+    cy.url().should('eq', linkSentUrl);
```
Zainzzkk, line 29: assertUrl cy command?

src/ui/templates/insurance/account/suspended.njk
```diff
@@ -0,0 +1,15 @@
...
+  {{ CONTENT_STRINGS.PAGE_TITLE }}
+{% endblock %}
+
+{% block content %}
+
+  <h1 class="govuk-heading-xl" data-cy="heading">{{ CONTENT_STRINGS.PAGE_TITLE }}</h1>
+
+  <p class="govuk-body" data-cy="body">{{ CONTENT_STRINGS.BODY }} </p>
+
+  <a class="govuk-button" href="#" data-cy="reactivate-account-link-button">{{ CONTENT_STRINGS.BUTTONS.REACTIVATE_ACCOUNT }}</a>
```
Zainzzkk, line 13: could we put the href in CONTENT_STRINGS as # so we only need to change it there then?

e2e-tests/cypress/support/insurance/account/complete-and-submit-password-reset-form-maximum-retries.js
ttbarnes: I don't see a benefit to using a for loop instead of forEach? We're also using forEach throughout the E2E suite, no other instances of for loops

src/ui/templates/insurance/account/suspended.njk
ttbarnes: Good idea, but when this link is implemented it'll have to come from the variables passed from the controller as we'll need to pass dynamic query parameter. We do this in other account related pages

Zainzzkk (approved):

PR #498, opened 2023-05-25
feat(EMS-1672): Application submission - migrate CSV generation to XLSX

ttbarnes:
This PR changes the CSV that is sent to the underwriting team on application submission, to be an XLSX. This is achieved with `ExcelJS`, replacing `csv-stringify`.

## Changes
- Replace all references of "CSV" with "XLSX", move all CSV related files into XLSX file names/directories.
- Create `XLSX_CONFIG` constants.
- Create `generate.XLSX` function. This function:
  - Maps application data via existing mapping function.
  - Creates a new workbook via `ExcelJS`.
  - Adds a worksheet to the workbook.
  - Adds header columns (via `XLSX_HEADER_COLUMNS` - "Field", "Answer").
  - For each item in the application data array, add the item to the XLSX's worksheet as a new row.
  - Write the file to the `XLSX` directory and return the file path name.

## Other improvements
- Remove some now unnecessary temporary debugging logs.
- Improve `formatCurrency` function.
- Add a missing `cy.deleteAccount()` call in an E2E test.


src/api/generate-xlsx/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * XLSX
+ * Generate an XLSX file with exceljs
+ * @param {Object} Application
+ * @returns {String} File path
+ */
+const XLSX = (application: Application): Promise<string> => {
+  try {
+    console.info('Generating XLSX file');
```
Zainzzkk, line 14: To be done later, is it maybe worth putting `for: ${application.id}` at the end for nice logging?

src/api/generate-xlsx/map-application-to-XLSX/map-buyer/index.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param {Object} Application
+ * @returns {Array} Array of objects for XLSX generation
+ */
+const mapBuyer = (application: Application) => {
+  const { buyer } = application;
+
+  const mapped = [
+    xlsxRow(XLSX.SECTION_TITLES.BUYER, ''),
+    xlsxRow(XLSX.FIELDS[NAME], buyer[NAME]),
+    xlsxRow(String(CONTENT_STRINGS[ADDRESS].SUMMARY?.TITLE), buyer[ADDRESS]),
```
Zainzzkk, line 30: do we need `?.` here?

ttbarnes: Unfortunately yes, typescript thinks it's possibly undefined otherwise

src/api/generate-xlsx/index.ts
ttbarnes: Great idea i'll put this into the next PR

Zainzzkk (approved):

PR #505, opened 2023-05-26
chore(EMS-1616): Application versioning

ttbarnes:
As EXIP is iterated and integrates with other systems, it would be very useful to know what state an application could or could not be in.

For example, currently we only support contracts that fall under a certain amount of monetary value. In the future, this will change - there could be many iterations of the service.

Adding versioning to applications is very sensible because it allows us to easily, conditionally consume an application's data for example.

## Changes
- Add `version` field to the application schema.
- Restructure API constants for application to:
  - Define a list/array of application versions.
  - ⚠️ Specifically define the latest application version in a standalone manner - this and the above should be manually updated by a developer when EXIP moves into a new project phase.
  - Dynamically get the latest version number and add to the application.
  - Dynamically add some core key definitions for application support based on values returned from the application version (`TOTAL_VALUE_OF_CONTRACT`, `MAXIMUM_BUYER_CAN_OWE`).
- Create a helper function to get the latest application definition/version: `getApplicationDefinition`
- Rename "product" constants to "eligibility" since they are specific to eligibility.
- Update README


src/api/constants/application/get-application-definition/index.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    expect(result.VERSION_NUMBER).toEqual('1');
+  });
+
+  describe('when no application definition is found', () => {
+    it('should throw an error', () => {
+      try {
+        getApplicationDefinition('123456');
+      } catch (err) {
+        const expected = new Error('Unable to find latest application version');
+
```
abhi-markan, line 21: Console.error?

abhi-markan (changes requested):
Good work, one minor comment please.

src/api/constants/application/get-application-definition/index.test.ts
ttbarnes: Good shout have added (to the main file not the test)

abhi-markan (approved):

Zainzzkk (approved):

PR #512, opened 2023-06-01
feat(EMS-1678): Your Buyer - Company or organisation - Change country to be text. Add buyer country during application creation

Zainzzkk:
# Changes made:
* Buyer country populated on application creation
* Country is now a text which is populated from eligibility
* Removed country api calls from company-or-organisation controller
* Updated tests
* Changed test footer HREFs to be QUOTE_HREF


e2e-tests/cypress/e2e/journeys/insurance/check-your-answers/your-buyer/change-your-answers/change-your-answers-change-company-or-organisation.spec.js
```diff
@@ -174,7 +173,6 @@ context('Insurance - Check your answers - Company or organisation - Your buyer p
 
         // as html, cannot use checkText so checking contains following fields
         row.value().contains(fieldVariables.newValueInput);
-        row.value().contains(application.BUYER[COUNTRY]);
```
ttbarnes, line 177: Whlie you're in here could you fix typo in this E2E test and some others?

```js
Your buyer page-
```

Missing a space

src/api/helpers/get-populated-application/index.test.ts
```diff
@@ -51,15 +50,6 @@ describe('api/helpers/get-populated-application', () => {
     expect(result.declaration.id).toEqual(application.declaration.id);
   });
 
-  it('should return an application with populated buyer country', async () => {
```
ttbarnes, line 54: Why are we removing this test? The main function `getPopulatedApplication` has not changed?

src/ui/server/api/keystone/application/buyer.ts
```diff
@@ -0,0 +1,40 @@
...
+
+      if (response?.data?.updateBuyer) {
+        return response.data.updateBuyer;
+      }
+
+      console.error(response);
+      throw new Error('Updating application buyer country');
+    } catch (err) {
+      console.error(err);
+      throw new Error('Updating application buyer country');
```
ttbarnes, line 35: Can we change the logs to `Updating application buyer` since it could not be country specific

Zainzzkk: this is only being used for updating the country

src/ui/server/helpers/mappings/map-applications/index.ts
```diff
@@ -17,7 +17,6 @@ export const mapApplication = (application: Application) => {
     status,
     lastUpdated: formatDate(new Date(updatedAt)),
     referenceNumber,
-    buyerLocation: buyer?.country?.name || DEFAULT.EMPTY,
```
ttbarnes, line 20: I don't think this should be removed and should instead be reverted - this will remove the buyer location column from the dashbaord table?

src/ui/server/api/keystone/application/buyer.ts
ttbarnes: Yes currently, but the actual GQL call is accepting an object and the object then updates the whole buyer

Zainzzkk: changed

PR #513, opened 2023-06-01
fix(EMS-1589): Name fields rendering previously submitted special characters incorrectly

ttbarnes:
This PR fixes an issue where if a name field contained certain special characters and is saved in the DB - when revisiting the field in the UI, the special characters would render as character codes, instead of the originally submitted values/chracters. 

This was because before calling the API, we encode/sanitise some specific special characters.

## Changes
- Create new helper function in the UI: `replaceCharacterCodesWithCharacters`.
- Create a new mapping function, `mapNameFields`. This function calls `replaceCharacterCodesWithCharacters` on specific "name" fields.
- Update `mapApplicationToFormFields` mapping function  to call `mapNameFields`.
- Update `getUserNameFromSession` helper function to use `replaceCharacterCodesWithCharacters`. Meaning, the name displayed in the header of all pages, will display special characters correctly.
- Update all GET controllers that consume an application, to consume the application with `mapApplicationToFormFields` mappings instead of a raw application.
- Update `mapApplications` function to replace character codes for the `buyerName` field.
- Update `generateSummaryListRows` function to replace character codes for any row's value.
- Add E2E test coverage for entering and rendering name values with special characters, for the fields that specifically replace character codes with characters.

## Other improvements
- Update some cypress commands to accept object params:
  - `mapApplicationToFormFields`
  - `completeAndSubmitYourContact`
  - `completeAndSubmitCompanyOrOrganisationForm`
- Remove unnecessary UI unit test.
- Fixed a typo in an E2E fixture.

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/submit-name-fields-with-special-characters/submit-name-fields-with-special-characters-dashboard.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/submit-name-fields-with-special-characters/submit-name-fields-with-special-characters-dashboard.spec.js
@@ -1,0 +1,50 @@
+import { FIELD_IDS } from '../../../../../constants';
+import partials from '../../../partials';
+import dashboardPage from '../../../pages/insurance/dashboard';
+import mockApplication from '../../../../fixtures/application';
+import mockNameWithSpecialCharacters from '../../../../fixtures/name-with-special-characters';
+
+const {
+  INSURANCE: {
+    YOUR_BUYER: {
+      COMPANY_OR_ORGANISATION: { NAME: BUYER_NAME },
+    },
+  },
+} = FIELD_IDS;
+
+const { taskList } = partials.insurancePartials;
+
+const nameWithSpecialCharacters = mockNameWithSpecialCharacters(mockApplication.BUYER[BUYER_NAME]);
+
+context('Insurance - Name fields - Dashboard fields should render special characters without character codes after submission', () => {
+  let referenceNumber;
+
+  before(() => {
+    cy.completeSignInAndGoToApplication().then((refNumber) => {
+      referenceNumber = refNumber;
+
+      taskList.prepareApplication.tasks.buyer.link().click();
+
+      cy.completeAndSubmitCompanyOrOrganisationForm({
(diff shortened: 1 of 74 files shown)


e2e-tests/cypress/e2e/journeys/insurance/submit-name-fields-with-special-characters/submit-name-fields-with-special-characters-dashboard.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+  beforeEach(() => {
+    cy.saveSession();
+  });
+
+  after(() => {
+    cy.deleteApplication(referenceNumber);
+  });
+
+  it('should alkds kasl', () => {
```
Zainzzkk, line 44: this looks weird for the it statement

ttbarnes: Whoops! Sorry I missed this - was whilst getting it working. Fixed

src/ui/server/helpers/mappings/map-applications/index.ts
```diff
@@ -17,7 +18,7 @@ export const mapApplication = (application: Application) => {
     lastUpdated: formatDate(new Date(updatedAt)),
     referenceNumber,
     buyerLocation: buyer?.country?.name || DEFAULT.EMPTY,
-    buyerName: buyer.companyOrOrganisationName || DEFAULT.EMPTY,
+    buyerName: buyer.companyOrOrganisationName ? replaceCharacterCodesWithCharacters(buyer.companyOrOrganisationName) : DEFAULT.EMPTY,
```
Zainzzkk, line 21: could this not be `replaceCharacterCodesWithCharacters(buyer.companyOrOrganisationName) ?? DEFAULT.EMPTY`

ttbarnes: Great spot, it can now because the character code helper function param is optional. I've changed to

```js
replaceCharacterCodesWithCharacters(buyer.companyOrOrganisationName) || DEFAULT.EMPTY
```

Zainzzkk (approved):

PR #514, opened 2023-06-02
feat(deployment): refactored deployment script

abhi-markan:
## Introduction
Refactor deployment script

## Resolution
* Simplified deployment
* Segregated deployment environment whilst maintain generic deployment script

## Miscellaneous
* Updated documentation
* Updated dependencies


.github/workflows/deploy.sh
```diff
@@ -60,7 +60,7 @@ else
 fi
 
 #######################################
-# UKEF deployment shell script v0.0.6
-# 11/05/2023
+# UKEF deployment shell script v0.0.7
+# 31/05/2023
```
ttbarnes, line 64: 02/06/2023?

ttbarnes (approved):
LGTM

.github/workflows/deploy.sh
ttbarnes: Although looks like this version/script is not specific to EXIP so makes sense to keep the date the same as other usages

PR #522, opened 2023-06-06
feat(EMS-1710): Account sign in - update OTP expiry time

ttbarnes:
This PR extends the OTP expiration time. This is due to some user/accessibility issues where the OTP/security code would not arrive in an email inbox in time and therefore a user would not be able to sign in to their account.

## Changes

- Create `DATE_30_MINUTES_FROM_NOW` constant.
- Update `OTP.VERIFICATION_EXPIRY` constant.
- Update unit tests.

### DIFF
--- a/src/api/constants/index.ts
+++ b/src/api/constants/index.ts
@@ -33,6 +33,21 @@ const DATE_5_MINUTES_FROM_NOW = () => {
   return future;
 };
 
+/**
+ * DATE_30_MINUTES_FROM_NOW
+ * Generate a date that is 30 minutes from now
+ * @returns {Date}
+ */
+const DATE_30_MINUTES_FROM_NOW = () => {
+  const now = new Date();
+
+  const minutes = 30;
+
+  const future = new Date(now.getTime() + minutes * 60000);
+
+  return future;
+};
+
 export const ACCOUNT = {
   EMAIL: {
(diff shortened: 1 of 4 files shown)


src/api/constants/index.ts
```diff
@@ -33,6 +33,21 @@ const DATE_5_MINUTES_FROM_NOW = () => {
...
+ * DATE_30_MINUTES_FROM_NOW
+ * Generate a date that is 30 minutes from now
+ * @returns {Date}
+ */
+const DATE_30_MINUTES_FROM_NOW = () => {
+  const now = new Date();
+
+  const minutes = 30;
+
+  const future = new Date(now.getTime() + minutes * 60000);
```
abhi-markan, line 46: Can we add BODMAS to this please `now.getTime() + (minutes * 60000))`

ttbarnes: We could, but linting flags this as being pointless and asks to remove it.

abhi-markan (changes requested):
LGTM one minor comment please.

src/api/constants/index.ts
abhi-markan: Haha okay we can ignore it then.

abhi-markan (approved):

PR #528, opened 2023-06-07
fix(EMS-1718): Page not found - Split for insurance and generic

Zainzzkk:
# Issue:
* Page not found was always showing generic header 

# Changes made:
* checked if originalUrl contains insurance and renders insurance page not found if so
* Added helper for checking if quote route and set IS_GENERIC_HEADER to be false if quote route
* Used core-page-variables in index.ts for page-not-found
* Added e2e test for signed in/signed out page-not found on insurance side

### DIFF
--- a/src/ui/server/helpers/is-quote-route/index.ts
+++ b/src/ui/server/helpers/is-quote-route/index.ts
@@ -1,5 +1,13 @@
 import { QUOTE } from '../../constants';
 
+/**
+ * isQuoteRoute
+ * checks if url contains quote
+ * if it contains quote, then is a quote route and returns true
+ * if does not contain quote, then returns false
+ * @param {String} url of the page the user is on
+ * @returns {Boolean}
+ */
 const isQuoteRoute = (url: string) => {
   // splits on / in url
   const urlSplit = url.split('/');


e2e-tests/cypress/e2e/journeys/insurance/page-not-found-when-logged-out.spec.js
```diff
@@ -0,0 +1,36 @@
+import { PAGES } from '../../../../content-strings';
+import { ROUTES } from '../../../../constants';
+
+const CONTENT_STRINGS = PAGES.PAGE_NOT_FOUND_PAGE;
+
+context('Insurance - page not found - logged out', () => {
```
ttbarnes, line 6: Can we rename this file and test description to be "signed out" instead of "logged out"

src/ui/server/index.ts
```diff
@@ -144,23 +147,35 @@ const errorHandler: express.ErrorRequestHandler = (err, req, res, next) => {
...
+    return res.render(INSURANCE_PAGE_NOT_FOUND_TEMPLATE, {
+      ...corePageVariables({
+        PAGE_CONTENT_STRINGS: PAGES.PAGE_NOT_FOUND_PAGE,
+        ORIGINAL_URL: req.originalUrl,
+      }),
+      userName: getUserNameFromSession(req.session.user),
+    });
+  }
+
+  // all other non-insurance page not found should use generic page-not-found template
```
ttbarnes, line 168: How does it work for quote pages/routes?

ttbarnes: Should we not have something like `isQuoteRoute(...)`? which then has quote links in the header and footer?

Zainzzkk: done

ttbarnes (approved):

PR #533, opened 2023-06-09
fix(EMS-512): Insurance - Start - Fix and add extra wording

Zainzzkk:
# Changes made:
* Rectify text on insurance start page to match designs
* Add extra required text for sign in, about the service and quote under start button

### DIFF
--- a/e2e-tests/content-strings/pages/insurance/index.js
+++ b/e2e-tests/content-strings/pages/insurance/index.js
@@ -107,28 +107,28 @@ const START = {
   BODY_3: "You'll usually get a decision back within 7 working days, if we do not need to ask you any further questions.",
   BODY_4: 'If we need to gather extra information, it may take around 2 to 3 weeks.',
   SIGN_IN: {
-    BODY_1: 'You can',
+    YOU_CAN: 'You can',
     LINK: {
       TEXT: 'sign in to your account',
       HREF: ROUTES.INSURANCE.ACCOUNT.SIGN_IN.ROOT,
     },
-    BODY_2: "to continue an application, if you've already started one.",
+    TO_CONTINUE_APPLICATION: "to continue an application, if you've already started one.",
   },
   FIND_OUT_MORE: {
-    BODY_1: 'You can',
+    YOU_CAN: 'You can',
     LINK: {
       TEXT: 'find out more about UKEF export insurance',
       HREF: LINKS.EXTERNAL.GUIDANCE,
     },
-    BODY_2: ", if you want to know who's eligible and what's covered.",
+    TO_FIND_OUT_MORE: "if you want to know who's eligible and what's covered.",
   },
   QUOTE: {
-    BODY_1: 'You can also',
+    YOU_CAN: 'You can also',
     LINK: {
       TEXT: 'get an instant quote',
       HREF: ROUTES.QUOTE.START,
     },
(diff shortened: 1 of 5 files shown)


e2e-tests/content-strings/pages/insurance/index.js
```diff
@@ -104,8 +104,32 @@ const START = {
...
+  FIND_OUT_MORE: {
+    BODY_1: 'You can',
+    LINK: {
+      TEXT: 'find out more about UKEF export insurance',
+      HREF: LINKS.EXTERNAL.GUIDANCE,
+    },
+    BODY_2: ", if you want to know who's eligible and what's covered.",
+  },
+  QUOTE: {
+    BODY_1: 'You can also',
```
ttbarnes, line 126: Could we be more descriptive instead of BODY_1/2 etc? e.g `YOU_CAN`, `TO_CONTINUE`

Also don't think we should a content string with a comma

src/ui/templates/insurance/start.njk
```diff
@@ -51,4 +51,20 @@
 
   </form>
 
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-three-quarters-from-desktop">
+        <p class="govuk-body" data-cy="sign-in">
+          <span>{{ CONTENT_STRINGS.SIGN_IN.BODY_1 }}</span> <a class="govuk-link govuk-link--no-visited-state" href="{{ CONTENT_STRINGS.SIGN_IN.LINK.HREF }}" data-cy="sign-in-link">{{ CONTENT_STRINGS.SIGN_IN.LINK.TEXT }}</a> <span>{{ CONTENT_STRINGS.SIGN_IN.BODY_2 }}</span>
+        </p>
```
ttbarnes, line 58: Can we have each tag on its own line?

Zainzzkk: it fails the e2e test because it expects a \n then

ttbarnes: Up to you, in other places we data `data-cy` on each element inside a paragraph and test each element individually.

It does complicates the test; The trade off to not doing this is the nunjucks template is hard to follow.

e2e-tests/cypress/e2e/journeys/insurance/start.spec.js
```diff
@@ -69,6 +69,36 @@ context('Insurance Eligibility - start page', () => {
...
+
+    it('renders a "get a quote" body text', () => {
+      const { QUOTE } = CONTENT_STRINGS;
+
+      cy.checkText(insurance.startPage.getAQuote.youCan(), QUOTE.YOU_CAN);
+
+      cy.checkLink(insurance.startPage.getAQuote.link(), QUOTE.LINK.HREF, QUOTE.LINK.TEXT);
+
+      cy.checkText(insurance.startPage.getAQuote.ifYouNeed(), QUOTE.IF_YOU_NEED);
+    });
```
ttbarnes, line 101: Not a blocker, could destructure the sections from `insurance.startPage`

ttbarnes (approved):

e2e-tests/cypress/e2e/journeys/insurance/start.spec.js
Zainzzkk: will do in next PR

PR #535, opened 2023-06-09
fix(EMS-1720): Quote - Tell us about your policy - Header

Zainzzkk:
# Issue:
* Authenticated header on Tell us about your policy quote page was missing my applications and sign out links

# Changes made:
* Added HEADER page content to pageVariables for tell-us-about-your-policy
* Updated and added e2e test
* Updated insurance/start test to dereference nested page test

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/quote/tell-us-about-your-policy/tell-us-about-your-multi-policy-signed-in.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/quote/tell-us-about-your-policy/tell-us-about-your-multi-policy-signed-in.spec.js
@@ -15,7 +15,7 @@ context('Tell us about your multiple policy page - Signed in', () => {
 
   before(() => {
     cy.completeSignInAndGoToApplication();
-    cy.login()
+    cy.login();
 
     completeAndSubmitBuyerCountryForm();
     completeAndSubmitBuyerBodyForm();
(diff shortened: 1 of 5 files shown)


e2e-tests/cypress/e2e/journeys/quote/tell-us-about-your-policy/tell-us-about-your-multi-policy-signed-in.spec.js
```diff
@@ -0,0 +1,48 @@
...
+      pageTitle: CONTENT_STRINGS.MULTIPLE_POLICY_PAGE_TITLE,
+      currentHref: ROUTES.QUOTE.TELL_US_ABOUT_YOUR_POLICY,
+      backLink: ROUTES.QUOTE.POLICY_TYPE,
+      assertAuthenticatedHeader: true,
+      isInsurancePage: false,
+      lightHouseThresholds: {
+        // accessibility threshold is reduced here because
+        // the radio component from design system has an invalid aria attribute.
+        // this is out of our control
+        accessibility: 92,
```
ttbarnes, line 44: Maybe can remove this since the lighthouse checks are disabled as they stopped working with cypress update.

ttbarnes (approved):

e2e-tests/cypress/e2e/journeys/quote/tell-us-about-your-policy/tell-us-about-your-multi-policy-signed-in.spec.js
ttbarnes: Actually I think we should keep it in sorry, just incase we can re-enable it. There is a ticket in the backlog to try and re-enable them when (hopefully) a new package/version is relesed. What do you think?

Zainzzkk: lets keep it in then :P

PR #536, opened 2023-06-09
fix(EMS-1729): Account - Sign in - Password as numbers only

Zainzzkk:
# Issue: 
* When password is a number only, it caused service to go to problem with service page

# Changes made:
* Cast password as a string so graphql does not receive only a number
* Added tests

### DIFF
--- a/src/ui/server/controllers/insurance/account/sign-in/index.test.ts
+++ b/src/ui/server/controllers/insurance/account/sign-in/index.test.ts
@@ -287,6 +287,20 @@ describe('controllers/insurance/account/sign-in', () => {
           expect(res.redirect).toHaveBeenCalledWith(`${SUSPENDED_ROOT}?id=${accountSignInResponse.accountId}`);
         });
       });
+
+      describe('when password only contains numbers', () => {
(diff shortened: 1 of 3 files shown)


src/ui/server/controllers/insurance/account/sign-in/index.test.ts
```diff
@@ -287,6 +287,20 @@ describe('controllers/insurance/account/sign-in', () => {
           expect(res.redirect).toHaveBeenCalledWith(`${SUSPENDED_ROOT}?id=${accountSignInResponse.accountId}`);
         });
       });
+
+      describe('when password only contains numbers', () => {
+        it('should call api.keystone.account.signIn with password as a string', async () => {
+          req.body[PASSWORD] = 12345;
```
ttbarnes, line 293: This would come through as a string from the HTML form

ttbarnes: If this is updated to be a string like it would be outside of tests, this test will fail

ttbarnes: I will address this and maybe able to simplify the E2E test in my next PR 👍

PR #538, opened 2023-06-12
fix(EMS-1735): Account - Password reset - link expired - ability to send a new link

ttbarnes:
This PR adds the ability to send a new link from the "password reset - link expired" page. previously, the "send new link" button was a dummy link that didn't do anything.

## Changes

- Update API `verifyAccountPasswordResetToken` GQL resolver to return an account ID.
- Create "password reset - link expired" UI POST controller to:
  - Consume an account ID via query params.
  - Get an account's email address from the ID.
  - Call the API to email a new pasword reset link.
  - If there is no account ID in the query params, redirect to root "password reset" page/URL.
- Update "new password" UI controller to redirect to the "link epired" page with an account ID query param, if the token verification response returns `expired=true`.
- Add E2E test coverage.
- Update cypress `updateAccount` API call to return `passwordResetHash` 

## Other improvements

- Fixed some typos in E2E tests.
- Use `baseUrl` constant and `assertUrl` command in some E2E tests.
- Fixed a typo in a UI unit test for account sign in (https://github.com/UK-Export-Finance/exip/pull/536#discussion_r1226752035)
- Simplified E2E tests for account sign in by moving "password as pure number" unit test, into the sign in validation E2E test file (https://github.com/UK-Export-Finance/exip/pull/536#discussion_r1226752035)


src/ui/server/controllers/insurance/account/password-reset/link-expired/index.ts
```diff
@@ -21,3 +29,40 @@ export const get = (req: Request, res: Response) =>
...
+
+    const urlOrigin = req.headers.origin;
+
+    const accountId = req.query.id;
+
+    if (!accountId) {
+      return res.redirect(PASSWORD_RESET_ROOT);
+    }
+
+    const account = await api.keystone.account.get(accountId);
```
Zainzzkk, line 52: do we need a check for if account does not exist? or comes back as undefined?

ttbarnes: In the previous lines, if there is no `accountId` it will redirect to `PASSWORD_RESET_ROOT`

src/ui/templates/insurance/account/password-reset/link-expired.njk
```diff
@@ -15,7 +16,20 @@
 
       <p class="govuk-body" data-cy="if-you-would-like">{{ CONTENT_STRINGS.IF_YOU_WOULD_LIKE }}</p>
 
-      <a class="govuk-button" href="#" data-cy="send-new-link-button">{{ CONTENT_STRINGS.BUTTONS.SEND_NEW_LINK }}</a>
+      <form method="POST" novalidate>
+
```
Zainzzkk, line 20: not a blocker.  Could remove empty some new lines

Zainzzkk (dismissed):

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #540, opened 2023-06-13
chore(EMS-1721): Move API GQL resolvers into their own directories

ttbarnes:
This PR improves the structure of the GQL resolvers in the API by moving each resolver function and unit test into their own directory.

## Changes

- Move resolvers into directories.
- Update import paths.


src/api/custom-resolvers/mutations/index.ts
```diff
@@ -10,7 +10,7 @@ import sendEmailPasswordResetLink from './send-email-password-reset-link';
 import accountPasswordReset from './account-password-reset';
 import sendEmailReactivateAccountLink from './send-email-reactivate-account-link';
 import deleteApplicationByReferenceNumber from './delete-application-by-refrence-number';
-import updateCompanyAndCompanyAddress from './update-company-and-company-address';
+import updateCompanyAndCompanyAddress from './update-company-and-company-address/inde';
```
Zainzzkk, line 13: is this right?

ttbarnes: Nope, good spot - fixed

Zainzzkk (approved):

PR #548, opened 2023-06-16
fix(EMS-1345): Add requestedApplicationCreation to session when creating an application

ttbarnes:
This PR should fix an issue where if a user spams particular submit buttons/form POSTs, multiple applications requests would be made.

## Changes
- In each POST controller that requests application creation from the API, add a `requestedApplicationCreation` flag to the session and after the call has been made, reverse the flag to be false.
- Update `canCreateAnApplication` helper so that if the session has a `requestedApplicationCreation` flag, the helper will return false.
- Documentation improvements.
- Note: unfortunately, it doesn't seem possible to write an E2E test for this scenario (spamming a submit button during a POST).

### DIFF
--- a/src/ui/server/controllers/insurance/account/create/your-details/index.test.ts
+++ b/src/ui/server/controllers/insurance/account/create/your-details/index.test.ts
@@ -192,6 +192,12 @@ describe('controllers/insurance/account/create/your-details', () => {
           expect(createApplicationSpy).toHaveBeenCalledWith(eligibilityAnswers, mockSaveDataResponse.id);
         });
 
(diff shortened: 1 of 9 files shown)


Zainzzkk (commented):
I was wondering if it's worth removing the flag once it redirects to dashboard? So there is no chance to create multiple applications in that case

ttbarnes:
> I was wondering if it's worth removing the flag once it redirects to dashboard? So there is no chance to create multiple applications in that case

I'm not following, all the controllers that actually create an application have the flag added and removed now, before redirecting to another page/route. The dashboard GET doesn't have anything to do with creating an application?

Zainzzkk (approved):

PR #557, opened 2023-06-22
fix(EMS-1345): Disable submit buttons when a form is submitting

ttbarnes:
This PR adds a new client side JS script to disable submit buttons when a form is submitted. This prevents form submission spamming and an issue where if a submit handler creates an application, multiple applications can be created.

## Changes
- Create new `form-submission` script to listen for form submission event and then disable any elements with 'submit-button' ID.
- Remove unrequired, previously added `requestedApplicationCreation` flag set during application creation - this doesn't work as intended.
- Align all application creation logic in UI controllers.
- Add `no-store` to UI security middleware. Without this, an issue can occur with client side JS scripts in google chrome where navigating back to previous page would render the page with cached client side JS. FYI @abhi-markan
- Add an ID to all forms.
- Create a `submit-button` nunjucks component.
- Update all submit buttons to use the new nunjucks component.


src/ui/scripts/form-submission.js
```diff
@@ -0,0 +1,23 @@
+var form = document.getElementById('form');
```
Zainzzkk, line 1: is there a reason for var instead of const?

Zainzzkk (commented):
are there any tests we can do for this?

ttbarnes:
> are there any tests we can do for this?

Had a look and unfortunately there doesn't seem to be a way to add E2E tests for this, as when a form is submitted, it almost instantly redirects to another page. In a test, even adding a `cy.wait()` after clicking the submit button, does not pick up the disabled button.

Whilst this is not ideal I think it's OK as this is purely a visual enhancement and no functional logic is involved.

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #563, opened 2023-06-23
chore(EMS-1728): Companies House API restructure and test

Zainzzkk:
# Changes made:
* Added companies house API call to integrations
* Restructured API call file to not return axios response directly
* Added test for companies house
* renamed company-details-api-error test file


src/api/integrations/companies-house/index.ts
```diff
@@ -0,0 +1,48 @@
...
+const companiesHouseURL: any = process.env.COMPANIES_HOUSE_API_URL;
+
+/**
+ * getCompanyHouseDetails
+ * makes companies house API call and returns company
+ * @param {String} companyNumber
+ * @returns {CompaniesHouseAPIResponse} CompanyHouseResponse object
+ */
+const companiesHouse = {
+  getCompanyHouseDetails: async (companyNumber: string): Promise<CompaniesHouseAPIResponse> => {
```
ttbarnes, line 17: `companyHouse` should always be "companies" to match the name of the API

I think this could just be called `get` though, what do you think? So consumption would be like this:

```js
const company = await companiesHouse.get('123456');
```

Zainzzkk: done

src/api/custom-resolvers/queries/get-companies-house-information/index.ts
```diff
@@ -21,29 +15,21 @@ const getCompaniesHouseInformation = async (root: any, variables: GetCompaniesHo
...
-    if (!response.data || response.status === 404) {
+    if (!response.success || !response.data) {
       return {
         success: false,
       };
     }
 
     // gets all industry sectors
-    const industrySectorNames = await getIndustrySectorNames();
+    const industrySectorNames = await getIndustrySectorNames.get();
```
ttbarnes, line 32: Not a blocker, I/we can update this in anotehr but this can be called `industrySectorNames.get();`

ttbarnes (approved):

PR #567, opened 2023-06-23
chore(EMS-1807-1808): Your Business - Construct payload

Zainzzkk:
# Changes made:
* Added `constructPayload` helper to make payload from provided fieldIds
* Used constructPayload in your-business section to ensure only relevant fields are provided
* Added and updated unit tests

### DIFF
--- a/src/ui/server/controllers/insurance/business/broker/index.test.ts
+++ b/src/ui/server/controllers/insurance/business/broker/index.test.ts
@@ -1,7 +1,7 @@
 import { PAGES } from '../../../../content-strings';
-import { pageVariables, get, post, TEMPLATE, BROKER_FIELD_IDS } from '.';
+import { pageVariables, get, post, TEMPLATE, FIELD_IDS } from '.';
 import { TEMPLATES, ROUTES } from '../../../../constants';
-import FIELD_IDS from '../../../../constants/field-ids/insurance/business';
+import BUSINESS_FIELD_IDS from '../../../../constants/field-ids/insurance/business';
 import insuranceCorePageVariables from '../../../../helpers/page-variables/core/insurance';
 import constructPayload from '../../../../helpers/construct-payload';
 import getUserNameFromSession from '../../../../helpers/get-user-name-from-session';
@@ -14,7 +14,7 @@ import { mockReq, mockRes, mockApplication, mockBroker } from '../../../../test-
 
 const { BROKER: BROKER_FIELDS } = FIELDS;
 
-const { USING_BROKER, HEADING, NAME, ADDRESS_LINE_1, ADDRESS_LINE_2, COUNTY, TOWN, POSTCODE, EMAIL, DETAILS } = FIELD_IDS.BROKER;
+const { USING_BROKER, HEADING, NAME, ADDRESS_LINE_1, ADDRESS_LINE_2, COUNTY, TOWN, POSTCODE, EMAIL, DETAILS } = BUSINESS_FIELD_IDS.BROKER;
 
 const { BROKER } = PAGES.INSURANCE.EXPORTER_BUSINESS;
 const { BROKER: BROKER_TEMPLATE } = TEMPLATES.INSURANCE.EXPORTER_BUSINESS;
@@ -49,9 +49,9 @@ describe('controllers/insurance/business/broker', () => {
     });
   });
 
-  describe('BROKER_FIELD_IDS', () => {
+  describe('FIELD_IDS', () => {
     it('should have the correct FIELD_IDS', () => {
-      expect(BROKER_FIELD_IDS).toEqual([USING_BROKER, NAME, ADDRESS_LINE_1, ADDRESS_LINE_2, TOWN, COUNTY, POSTCODE, EMAIL]);
+      expect(FIELD_IDS).toEqual([USING_BROKER, NAME, ADDRESS_LINE_1, ADDRESS_LINE_2, TOWN, COUNTY, POSTCODE, EMAIL]);
     });
   });
 
@@ -147,9 +147,7 @@ describe('controllers/insurance/business/broker', () => {
       it('should render template with validation errors', async () => {
         req.body = {};
 
-        const submittedValues = {
-          [USING_BROKER]: req.body[USING_BROKER],
-        };
+        const payload = constructPayload(req.body, FIELD_IDS);
 
         await post(req, res);
 
@@ -164,7 +162,7 @@ describe('controllers/insurance/business/broker', () => {
           ...pageVariables(mockApplication.referenceNumber),
           validationErrors,
           application: mapApplicationToFormFields(mockApplication),
-          submittedValues,
+          submittedValues: payload,
         });
       });
     });
@@ -187,11 +185,11 @@ describe('controllers/insurance/business/broker', () => {
 
         await post(req, res);
 
-        const payload = constructPayload(req.body, BROKER_FIELD_IDS);
+        const payload = constructPayload(req.body, FIELD_IDS);
 
         expect(mapAndSave.broker).toHaveBeenCalledTimes(1);
 
-        expect(mapAndSave.broker).toHaveBeenCalledWith(payload, mockApplication);
+        expect(mapAndSave.broker).toHaveBeenCalledWith(req.body, mockApplication);
       });
 
       describe("when the url's last substring is `check-and-change`", () => {
--- a/src/ui/server/controllers/insurance/business/broker/index.test.ts
+++ b/src/ui/server/controllers/insurance/business/broker/index.test.ts
@@ -1,5 +1,5 @@
 import { PAGES } from '../../../../content-strings';
-import { pageVariables, get, post, TEMPLATE } from '.';
+import { pageVariables, get, post, TEMPLATE, BROKER_FIELDS_IDS } from '.';
 import { TEMPLATES, ROUTES } from '../../../../constants';
 import FIELD_IDS from '../../../../constants/field-ids/insurance/business';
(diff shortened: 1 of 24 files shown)


src/ui/server/controllers/insurance/business/broker/index.test.ts
```diff
@@ -195,6 +198,23 @@ describe('controllers/insurance/business/broker', () => {
           expect(res.redirect).toHaveBeenCalledWith(expected);
         });
       });
+
+      describe('when an extra field is inserted onto the page', () => {
+        it('should call mapAndSave.broker once without the extra field', async () => {
```
ttbarnes, line 203: I don't think we need this as `constructPayload` will test the removal of the field, for each controller we can just update the existing call so e.g

```js
it('should call mapAndSave.broker once with data from constructPayload function', async () => {
```

Makes it simpler, what do you think?

src/ui/server/controllers/insurance/business/broker/index.ts
```diff
@@ -9,6 +9,7 @@ import generateValidationErrors from './validation';
 import mapAndSave from '../map-and-save/broker';
 import isCheckAndChangeRoute from '../../../../helpers/is-check-and-change-route';
 import { Request, Response } from '../../../../../types';
+import constructPayload from '../../../../helpers/construct-payload';
```
ttbarnes, line 12: Can we have the import in the order it's executed? Recently tidied up a lot of imports so that `types` is the last thing imported

src/ui/server/controllers/insurance/business/broker/save-and-back/index.test.ts
```diff
@@ -50,6 +52,23 @@ describe('controllers/insurance/business/broker/save-and-back', () => {
...
+
+          await post(req, res);
+
+          const payload = constructPayload(req.body, BROKER_FIELDS_IDS);
+
+          expect(updateMapAndSave).toHaveBeenCalledTimes(1);
+
+          expect(updateMapAndSave).toHaveBeenCalledWith(payload, mockApplication, false);
+        });
+      });
```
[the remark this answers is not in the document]
ttbarnes: We could even potentially wrap `constructPayload` into a modular/object sturcture so in the controllers and map/save functions we can test that the function was called e.g

```
it('should call construct.payload with form data', ...
```

What do you think? Obviously with better naming.

src/ui/server/controllers/insurance/business/contact/index.ts
```diff
@@ -22,6 +23,8 @@ const { CONTACT: CONTACT_TEMPLATE } = TEMPLATES.INSURANCE.EXPORTER_BUSINESS;
 
 export const TEMPLATE = CONTACT_TEMPLATE;
 
+export const CONTACT_FIELDS_IDS = [FIRST_NAME, LAST_NAME, EMAIL, POSITION];
```
ttbarnes, line 26: typo, `FIELD_IDS` - love this though, very nice. Can we add a unit test for this as well?

src/ui/server/helpers/construct-payload/index.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Array<String>} fieldIds
+ * @returns {Object} constructed payload
+ */
+const constructPayload = (requestBody: RequestBody, fieldIds: Array<string>) => {
+  let payload = {};
+
+  // if either requestBody or fieldsIds are empty, return empty object
+  if (!objectHasKeysAndValues(requestBody) || !isPopulatedArray(fieldIds)) {
+    return payload;
+  }
```
ttbarnes, line 19: It looks like this condition might not be required  because they are required params. At least for the `fieldIds`? I maybe wrong though.

src/ui/server/helpers/construct-payload/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * loops through fieldsIds array
+   * if requestBody contains the fieldId: add it to payload
+   * else will add an empty string
+   */
+  fieldIds.forEach((fieldId) => {
+    payload = {
+      ...payload,
+      [fieldId]: requestBody[fieldId] ?? '',
+    };
```
ttbarnes, line 30: Could save some lines here 

```js
if (objectHasProperty(requestBody, fieldId)) {
  payload[fieldId] = requestBody[fieldId];
}
```

src/ui/server/helpers/construct-payload/index.ts
Zainzzkk: I think you can pass an empty array too

src/ui/server/controllers/insurance/business/contact/index.ts
Zainzzkk: FIELD_IDS is already imported at the top

ttbarnes: I think we should rename the top import to be `BUSINESS_FIELD_IDS` and then we can have a generic `FIELD_IDS` in all controllers.

Otherwise all controllers will have different naming conventions for their field IDs,

src/ui/server/helpers/construct-payload/index.ts
ttbarnes: Am I missing something, why would a payload from a form be an array?

Zainzzkk: fixed

src/ui/server/controllers/insurance/business/broker/index.test.ts
Zainzzkk: done

src/ui/server/controllers/insurance/business/broker/save-and-back/index.test.ts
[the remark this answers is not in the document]
Zainzzkk: I don't think spying on construct payload would achieve much because we already have a unit test for it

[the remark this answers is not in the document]
ttbarnes: Only concern is that if `constructPayload` is removed from the controller a test could still pass and we'd not know that `constructPayload` was used. Can you try removing it and see if the test passes?

src/ui/server/controllers/insurance/business/broker/index.test.ts
```diff
@@ -48,6 +49,12 @@ describe('controllers/insurance/business/broker', () => {
     });
   });
 
+  describe('FIELD_IDS', () => {
+    it('should have the correct FIELD_IDS', () => {
+      expect(FIELD_IDS).toEqual([USING_BROKER, NAME, ADDRESS_LINE_1, ADDRESS_LINE_2, TOWN, COUNTY, POSTCODE, EMAIL]);
```
ttbarnes, line 54: Not a blocker, can be done in the next PR, we can use `expected` const here so it's easier to read

src/ui/server/controllers/insurance/business/broker/save-and-back/index.test.ts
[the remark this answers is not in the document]
ttbarnes: Caught up - if the function is removed the test will fail so all good.

src/ui/server/controllers/insurance/business/company-details/index.test.ts
```diff
@@ -56,6 +55,12 @@ describe('controllers/insurance/business/companies-details', () => {
     });
   });
 
+  describe('FIELD_IDS', () => {
+    it('should have the correct FIELD_IDS', () => {
+      expect(FIELD_IDS).toEqual([COMPANY_HOUSE.INPUT, TRADING_NAME, TRADING_ADDRESS, WEBSITE, PHONE_NUMBER]);
```
ttbarnes, line 60: Same as above not a blocker

src/ui/server/controllers/insurance/business/company-details/index.test.ts
```diff
@@ -347,10 +346,24 @@ describe('controllers/insurance/business/companies-details', () => {
           userName: getUserNameFromSession(req.session.user),
           ...pageVariables(mockApplication.referenceNumber, COMPANY_DETAILS_ROUTE),
           SUMMARY_LIST: companyHouseSummaryList(mockCompany),
-          submittedValues,
+          submittedValues: payload,
           validationErrors: {},
         });
       });
+
+      describe('when an extra field is inserted onto the page', () => {
```
ttbarnes, line 354: We don't need to test the extra field/value here, that's the responsibility of `constructPayload`, just need to test that the function is callled

Can be simplified next PR so we unblock things

src/ui/server/controllers/insurance/business/company-details/save-and-back/index.POST-save-and-back.test.ts
```diff
@@ -61,28 +65,39 @@ describe('controllers/insurance/business/companies-details', () => {
...
       });
 
-      it('should call mapAndSave.companyDetails once', async () => {
+      it('should call mapAndSave.companyDetails once with the data from constructPayload function and company', async () => {
         req.body = {
-          [TRADING_NAME]: 'true',
-          [TRADING_ADDRESS]: 'false',
-          [PHONE_NUMBER]: VALID_PHONE_NUMBERS.LANDLINE,
+          ...body,
+          injection: 1,
```
ttbarnes, line 86: Same as above, there's no value in adding extra fields here

src/ui/server/helpers/construct-payload/index.ts
ttbarnes: Nice one - comment is invalid now though it mentions `requestBody` but it only cares about the array

ttbarnes (approved):
Happy to merge can address minor comments in next PR.

PR #580, opened 2023-06-26
chore(EMS-1816): Fix various issues reported by SonarCloud

ttbarnes:
This PR fixes some security and code smell issues reported by SonarCloud on the `main-application` branch.

## Changes
- Wrap some "expected message" test assertions (that use content strings) in a `String()`.
- Simplify various conditions and destructurings.
- Remove some commented code.
- Remove some unnecessary type declarations.
- Add "https" to some mock website strings.
- Remove a `throw new Error()` from an E2E API call as this was throwing false errors (where a call to delete account does not return successful because the account does not exist.
- Change all instances of "google.com" to "gov.uk"

### DIFF
--- a/e2e-tests/cypress/e2e/journeys/insurance/account/create/your-details/account-create-your-details-validation-account-already-exists.spec.js
+++ b/e2e-tests/cypress/e2e/journeys/insurance/account/create/your-details/account-create-your-details-validation-account-already-exists.spec.js
@@ -59,7 +59,7 @@ context('Insurance - Account - Create - Your details page - form validation - As
     const value = mockAccount.email;
     const fieldIndex = 0;
     const TOTAL_REQUIRED_FIELDS = 1;
-    const expectedMessage = YOUR_DETAILS_ERROR_MESSAGES[EMAIL].ACCOUNT_ALREADY_EXISTS;
+    const expectedMessage = String(YOUR_DETAILS_ERROR_MESSAGES[EMAIL].ACCOUNT_ALREADY_EXISTS);
 
     cy.submitAndAssertFieldErrors(field, value, fieldIndex, TOTAL_REQUIRED_FIELDS, expectedMessage);
   });
--- a/src/ui/server/helpers/is-string-with-http/index.test.ts
+++ b/src/ui/server/helpers/is-string-with-http/index.test.ts
@@ -9,7 +9,7 @@ describe('helpers/isStringWithHttp', () => {
   });
 
   it('should not add "http://" to beginning of string incoming string already contains "http://"', () => {
-    const website = 'http://www.google.com';
+    const website = 'https://www.google.com';
 
     const result = isStringWithHttp(website);
     expect(result).toEqual(website);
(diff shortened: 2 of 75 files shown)


e2e-tests/cypress/e2e/journeys/insurance/account/create/your-details/account-create-your-details-validation-account-already-exists.spec.js
```diff
@@ -59,7 +59,7 @@ context('Insurance - Account - Create - Your details page - form validation - As
     const value = mockAccount.email;
     const fieldIndex = 0;
     const TOTAL_REQUIRED_FIELDS = 1;
-    const expectedMessage = YOUR_DETAILS_ERROR_MESSAGES[EMAIL].ACCOUNT_ALREADY_EXISTS;
+    const expectedMessage = String(YOUR_DETAILS_ERROR_MESSAGES[EMAIL].ACCOUNT_ALREADY_EXISTS);
```
Zainzzkk, line 62: but this error message is already a string?

ttbarnes: SonarCloud does not think so because it's a dynamic object structure. e.g here, if the string that is `EMAIL`,  is not found in `YOUR_DETAILS_ERROR_MESSAGES`, it'll return `[Object Object]. This is what it flags so wrapping into a string fixes this concern.

Zainzzkk: I guess it is counter productive in a test since that Error message will always exist for the test.  Would make sense in a non-test file

Zainzzkk (approved):

e2e-tests/cypress/e2e/journeys/insurance/account/create/your-details/account-create-your-details-validation-account-already-exists.spec.js
ttbarnes: I do agree it feels strange - regardless of if it's in a test or not; but it is a genuine concern, e.g if the `EMAIL` constant is changed to have a value of eg `emailAddress`, this message would not return a string.

src/ui/server/helpers/is-string-with-http/index.test.ts
```diff
@@ -9,7 +9,7 @@ describe('helpers/isStringWithHttp', () => {
   });
 
   it('should not add "http://" to beginning of string incoming string already contains "http://"', () => {
-    const website = 'http://www.google.com';
+    const website = 'https://www.google.com';
```
abhi-markan, line 12: Can we please change all the reference to `google.com` to `gov.uk` please.

abhi-markan (changes requested):
All good, just one minor comment to address please.

abhi-markan (approved):

PR #581, opened 2023-06-26
security(EMS-1757): Sanitise referer header

ttbarnes:
This PR updates the UI's security middleware to sanitise the "referer" and "origin" request headers, matching the approach in a [DTFS PR](https://github.com/UK-Export-Finance/dtfs2/pull/1780).

## Changes
* Add `is-url` package to check that the said headers are valid URLs.
* Create new security middleware function `getUrlOrigin` to split a URL and return the origin.
* Create new security middleware function `isRequestHeaderOriginValid`, to check if a provided header URL matches the true origin.
* Update headers security middleware to check the said headers via `isRequestHeaderOriginValid` and if they are invalid, change the headers to an empty string.
- Add missing test coverage for security middleware.


Zainzzkk (approved):
Would like @abhi-markan to have a look before this is merged

abhi-markan (approved):

PR #688, opened 2023-07-21
feat(EMS-1787-1802-1856): API security - Add x-api-key checks and rate limiter

ttbarnes:
This PR updates the API so that any calls to the API without a valid `x-api-key` will be rejected. Also added rate limiter config.

## Changes
- Add `API_KEY` environment variable.
- Extend keystone config via `extendExpressApp` and call new middleware functions.
- Create new middlewares: `checkApiKey` and `rateLimiter`.
- Update the UI's GQL config to pass `x-api-key` header.
- Update E2E API calls to pass `x-api-key` header.
- Tidy some documentation.
- Fixed some failing API and UI date related tests due to an invalid dates being generated/asserted on the last day of a month.
- Bump dependencies.


src/api/.keystone/config.js
```diff
@@ -33,9 +33,33 @@ __export(keystone_exports, {
...
+var import_config = require("dotenv/config");
+var { API_KEY } = process.env;
+var checkApiKey = (req, res, next) => {
+  const { "x-api-key": xApiKey } = req.headers;
+  if (xApiKey !== API_KEY) {
+    return res.status(401).send("Unauthorised");
+  }
+  next();
+};
+var check_api_key_default = checkApiKey;
```
ChristianMcc, line 49: what's the point of this 'default' var? Is there going to be other overridden functions?

src/api/.keystone/config.js
```diff
@@ -33,9 +33,33 @@ __export(keystone_exports, {
...
+
+// middleware/rate-limiter/index.js
+var import_express_rate_limit = __toESM(require("express-rate-limit"));
```
ChristianMcc, line 52: out of curiosity as I'm not familiar with the package how does this work? Is it 1000 requests a minute per express user? or across the whole API? (And does this mean that it's per server hosted in azure that we have this rate limit?)

src/api/.keystone/config.js
```diff
@@ -33,9 +33,33 @@ __export(keystone_exports, {
...
 });
 module.exports = __toCommonJS(keystone_exports);
-var import_config2 = require("dotenv/config");
+var import_config3 = require("dotenv/config");
 var import_core2 = require("@keystone-6/core");
 
+// middleware/headers/check-api-key/index.ts
+var import_config = require("dotenv/config");
+var { API_KEY } = process.env;
+var checkApiKey = (req, res, next) => {
```
ChristianMcc, line 42: I'm not familiar with keystone - why does this need tobe defined here and then also defined again i the check-api-key/index.ts file?

src/ui/server/graphql/apollo.ts
```diff
@@ -38,6 +38,11 @@ const doQuery = async (query: DocumentNode, variables: OperationVariables) => {
     return await client.query({
       query,
       variables,
+      context: {
```
ChristianMcc, line 41: Just to confirm, the whole frontend server runs on a graphql API that wasn't send an API key before so everything is auto covered by attaching it to every query/mutation?

ChristianMcc (commented):
Looks good to me! A few questions as I'm not very familiar with the codebase and for my own understanding 😄

src/api/.keystone/config.js
ttbarnes: That's is, this is across the whole API so after 1000 requests in a minute the rate limiter will reject calls. We have the exact same in place for the UI and probably in some DTFS services.

AFAIK there's nothing specific for rate limiting in Azure setup as this handles it

src/api/.keystone/config.js
ttbarnes: Same as above, this is automatically compiled by KeystoneJS.

We have to include this in the repo unfortunately, we tried generating it behind the scenes with docker/GHA (to avoid having to commit this file), but we couldn't get it to work.

src/ui/server/graphql/apollo.ts
ttbarnes: Yes exactly 😄 All calls to our API from the UI are made via GraphQL, apollo client. So every call will go through this file or more specifically here: https://github.com/UK-Export-Finance/exip/blob/c52d1284cbba3ca3b3a7b879ede69ff5fd02473c/src/ui/server/graphql/apollo.ts#L69

ChristianMcc (approved):
LGTM :) would expect rate limiting to be stripped out in future once we have the same azure API management solution on this project but nice to have this for now!

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #707, opened 2023-07-27
chore(EMS-1804): Account - construct payload, sanitise req query

ttbarnes:
This PR updates all form submissions in the "Account" section so that when a form is submitted:

- Only specified fields are consumed
- Any other provided fields are ignored.
- Request query/params are sanitised before being consumed.

## Changes

- Added `FIELD_ID` or `FIELD_IDS` exports and test for relevant controllers.
- Updated relevant controllers so that:
  - All `req.body` consumptions use `constructPayload`.
  - All `req.query` consumptions se `sanitiseValue`.
- Updated `sanitiseValue` and `shouldChangeToNumber` sanitise-data functions to accept an object param so that they can be consumed with an optional key, therefore simplifying new usages of `sanitiseValue`, where a key is irrelevant.
- Fixed some typos and incorrectly structured unit test orderings.
- Simplified some unit tests.


src/ui/server/controllers/insurance/account/create/confirm-email/index.test.ts
```diff
@@ -64,9 +65,11 @@ describe('controllers/insurance/account/create/confirm-email', () => {
 
         await get(req, res);
 
+        const sanitisedId = String(sanitiseValue({ value: req.query.id }));
```
Zainzzkk, line 68: same here

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #750, opened 2023-08-03
chore(EMS-1893): Reject any GraphQL calls/resolvers that are not in an allowed list

ttbarnes:
This PR updates the API so that there is an explicit list of GQL resolvers that are allowed to be used. If a request is made and the requested resolver is not in the allowed list, the call will be rejected.

This is achieved by using `requestDidStart` and  `didResolveOperation` in apollo plugins. This operation fires after GraphQL determines which operation to execute. At this stage, the resolvers have not executed yet, but the operation/resolver name is available. We can therefore check the name of the resolver and reject accordingly.

## Changes
- Create `apollo-plugins` with new plugin to check the operation name as per PR description.
- Create `ALLOWED_GRAPHQL_RESOLVERS` constants.
- Fix an ordering issue with API key middleware.
- Add new `apollo-plugins` directory to Dockerfile; update ordering to match the actual directory.
- Update UI GQL queries so that the namings are lowercase.


Zainzzkk (approved):
lgtm

abhi-markan (approved):
LGTM 👍🏼

PR #751, opened 2023-08-03
chore(deps): dependencies updates

abhi-markan:
## Introduction ✏️ 
Dependency updates

## Resolution ✔️ 
* Removed few GitHub notified vulnerabilities.
* Updated dependencies.

## Miscellaneous ➕ 
* Updated `.dockerignore`
* Updated `.gitignore`
* Updated GHA : Dependabot lint and test pipeline execution.


Zainzzkk (dismissed):

src/ui/.dockerignore
```diff
@@ -1,10 +1,53 @@
...
+**/generated_reports
+**/temp/*
+**/actionsheets/*
+**/json/*
+**/amendments/*
+**/exports/*
+**/bss-ewcs/dump/*
+**/bss-ewcs/data/*
+**/gef/dump/*
+**/gef/data/*
```
ttbarnes, line 51: BSS EWCS, GEF?

abhi-markan: Have removed.

src/ui/server/helpers/task-list/index.ts
```diff
@@ -42,7 +42,7 @@ export const generateSimplifiedTaskList = (taskList: TaskListData): Array<TaskLi
           status: task.status,
           title: task.title,
         })),
-      } as TaskListGroup),
+      }) as TaskListGroup,
```
Zainzzkk, line 45: my linting normally throws an error for this if its not all in the bracket

abhi-markan: Not sure why as we all should have same linting. Are you using VS code linting on top of prettier?

ttbarnes: I think this is now enforced with new version of prettier so without an `npm install` the linting would be different?

Zainzzkk (approved):

PR #752, opened 2023-08-04
chore(EMS-1891): Merge main into main-application

ttbarnes:
This PR updates the `main-application` branch with the latest `main` branch, containing updates to the quote tool/pricing grid.

## Changes
 - Merge `main` into `main-application`.
 - Update Dockerfile node versions so they're aligned with `main`.


e2e-tests/cypress/e2e/journeys/quote/your-quote/your-quote-multi-policy.spec.js
```diff
@@ -109,7 +109,7 @@ context('Get a quote/your quote page (multiple policy) - as an exporter, I want
 
         cy.checkText(row.key(), expectedKeyText);
 
-        const expected = '1.15%';
+        const expected = '1.13%';
```
Zainzzkk, line 112: should these not come from constants?

ttbarnes: Do you have something in mind? There's a few quote tool tests that assert specific percentages and prices. This is already in `main` branch/production

Zainzzkk: for later but we could have a const that does the calculation for us so we don't need to keep updating it manually

Zainzzkk (approved):

e2e-tests/cypress/e2e/journeys/quote/your-quote/your-quote-multi-policy.spec.js
ttbarnes: That'd be good, we already have a calculation function (`generate-quote` directory). Without having a new shared package for helpers, constants etc, it could only be achieved by copying that directory and the pricing grid JSON into the E2E directory.

Zainzzkk (approved):

PR #757, opened 2023-08-07
fix(EMS-1890): add temporary logs to debug in dev environment

ttbarnes:
There's currently an issue in the dev environment where some UI request headers do not match or could be undefined. 

We have some middleware that checks these headers and if they don't match, we nullify the headers. One of these headers is used to render a back link in each page which is currently not working.

Unable to replicate the issue locally so need to add some logs so that we can debug in the dev envrionment.


Zainzzkk (approved):
impressive work

PR #809, opened 2023-08-14
chore(EMS-848): Change yes/no string fields to booleans

ttbarnes:
This PR changes various fields defined by the KeystoneJS schema to be a boolean/"checkbox" field instead of a text field with "yes/no" values.

## Changes
- Create a custom KeystoneJS field, `nullableCheckbox`. This allows us to create checkbox keystone/graphQL/DB fields that are null by default.
  - This is not possible with Keystone's default checkbox field - it has to have a true or false value by default, which is not appropriate for our use case.
- Update all type definitions, conditions, tests and mock data for the following fields:
  - `hasTradingAddress`
  - `hasTradingName`
  - `hasAntiBriberyCodeOfConduct`
  - `exporterHasTradedWithBuyer`
  - `isUsingBroker`
  - `canContactBuyer`
  - `exporterIsConnectedWithBuyer`
  - `exporterHasTradedWithBuyer`
  - Section review: `eligibility`, `policyAndExport`, `business`, `buyer`
- Remove now unnecessary logic for "yes/no" strings in the `yes-no-radio-buttons` nunjucks component.
- Update and rename `getYesAndTrueAnswerProperties` to `getTrueAndFalseAnswers`,
- Update "required declarations fields" to handle a particular declaration answer being null, undefined, true or false.
- Update "your broker" and "your buyer" summary list mappings so that "yes/no" values will appear instead of booleans.
- Bump KeystoneJS version.


Zainzzkk (commented):
Is there anything in submittedValues, where if the value is null, it does not default to false which selects the no radio?

ttbarnes:
> Is there anything in submittedValues, where if the value is null, it does not default to false which selects the no radio?

The radio's will only be selected it the value is either true or false due to updated nunjucks conditions for `checked` i.e:

```js
checked: submittedAnswer === true
checked: submittedAnswer === false
```

Zainzzkk (approved):

PR #829, opened 2023-08-17
fix(EMS-1300): Quote - tell us about your policy heading, improve test coverage

ttbarnes:
This PR adds a missing content string to the "quote - tell us about your policy" page/controller, so that the error summary will display a "There is a problem" heading. This was caused by a recent tech improvement.

## Changes
- Update "tell us about your policy" controller to include `ERROR_MESSAGES` content strings.
- Create a new cypress command to check an error summary list's heading, `cy.checkErrorSummaryListHeading`.
- Update `submitAndAssertFieldErrors` and `submitAndAssertRadioErrors` cypress commands to check the summary list heading.
- Update all other E2E tests that do not use `submitAndAssertFieldErrors`, to call `cy.checkErrorSummaryListHeading();

### DIFF
--- a/e2e-tests/commands/check-buyer-country-form.js
+++ b/e2e-tests/commands/check-buyer-country-form.js
@@ -84,7 +84,8 @@ const checkSubmitButton = () => {
 };
 
 const checkValidationErrors = () => {
-  partials.errorSummaryListItems().should('exist');
+  cy.checkErrorSummaryListHeading();
+
   partials.errorSummaryListItems().should('have.length', 1);
 
   const expectedMessage = String(ERROR_MESSAGES.ELIGIBILITY[FIELD_IDS.ELIGIBILITY.BUYER_COUNTRY]);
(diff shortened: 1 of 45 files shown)


e2e-tests/commands/check-buyer-country-form.js
```diff
@@ -84,7 +84,8 @@ const checkSubmitButton = () => {
 };
 
 const checkValidationErrors = () => {
-  partials.errorSummaryListItems().should('exist');
+  cy.checkErrorSummaryListHeading();
+
   partials.errorSummaryListItems().should('have.length', 1);
```
Zainzzkk, line 89: what about the function we have to check errors including summary list and field errors all in one?

ttbarnes: Do you mean `submitAndAssertFieldErrors`? I did think of this, it'd be a pretty epic job, there is a ticket in the backlog for this, EMS-969.

Zainzzkk (approved):

PR #830, opened 2023-08-17
feat(EMS-1919-1933): Quote - Policy length updates, simplify policy type/length IDs usage

ttbarnes:
This PR updates the quote tool so that the "policy length" (only required for a single policy contract), will be rendered in the "tell us about your policy" page instead of the "type of policy" page.

## Changes

- Minimise usages of `SINGLE_POLICY_LENGTH`, `MULTIPLE_POLICY_LENGTH`, `SINGLE_POLICY_LENGTH` and `MULTIPLE_POLICY_LENGTH` as much as possible; Use `POLICY_LENGTH` and `POLICY_TYPE` instead.
- Extract policy length field and validation from "type of policy " page to the "tell us about your policy" page.
- Update E2E tests/flows so that the policy length field is completed/submitted in the "tell us about your policy" page.
- Simplify policy type mapping in the UI helper function `mapAnswersToContent`.
- Add "policy type" field to UI helper function `mapQuoteToContent`.
- Update the "required data provided" UI middleware function so that `POLICY_LENGTH` is required as part of "tell us about your policy" instead of the "type of policy" page.


e2e-tests/quote/cypress/e2e/journeys/quote/tell-us-about-your-policy/tell-us-about-your-single-policy-validation.spec.js
```diff
@@ -85,20 +97,100 @@ context('Tell us about the policy you need page - form validation', () => {
...
   });
 
+  describe('when `policy length` has a non-numeric value', () => {
+    it('should render a validation error', () => {
+      cy.navigateToUrl(url);
+
+      cy.keyboardInput(tellUsAboutYourPolicyPage[POLICY_LENGTH].input(), 'a');
+      submitButton().click();
+
+      cy.checkText(
```
Zainzzkk, line 125: is it worth using the error cy command here? to make it 1 line?

src/ui/server/controllers/quote/policy-type/index.ts
```diff
@@ -68,7 +60,7 @@ export const post = (req: Request, res: Response) => {
   if (isSinglePolicyType(payload[POLICY_TYPE])) {
     populatedData = {
       [POLICY_TYPE]: payload[POLICY_TYPE],
-      [POLICY_LENGTH]: payload[SINGLE_POLICY_LENGTH],
+      // [POLICY_LENGTH]: payload[POLICY_LENGTH],
```
Zainzzkk, line 63: remove this comment?

e2e-tests/quote/cypress/e2e/journeys/quote/tell-us-about-your-policy/tell-us-about-your-single-policy-validation.spec.js
ttbarnes: Makes sense, but this is in a lot of other places and we have a ticket to use more cypress commands for this kind of thing. I think we  should tackle it there. In this PR, just moved these assertions from one place to another

Zainzzkk (approved):

PR #865, opened 2023-08-24
fix(EMS-974): Turnover - Remove financialYearEndDate if does not exist

Zainzzkk:
# Issue: 
* on resubmission of companies house number form for a company without a financialYearEndDate, the previous companies value would be kept

# Changes made:
* Added check if companyNumber and no financialYearEndDate, then to set it as null
* Added to e2e test


src/api/types.ts
```diff
@@ -440,6 +440,11 @@ interface TestHelperAccountCreate {
   deleteAccounts?: boolean;
 }
 
+interface TestHelperCreate {
+  context: Context;
+  companyId?: string;
+}
```
ttbarnes, line 446: Can we rename this type to be more descriptive as its not generic, not a blocker.

ttbarnes (approved):

PR #879, opened 2023-08-24
feat(EMS-1159): Dashboard pagination

ttbarnes:
This PR adds pagination functionality to the dashboard.

## Changes
- Add `createApplications` GQL mutation and mock generation for E2E tests. This allows us to create many applications with a single API call.
- Update various E2E commands to return `accountId`, we need this in order to create an application with an owner ID.
- Add various E2E commands to assert pagination elements - link items, "next", "previous" links, active link.
- Add E2E test coverage for checking the pagination depending on how many applications there are - it acts differently depending on the amount of applications.
- Add `createApplications` GQL resolver to the list of "allowed resolvers" when running in dev environment.
- Update the UI's API/GQL call to "get applications" to pass "take" and "skip" variables and return `totalApplications`. Now, only 15 applications are taken per page and we can skip to e.g 30 applications if a user is on the 3rd page.
- Create `generatePaginationItems` UI helper function to generate an array of pages/URLs. This is then passed to the nunjucks component.
- Create `getSkipCount` UI helper function to get the amount of applications we should skip, depending on the current page number.
- Create `getTotalPages` UI helper function to get the total amount of pages required for pagination.
- Create a new UI route for for dashboard pages - `/dashboard/page/:pageNumber`.
- Update the UI's `applicationAccess` middleware function to ignore the `DASHBOARD_PAGE` route, similar to the `DASHBOARD` route.
- Update the dashboard nunjucks template to include "(Page X of Y)" in the page title if pagination is in use.


e2e-tests/insurance/cypress/e2e/journeys/dashboard/dashboard-invalid-page-number-param.spec.js
```diff
@@ -0,0 +1,34 @@
+import dashboardPage from '../../../../../pages/insurance/dashboard';
+import { INSURANCE_ROUTES } from '../../../../../constants/routes/insurance';
+
+const { table } = dashboardPage;
+const { DASHBOARD_PAGE } = INSURANCE_ROUTES;
+
+context('Insurance - Dashboard - visit with 1 existing application and an invalid page nubmer param', () => {
```
Zainzzkk, line 7: typo :P

src/ui/server/api/keystone/applications/index.ts
```diff
@@ -1,13 +1,18 @@
 import { ApolloResponse } from '../../../../types';
 import apollo from '../../../graphql/apollo';
 import getApplicationsQuery from '../../../graphql/queries/applications';
+import { MAX_APPLICATIONS_PER_PAGE } from '../../../constants';
 
 const applications = {
-  getAll: async (accountId: string) => {
+  getAll: async (accountId: string, skip: number) => {
```
Zainzzkk, line 7: could we get a comment to explain what skip is for if anyone else needs to work on this in the future?

ttbarnes: Good shout, added some documentation 👍

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #894, opened 2023-08-29
chore(docs): Spectaql config

ttbarnes:
This PR adds an npm script and some config to generate documentation for the GraphQL API.

## Changes
- Add `spectaql` npm package.
- Update gitignore.
- Update README.


src/api/README.md
```diff
@@ -1,6 +1,6 @@
 # Export Insurance Policies API
 
-CRUD API for EXIP applications. This API is built with [keystone](https://keystonejs.com/).
+GraphQL API for EXIP accounts and applications. This API is built with [Keystone](https://keystonejs.com/).
```
Zainzzkk, line 3: isn't it better to keep this as API for EXIPs (incase it does more than accounts and applications etc)

src/api/documentation/config.yml
```diff
@@ -0,0 +1,21 @@
...
+
+extensions:
+  graphqlScalarExamples: true
+
+info:
+  title: EXIPS GraphQL API Reference
+  description: Export Insurance Policies
+
+servers:
+  - url: https://app-exip-api-staging-001.azurewebsites.net/api/graphql
```
Zainzzkk, line 17: can this come from an env?

src/api/README.md
ttbarnes: Good thinking, I think it's better to have a description though? I can't envision the API doing something unrelated to accounts or applications - if it does some other major piece of functionality that's not account or application related, we can just update the description?

src/api/documentation/config.yml
ttbarnes: Had a look, opens doesn't seem like there is a nice way to do it. I actually think it'll be better to not expose this URL anyway, i've chagned it to be just `https://`.
Without a URL or at least a `https`, the documentation config fails.

Zainzzkk (approved):

PR #914, opened 2023-08-31
fix(EMS-1952-53-56): Working with buyer, XLSX and declaration bug fixes

Zainzzkk:
# Issue:
* EMS-1953 - Company without financial year end date would display as 1st January due to invalid date
* EMS-1952 - Working with buyer would show conditional HTML for traded with buyer after selecting yes
* EMS-1956 - Various pages on declarations section had invalid special characters which were not displayed correctly

# Changes made:
* Added check for financial year end date in mapExporter for XLSX to be empty string if no financial year end date
* Removed conditional HTML for traded with buyer and updated tests
* Fixed sql file and substituted with correct characters


src/api/generate-xlsx/map-application-to-XLSX/map-exporter/index.ts
```diff
@@ -79,6 +79,12 @@ export const mapBroker = (application: Application) => {
  */
 const mapExporter = (application: Application) => {
   const { company, companySicCodes, business } = application;
+  let financialYearEndDate = '';
+
+  // financial year end date should only be populated if it exists
+  if (company[FINANCIAL_YEAR_END_DATE]) {
```
ttbarnes, line 85: It looks like we may need to adjust `XLSX_ROW_INDEXES` from this change; some rows in the XLSX should be bold with a larger heading. It has to be done by row indexes, so if some rows become conditional we need to test some scenarios and adjust accordingly

src/ui/templates/insurance/your-buyer/working-with-buyer.njk
```diff
@@ -12,15 +12,6 @@
   {{ CONTENT_STRINGS.PAGE_TITLE }}
 {% endblock %}
 
-{% set tradedHTML %}
-  {{ govukHint({
```
ttbarnes, line 16: Can remove `govukHint` import

src/api/generate-xlsx/map-application-to-XLSX/map-exporter/index.test.ts
```diff
@@ -128,5 +128,49 @@ describe('api/generate-xlsx/map-application-to-xlsx/map-exporter', () => {
...
+          xlsxRow(XLSX.FIELDS[GOODS_OR_SERVICES], business[GOODS_OR_SERVICES]),
+          xlsxRow(XLSX.FIELDS[YEARS_EXPORTING], business[YEARS_EXPORTING]),
+          xlsxRow(XLSX.FIELDS[EMPLOYEES_UK], business[EMPLOYEES_UK]),
+          xlsxRow(XLSX.FIELDS[EMPLOYEES_INTERNATIONAL], business[EMPLOYEES_INTERNATIONAL]),
+          xlsxRow(XLSX.FIELDS[ESTIMATED_ANNUAL_TURNOVER], formatCurrency(business[ESTIMATED_ANNUAL_TURNOVER], GBP_CURRENCY_CODE)),
+          xlsxRow(CONTENT_STRINGS[PERCENTAGE_TURNOVER].SUMMARY?.TITLE, `${business[PERCENTAGE_TURNOVER]}%`),
+
+          // broker fields
+          ...mapBroker(mockApplication),
+        ];
```
ttbarnes, line 170: Can we extract some of the array item expectations into some kind of shared expected array? Otherwise lots of duplication for 2 tests

Zainzzkk: done

src/api/generate-xlsx/map-application-to-XLSX/map-exporter/index.ts
ttbarnes: Discussed - the field will always exist now, so need for this 🎉

ttbarnes (approved):

PR #922, opened 2023-09-01
fix(EMS-1957): Type of policy - Specifically check the submitted type

ttbarnes:
This PR fixes an issue where the "policy type" field could be manipulated and submitted as something other than "single" or "multiple".

## Changes
- Create `isValidPolicyType` helper function.
- Update validation rules for both "policy type" instances in quote and insurance.
- Simplify some validation functions.
- Improve `mapSubmittedData` documentation.
- Add  additional test coverage.
- Rename instances of `isMultiPolicyType` to `isMultiplePolicyType`.


src/ui/server/controllers/insurance/policy-and-export/type-of-policy/validation/index.ts
```diff
@@ -1,19 +1,28 @@
...
 const { POLICY_AND_EXPORTS } = FIELD_IDS.INSURANCE;
 const FIELD_ID = POLICY_AND_EXPORTS.POLICY_TYPE;
 
+const ERROR_MESSAGE = ERROR_MESSAGES.INSURANCE.POLICY_AND_EXPORTS.TYPE_OF_POLICY[FIELD_ID].IS_EMPTY;
+
+/**
+ * validation
+ * Check if the policy type is a valid type and not empty
+ * @param {Express.Response.body} Express response body
+ * @param {Object} Errors object from previous validation errors
```
Zainzzkk, line 16: only 1 param in the function

ttbarnes: Good spot

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #924, opened 2023-09-04
chore(EMS-1299): E2E tests - Common assertSummaryListRow command

Zainzzkk:
# Changes made: 
* Added shared commands for asserting summaryList row, value, key and changeLink
* Updated tests to use it


e2e-tests/commands/assert-summary-list-row-change-link.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {Object} Summary list cypress selectors from pages/partials.
+ * @param {String} Field ID in the summary list to assert
+ * @param {String} expectedChangeLinkText text for the change link
+ */
+const assertSummaryListRowChangeLink = (summaryList, fieldId, expectedChangeLinkText) => {
+  if (expectedChangeLinkText) {
+    cy.checkText(
+      summaryList.field(fieldId).changeLink(),
+      expectedChangeLinkText,
+    );
```
ttbarnes, line 13: Should we assert the URL as well as text with `checkLink()`?

e2e-tests/commands/assert-summary-list-row.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {String} expectedChangeLinkText expected change link text - either add/change or should not exist
+ */
+const assertSummaryListRow = (summaryList, fieldId, expectedKey, expectedValue, expectedChangeLinkText) => {
+  const row = summaryList.field(fieldId);
+
+  cy.assertSummaryListRowKey(summaryList, fieldId, expectedKey);
+
+  /**
+   * if value exists then checkText
+   * else check if dash
```
ttbarnes, line 19:
```
if value exists then check the value
else, check for default empty string
```

e2e-tests/commands/assert-summary-list-row.js
```diff
@@ -0,0 +1,48 @@
...
+        row.changeLink(),
+        `${LINKS.CHANGE} ${expectedChangeLinkText}`,
+      );
+    } else {
+      cy.assertSummaryListRowChangeLink(summaryList, fieldId, `${LINKS.ADD} ${expectedChangeLinkText}`);
+    }
+  } else {
+    cy.assertSummaryListRowChangeLink(summaryList, fieldId);
+  }
+};
```
ttbarnes, line 46: Looks like we don't need `checkText` on line 36 as `assertSummaryListRowChangeLink` does this?

This feels strange to read as it has a `expectedChangeLinkText, param, but then different text is checked ("add"/"change"?) Could it just be this?

```js
if (expectedChangeLinkText) {
  cy.assertSummaryListRowChangeLink(summaryList, fieldId, expectedChangeLinkText);
} else {
  cy.assertSummaryListRowChangeLink(summaryList, fieldId);
}
```

Also looks 4 params are being passed to `assertSummaryListRowChangeLink` but it only uses 3 of them.

e2e-tests/commands/assert-summary-list-row-change-link.js
Zainzzkk: discussed.  Changed name of function since we are testing url elsewhere

PR #935, opened 2023-09-05
chore(deps): Update dependencies

ttbarnes:


package.json
```diff
@@ -46,7 +46,7 @@
     "@commitlint/cli": "^17.6.7",
     "@commitlint/config-conventional": "^17.6.7",
     "@types/jest": "^29.5.3",
-    "@types/node": "^20.4.7",
+    "@types/node": "^20.5.9",
```
Zainzzkk, line 49: is this the main node version being updated? Might need to check compatibility with Azure

ttbarnes: This is just for type checking, i.e not the actual node version

Zainzzkk (approved):

PR #950, opened 2023-09-07
chore(EMS-1965): Optimise "create application" API call

ttbarnes:
This PR creates a new GQL resolver, `createAnApplication` so that the UI can make a single API call to create an application - removing 3 other API calls that were previously necessary and also causing the call to take longer than it should.

## Changes
- Create new GQL resolver, `createAnApplication`. This mutation:
  - Gets a country from a provided isoCode.
  - Creates a new application with owner relationship.
  - Creates an eligibility relationship.
  - Creates a buyer relationship.
  - Updates the application with relationship IDs.
- Remove now unnecessary relationship creation logic in the keystone schema's application hooks.
- Create new API test helpers to:
  - Get an application.
  - Update an application.
  - Create a buyer.
  - Create an eligibility.
- Add mock countries to API test mocks.
- Update the UI's "create application" GQL call to use the new mutation.
- Update all instances in the UI where the "create application" GQL call is use a new function `mapEligibilityAnswers` - this allows us to send all eligibility answers with a `buyerCountryIsoCode` field, instead of a `buyerCountry` object.
- Update the required eligibility function to include `buyerCountryIsoCode`.
- Update "create applications" functionality in E2E tests to:
  - Get a country and create a buyer with a country.
  - Create applications with a buyer relationship.
- Fix some typos.


src/api/custom-resolvers/mutations/create-an-application/index.ts
```diff
@@ -0,0 +1,121 @@
...
+        },
+      },
+    });
+
+    /**
+     * Create eligibility and relationships for:
+     * 1) The buyer country
+     * 2) The application
+     */
+    const eligibility = await context.db.Eligibility.createOne({
```
Zainzzkk, line 63: this file is quite long.  What if we put these in their own files? like eligibility and buyer

src/api/custom-resolvers/mutations/create-an-application/index.ts
```diff
@@ -0,0 +1,121 @@
...
+        },
+      },
+    });
+
+    /**
+     * Update the application with relationships for:
+     * 1) Buyer
+     * 2) Eligibility
+     */
+    const updatedApplication = await context.db.Application.updateOne({
```
Zainzzkk, line 96: if we had this in a shared helper/function, could be used in test-helpers as well

src/api/custom-resolvers/mutations/create-an-application/index.ts
ttbarnes: Great shout, i've moved eligibility and buyer into helper functions. 

Would be great to do the same with `Application`, but there are few instances of this, would like to do this (along with other context methods), in another PR as it would make this PR even bigger. Will create a ticket if that's ok.

src/api/custom-resolvers/mutations/create-an-application/index.ts
ttbarnes: Same as above

Zainzzkk (approved):

Zainzzkk (approved):

PR #964, opened 2023-09-12
feat(EMS-1979): Cypress optimisation

abhi-markan:
## Introduction
Optimise cypress tests execution

## Resolution
* Previous job successful validation using `needs`
* Project segregation.
* Project ID as a secret.
* Concurrency introduction
* Concurrent machine id as a tag `#1` on Cypress cloud.
* Increased concurrent machines to `10`.
* Set `fail-fast` to `true`.


ttbarnes (dismissed):
Beautiful

.github/workflows/test.yml
```diff
@@ -213,20 +220,25 @@ jobs:
...
     environment:
       name: ${{ needs.setup.outputs.environment }}
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: true
+      # Single dimension matrix - 10 jobs
       matrix:
-        machines: [1, 2, 3, 4]
+        machines: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
Zainzzkk, line 237: not a blocker, quote runs really quick with 4-6 if we wanted to save resources

.github/workflows/test.yml
```diff
@@ -269,18 +283,25 @@ jobs:
...
+    needs: [setup, unit-tests, api-tests, e2e-tests-insurance, e2e-tests-quote]
     environment:
       name: ${{ needs.setup.outputs.environment }}
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: true
+      # Single dimension matrix - 10 jobs
+      matrix:
+        machines: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
Zainzzkk, line 300: 1 machine is enough for this, it runs on 1 machine in less than 5mins usually

Zainzzkk (dismissed):

e2e-tests/switch-between-services/cypress.config.js
```diff
@@ -9,7 +9,6 @@ dotenv.config();
  * Cypress configuration
  */
 const cypressConfig = defineConfig({
-  projectId: 'sefntb',
```
ttbarnes, line 12: Why is `projectId` removed for all instances?

ttbarnes (approved):

PR #984, opened 2023-09-14
fix(EMS-1159): Dashboard pagination - Fix issue where "total pages" count would not round up

ttbarnes:
This PR fixes an issue where the dasboard pagination would not diplay if the total amount of applications is just over the maximum amount allowed per page, for example if there are ~16-19 applications, pagination would not display because the total pages functionality was not rounding **_up_** to the nearest number.

## Changes
- Update `getTotalPages` function to use `Math.ceil` instead of `Math.round`, therefore it'll always round up.
- Add E2E test coverage for pagination when there are 16 applications.

### DIFF
--- a/src/ui/server/helpers/pagination/get-total-pages/index.test.ts
+++ b/src/ui/server/helpers/pagination/get-total-pages/index.test.ts
@@ -7,7 +7,7 @@ describe('server/helpers/pagination/get-total-pages', () => {
       const mockTotalApplications = MAX_APPLICATIONS_PER_PAGE + 1;
       const result = getTotalPages(mockTotalApplications);
 
-      const expected = Math.round(mockTotalApplications / MAX_APPLICATIONS_PER_PAGE);
+      const expected = Math.ceil(mockTotalApplications / MAX_APPLICATIONS_PER_PAGE);
 
       expect(result).toEqual(expected);
     });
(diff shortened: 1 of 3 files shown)


src/ui/server/helpers/pagination/get-total-pages/index.test.ts
```diff
@@ -7,7 +7,7 @@ describe('server/helpers/pagination/get-total-pages', () => {
       const mockTotalApplications = MAX_APPLICATIONS_PER_PAGE + 1;
       const result = getTotalPages(mockTotalApplications);
 
-      const expected = Math.round(mockTotalApplications / MAX_APPLICATIONS_PER_PAGE);
+      const expected = Math.ceil(mockTotalApplications / MAX_APPLICATIONS_PER_PAGE);
```
abhi-markan, line 10: Could there be risk of getting a higher number i.e. 
```javascript
Math.ceil(8.000000000000001) => 9
Math.round(8.000000000000001) => 8
```

Is above a possibility?

abhi-markan (changes requested):
Good work, one minor comment.

src/ui/server/helpers/pagination/get-total-pages/index.test.ts
ttbarnes: @abhi-markan I if understand the concern correctly - no, the `totalApplications` will always be a whole number, it's essentially a `.length` from the API. `MAX_APPLICATIONS_PER_PAGE` is a constant set at 15. So this will never having anything other than whole numbers.

We have a lot of tests for making sure the total pages/pagination items are rendered correctly with the following:

- 15 applications (no pagination)
- 16 applications (this is when pagination starts displaying)
- 45 applications
- 60 applications
- 75 applications
- 90 applications
- 135 applications

abhi-markan: Understood thanks for the clarification.

abhi-markan (approved):

PR #992, opened 2023-09-15
fix(EMS-1760): added sub resource integrity to all the script tags

abhi-markan:
## Introduction 🖊️ 
Ensure all the scripts have integrity hash introduced

## Resolution ✔️ 
* Added SRI to all the internally hosted scripts, with hash calculated using `SHA-512` algorithm.
* Enhanced CSP policies
* Improved NJK template.
* Added `og:*` meta attributes
* Migrated inline script to external `google-analytics.js` file.
* Introduced 2 new middlewares:
  * SRI: To populate SRIs using `content-string` constants.
  * Meta-data: To populte `og:*` respective properties using the same.
* Mock `res.local` loaded from `mockRes`.

## Miscellaneous ➕ 
* Re-build UI assets.
* Refactored `cookies.js`.
* Updated `webpack.prod.config.js`
* Removed `js-enabled.js`.


src/ui/templates/index.njk
```diff
@@ -16,17 +16,17 @@
     <link rel="apple-touch-icon" href="/assets/images/govuk-apple-touch-icon.png">
 
     <link rel="stylesheet" href="/assets/css/styles.css" />
-    <!--[if IE 8]>
```
abhi-markan, line 19: Note this has been removed, as `IE8` browser is not longer supported by MS and therefore we should not be supporting legacy browsers.

src/ui/templates/index.njk
```diff
@@ -45,14 +45,20 @@
...
-      <script async src="https://www.googletagmanager.com/gtag/js?id={{ googleAnalyticsId }}"></script>
-      <script>window.dataLayer = window.dataLayer || []; function gtag(){dataLayer.push(arguments);} gtag('js', new Date()); gtag('config', '{{ googleAnalyticsId }}');</script>
-    {% endif %}
+      {%block scripts %}
+      <script src="/assets/js/jsEnabled.js" type="text/javascript" integrity="sha512-LUW2MMgLIyJzPPvhODhZMciLaqLNgBGrSMMj6zOJsH+7uSahGknt3c+VSzyUYlSnngNXZwvvNA7kc+Zvx0Mh9A==" crossorigin="anonymous"></script>
+      <script src="/assets/js/govukFrontend.js" type="text/javascript" integrity="sha512-EdyNYxz0W40gqYt1R48s+ye7mW0p7SwUTnkyUcMJ0eGRsucWodRmVpC5+QAtUBZSYeOMeWumRLnqJZ2zzkDRqg==" crossorigin="anonymous"></script>
+      <script src="/assets/js/formSubmission.js" type="text/javascript" integrity="sha512-uyTmQxJNGlnB71lDj1CCOUhKJLch3djXW1Av0BOT4g7K7riFXsDbyUmhNkhUrtpbdFICkh3lFxmviTlOKNTh7w==" crossorigin="anonymous"></script>
+      <script src="/assets/js/mojFrontend.js" type="text/javascript" integrity="sha512-Ou5NaBw3i2mTX9wG4T5qe8iVVv6Y/67xtvY1ByVY3fYf82Mc2vLGaXfs2z89f9IJ6GCKC64Cv+oRKKYWszDi5Q==" crossorigin="anonymous"></script>
```
abhi-markan, line 53: Note I have added `mojFrontend.js` here as the file size is relatively small and easier to update SRI when file is updated rather than updating numerous files.

src/ui/templates/index.njk
```diff
@@ -16,17 +16,17 @@
     <link rel="apple-touch-icon" href="/assets/images/govuk-apple-touch-icon.png">
 
     <link rel="stylesheet" href="/assets/css/styles.css" />
-    <!--[if IE 8]>
-      <link rel="stylesheet" href="/govuk-frontend-ie8-3.5.0.min.css">
-    <![endif]-->
 
+    <meta property="og:locale" content="en_GB" />
+    <meta property="og:type" content="website" />
+    <meta property="og:title" content="" />
```
abhi-markan, line 22: @ttbarnes @Zainzzkk need to talk about this as I am unable to fetch `pageTitle` value.

src/ui/scripts/cookies.js
```diff
@@ -7,72 +7,93 @@ const https = window.location.protocol == 'https:';
...
 
-      if (answer === 'accept') {
-        createAcceptCookie();
-      }
-    }
-  });
-}
+// Call init when element is loaded
+cookieBanner.addEventListener('load', () => {
+  console.log('Loaded cookie');
```
ttbarnes, line 97: `console.log`?

src/ui/templates/index.njk
```diff
@@ -16,17 +16,17 @@
...
+    <meta property="og:description" content="" />
     <meta property="og:image" content="/assets/images/govuk-opengraph-image.png">
+    <meta property="og:image:width" content="512" />
+    <meta property="og:image:height" content="512" />
   </head>
 
   <body class="govuk-template__body govuk-body">
-
-    <script src="/assets/js/jsEnabled.js" type="text/javascript"></script>
```
ttbarnes, line 28: jsEnabled removed - I don't think we need this actually. Looks like the script/map still exists though, can probably remove now?

src/ui/templates/index.njk
```diff
@@ -45,14 +45,20 @@
...
-    {% endif %}
+      {%block scripts %}
+      <script src="/assets/js/jsEnabled.js" type="text/javascript" integrity="sha512-LUW2MMgLIyJzPPvhODhZMciLaqLNgBGrSMMj6zOJsH+7uSahGknt3c+VSzyUYlSnngNXZwvvNA7kc+Zvx0Mh9A==" crossorigin="anonymous"></script>
+      <script src="/assets/js/govukFrontend.js" type="text/javascript" integrity="sha512-EdyNYxz0W40gqYt1R48s+ye7mW0p7SwUTnkyUcMJ0eGRsucWodRmVpC5+QAtUBZSYeOMeWumRLnqJZ2zzkDRqg==" crossorigin="anonymous"></script>
+      <script src="/assets/js/formSubmission.js" type="text/javascript" integrity="sha512-uyTmQxJNGlnB71lDj1CCOUhKJLch3djXW1Av0BOT4g7K7riFXsDbyUmhNkhUrtpbdFICkh3lFxmviTlOKNTh7w==" crossorigin="anonymous"></script>
+      <script src="/assets/js/mojFrontend.js" type="text/javascript" integrity="sha512-Ou5NaBw3i2mTX9wG4T5qe8iVVv6Y/67xtvY1ByVY3fYf82Mc2vLGaXfs2z89f9IJ6GCKC64Cv+oRKKYWszDi5Q==" crossorigin="anonymous"></script>
+      <script src="/assets/js/cookies.js" type="text/javascript" integrity="sha512-1MU/q7OlnUX2SMKAZWls7I/EfPiitbX+ym3Dzj+f5r4qI+nQKF3Q6LZk3LRk3cqO726dQBlwhhEcJW455LM69w==" crossorigin="anonymous"></script>
```
ttbarnes, line 54: Could we create constants for the integrity values and then pass to the template via `corePageVariables`? https://github.com/UK-Export-Finance/exip/blob/main-application/src/ui/server/helpers/page-variables/core/index.ts#L55

We use this for all generic strings etc

src/ui/templates/insurance/account/sign-in/sign-in.njk
```diff
@@ -150,6 +150,4 @@
 
   </form>
 
-  <script src="/assets/js/mojFrontend.js" type="text/javascript"></script>
```
ttbarnes, line 153: Looks like MOJ frontend will be called in all pages now(?) If so, I think we should revert - currently only need to load the scripts for pages with password field (for show/hide functionality). Otherwise it seems the script could be run on every page when only ~1% of the pages need it

src/ui/templates/index.njk
ttbarnes: Will have a look into this. I don't think it's possible to consume variables from a controller on this level. In other pages we use `CONTENT_STRINGS.PAGE_TITLE`, generated from core page variables and then call `{% block pageTitle %}`

src/ui/templates/index.njk
abhi-markan: Have added across all the pages apart from `index.njk` as it appears it does not call `corePageVariables`.
Would be good to look into it on Monday as it will also help resolve `og:*` meta values population.

src/ui/server/content-strings/integrity.ts
```diff
@@ -0,0 +1,8 @@
+export const INTEGRITY = {
+  ACCESSIBILITY: 'sha512-Y98fJqTDpxfo1b4+WVq2QXK1dXFAHqsLOVCx5TrKVDA5DxYxUWg9F1DQ52R6IgZQ/mVRuj8vTSagcsWDqKCcTw==',
```
Zainzzkk, line 2: should these be in envs or fine here?

src/ui/server/controllers/insurance/account/create/your-details/index.ts
```diff
@@ -99,6 +105,12 @@ export const post = async (req: Request, res: Response) => {
   let validationErrors = generateValidationErrors(payload);
 
   if (validationErrors) {
+    // Add SRI to the locals
```
Zainzzkk, line 108: could we have a function for this to be dry? can pass it res.locals.SRI

src/ui/server/middleware/meta/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param res - The response object used to send the HTTP response.
+ * @param next - The next function to be called after setting the `res.locals.meta` property.
+ */
+export const meta = (req: Request, res: Response, next: () => void) => {
+  // Set the meta information in res.locals
+  res.locals.meta = {
+    url: `${req.hostname}${req.originalUrl}`,
+    title: PRODUCT.DESCRIPTION.GENERIC,
+    organisation: PRODUCT.DESCRIPTION.ORGANISATION,
+  };
```
ttbarnes, line 16: Shall we capitilise these so `URL`, `TITLE` etc?

src/ui/server/middleware/meta/index.test.ts
```diff
@@ -0,0 +1,45 @@
+import { mockNext, mockReq, mockRes } from '../../test-mocks';
+import { Request, Response } from '../../../types';
+import { meta } from '.';
+import { PRODUCT } from '../../content-strings';
+
+const req: Request = mockReq();
+const res: Response = mockRes();
+const next = mockNext;
+
+describe('meta-data middleware unit test cases', () => {
```
ttbarnes, line 10: Can we align with other unit tests, we just have the name of the path and no need to mention it's a test 😛  keep it simple

```js
describe('middleware/headers/meta' () => ...
```

src/ui/server/controllers/insurance/account/create/your-details/index.ts
ttbarnes: Agree, there is already `integrity` middleware (nice) so doing it in the controller I believe is just doing the same thing twice. The middleware will pass `res.locals` to the controller and therefore the rendered template with scripts.

We do this with "get application" every controller that requires an application from the API just consumes `res.locals.application` from `get-application` middleware

Seems we can just remove `res.locals.SRI` from all controller because of the middleware, DRY

ttbarnes: Sorry currently the middleware doesn't return all SRI's but if it does we wouldn't need such things in the controllers.

Also if we need to add the scripts to other or new pages, with the controller approach, there's a risk that passing SRI could be forgotten, but it'd be done automatically with the middleware

src/ui/server/middleware/meta/index.test.ts
abhi-markan: Have updated.

abhi-markan (commented):
Have refactored. Thanks.

src/ui/server/middleware/integrity/index.ts
```diff
@@ -0,0 +1,30 @@
...
+    GOVUK,
+    FORM,
+    COOKIES,
+    GA,
+  };
+
+  res.locals.SRI = {
+    ...res.locals.SRI,
+    ...SRI,
+  };
```
ttbarnes, line 27: I think it can just be `res.locals.SRI = SRI;`, not a blocker

src/ui/types/express/index.d.ts
```diff
@@ -23,13 +23,30 @@ interface RequestHeaders {
...
+
 interface ResponseLocals {
   application?: Application;
   csrfToken: string;
   cookieConsent?: boolean;
   cookieConsentDecision?: boolean;
   cookieConsentNewDecision?: boolean;
   googleAnalyticsId?: string;
+  meta?: MetaData;
+  SRI?: SRI;
```
ttbarnes, line 49: I think all of these new types should be required so no `?`

src/ui/server/middleware/integrity/index.ts
abhi-markan: Have pushed.

ttbarnes (dismissed):

Zainzzkk (dismissed):

ttbarnes (dismissed):

src/ui/server/controllers/insurance/business/companies-house-number/index.test.ts
```diff
@@ -231,7 +231,7 @@ describe('controllers/insurance/business/companies-house-number', () => {
 
     describe('when there is no application', () => {
       beforeEach(() => {
-        res.locals = { csrfToken: '1234' };
+        res.locals = { csrfToken: '1234', SRI: {}, meta: {} };
```
ttbarnes, line 234: We should create a `mock-locals` in tests mocks, not a blocker. Tech improvement just making a note

src/ui/server/middleware/integrity/index.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Request, Response } from '../../../types';
+import { integrity } from '.';
+import { INTEGRITY } from '../../constants';
+
+const { GOVUK, FORM, COOKIES, GA, MOJ, ACCESSIBILITY } = INTEGRITY;
+const req: Request = mockReq();
+const res: Response = mockRes();
+const next = mockNext;
+
+describe('sub resource integrity (SRI) middleware unit test cases', () => {
```
ttbarnes, line 11: For all new middleware can we match the other test descriptions? Just the directory name

e.g

```js
describe('middleware/cookies-consent', () => {
```

I don't think we need to mention they're unit tests in the describe block

src/ui/types/express/index.d.ts
```diff
@@ -23,13 +23,30 @@ interface RequestHeaders {
...
+  ACCESSIBILITY?: string;
+  GA?: string;
+}
+
+interface MetaData {
+  URL?: string;
+  TITLE?: string;
+  ORGANISATION?: string;
+}
+
```
ttbarnes, line 40: Seems all of these should be required instead of optonal? Not a blocker

ttbarnes (dismissed):
Great work

ttbarnes (approved):

PR #1005, opened 2023-09-19
fix(cypress): GHA matrix test execution

abhi-markan:
## Introduction
Ensure Cypress tests are executed on runners only.

## Resolution
* Introduced spec matrix.
* Execution via CLI
  * Video recording set to false
  * Screenshot upon failure set to false.

### DIFF
--- a/e2e-tests/switch-between-services/cypress.config.js
+++ b/e2e-tests/switch-between-services/cypress.config.js
@@ -9,6 +9,7 @@ dotenv.config();
  * Cypress configuration
  */
 const cypressConfig = defineConfig({
+  video: true,
   numTestsKeptInMemory: 1,
   viewportWidth: 1920,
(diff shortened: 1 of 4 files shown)


e2e-tests/switch-between-services/cypress.config.js
```diff
@@ -9,14 +9,15 @@ dotenv.config();
  * Cypress configuration
  */
 const cypressConfig = defineConfig({
+  video: true,
```
ttbarnes, line 12: `video`? Same above

abhi-markan: This is for local tests only, when they fail I can see the recording. This is overwritten during CLI execution on CI.

ttbarnes (approved):

ttbarnes (approved):

PR #1006, opened 2023-09-19
chore(EMS-1963): Move UI CIS API calls and mappings into the API

ttbarnes:
## Introduction ✏️ 
- The UI was calling the APIM CIS (Country information system) API directly instead of via our own API.
- The UI was then:
  - Mapping the countries in various controllers before being able to consumes the countries.
  - Executing some functions in various controllers to determine a country's level of support.
- This means that the UI is not simply consuming a list of countries.
- We went our UI to only call one API (our own API).
- This change removes 1 of 2 external API calls that are made directly in the UI.

## Resolution ✔️ 
- Create a new GQL resolver in the API, `getApimCisCountries`.
- Replace the UI’s CIS API call with the new GQL call to our own API.
- Move all CIS country mappings and “country support” checks from the UI, to the API, using a new `MappedCisCountry` type.
- Split up all CIS country mappings and “country support” checks into their own directories, also improved test coverage.
- The UI now just consumes certain flags returned in a response instead of running the logic throughout the UI. E.g `if (country.canApplyOnline) { ...` instead of `if (canApplyOnline(country)) { ... `.
- Update appropriate test mocks so we have mock countries, mock CIS countries, mock mapped CIS countries.

## Miscellaneous ➕ 
- Remove all instaces of `.only` in tests.
- Rename/move API constants to “external-apis”.


src/api/custom-resolvers/queries/get-APIM-CIS-countries/index.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      const expected = { success: false };
+
+      expect(response).toEqual(expected);
+    });
+  });
+
+  describe('when APIM CIS API returns empty data object', () => {
+    beforeEach(() => {
+      APIM.getCisCountries = jest.fn(() => Promise.resolve({ success: true, data: undefined }));
```
Zainzzkk, line 44: do we need the success: true in here?

src/api/custom-resolvers/queries/get-APIM-CIS-countries/index.ts
```diff
@@ -0,0 +1,30 @@
+import APIM from '../../../integrations/APIM';
+import mapCisCountries from '../../../helpers/map-CIS-countries';
+
+/**
+ * getApimCisCountries
+ * Get countries from APIM
+ * @param {Object} GraphQL root variables
```
Zainzzkk, line 7: do we need this param?

src/api/helpers/map-CIS-countries/index.ts
```diff
@@ -0,0 +1,32 @@
...
+import mapCisCountry from './map-CIS-country';
+import sortArrayAlphabetically from '../sort-array-alphabetically';
+import { CisCountry } from '../../types';
+
+const { CIS } = EXTERNAL_API_DEFINITIONS;
+
+/**
+ * filterCisCountries
+ * Filter out countries from CIS API that have an invalid name
+ * @param {Array} All CIS countries
```
Zainzzkk, line 11: not a blocker, should we set this as `Array<CisCountry>`

src/api/helpers/map-CIS-countries/map-CIS-country/cannot-get-a-quote/index.ts
```diff
@@ -0,0 +1,17 @@
+import { MappedCisCountry } from '../../../../types';
+
+/**
+ * cannotGetAQuote
+ * Check if a country cannot get a quote online or offline
+ * @param {Object} Country from CIS API
```
Zainzzkk, line 6: should this be {MappedCisCountry}?

src/api/helpers/map-CIS-countries/map-CIS-country/map-short-term-cover-available/index.ts
```diff
@@ -0,0 +1,33 @@
...
+} = EXTERNAL_API_DEFINITIONS;
+
+/**
+ * mapShortTermCoverAvailable
+ * Transform a countries 'short term cover available' string to a boolean
+ * @param {String} Risk category
+ * @returns {Boolean}
+ */
+export const mapShortTermCoverAvailable = (str: string): boolean => {
+  if (str === SHORT_TERM_COVER_AVAILABLE.YES) {
```
Zainzzkk, line 14: could this be more efficient in a switch statement?

src/ui/server/api/keystone/APIM/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import apollo from '../../../graphql/apollo';
+import getApimCisCountries from '../../../graphql/queries/APIM/CIS-countries';
+
+const APIM = {
+  getCisCountries: async () => {
+    try {
+      const response = (await apollo('GET', getApimCisCountries, {})) as ApolloResponse;
+
+      if (response.errors) {
+        console.error('GraphQL network error querying NEW API CALL %O', response.errors);
```
Zainzzkk, line 11: should it be APIM getCisCountries instead of NEW API CALL? same for rest in this file

src/api/helpers/map-CIS-countries/map-CIS-country/index.ts
```diff
@@ -0,0 +1,39 @@
...
+import canGetAQuoteByEmail from './can-get-a-quote-by-email';
+import cannotGetAQuote from './cannot-get-a-quote';
+import canApplyOnline from './can-apply-online';
+import canApplyOffline from './can-apply-offline';
+import { CisCountry, MappedCisCountry } from '../../../types';
+
+/**
+ * mapCisCountry
+ * Map a CIS country to cleaner structure
+ * @param {Object} CIS Country
```
Zainzzkk, line 14: should these be object or CisCountry and MappedCisCountry?

src/api/helpers/map-CIS-countries/index.ts
ttbarnes: Great shout, i've added a return type as well.

src/api/helpers/map-CIS-countries/map-CIS-country/map-short-term-cover-available/index.ts
ttbarnes: Great shout, this was created ages ago, have updated to switch/case

src/api/custom-resolvers/queries/get-APIM-CIS-countries/index.test.ts
ttbarnes: Yes, it's a required type for `APIM.getCisCountries`

src/ui/server/api/keystone/APIM/index.ts
ttbarnes: Whoops, have updated

Zainzzkk (approved):

PR #1022, opened 2023-09-20
chore(EMS-2008): Rename DB table, API schema - policyAndExport to policy

ttbarnes:
## Introduction ✏️ 
In the next phase of EXIP, "Policy and export" will be split into 2 different sections/areas. Therefore, to avoid data migration/mappings in the future, on the data side, we should rename "policy and export" to simply "policy".

This is the first step in ensuring that the next phase will not introduce or will minimise any data changes as the design and requirements evolve.

Note - the UI intentionally retains some naming of "policy and export", for example with summary lists, routes, controller names etc. This is because the design is currently "policy and export". We are just updating the data side of things here. In the next phase, the UI's "policy and export" section will be split up and renamed.

## Resolution ✔️ 
- Rename all appropriate instances of `policyAndExport` / `policy and export` to `policy`.
- Rename UI saving functions e.g `application.update.policy`, `mapAndSave.policy`.
- Rename `mockSinglePolicyAndExport` to mockSinglePolicy.
- Rename `mockMultiplePolicyAndExport` to mockMultiplePolicy.
- Update types.


src/api/content-strings/XLSX.ts
```diff
@@ -27,7 +27,7 @@ export const XLSX = {
   SECTION_TITLES: {
     KEY_INFORMATION: 'Key information',
     EXPORTER_CONTACT_DETAILS: 'Exporter contact details',
-    POLICY_AND_EXPORT: 'Type of policy and exports',
+    POLICY_AND_EXPORT: 'Type of policys',
```
Zainzzkk, line 30: policies?

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #1025, opened 2023-09-21
chore(EMS-2009): Create exportContract DB table/API schema, owning 2x policy and export fields

ttbarnes:
## Introduction ✏️ 
In the next phase of EXIP, 2x "Policy and export" fields will be moved into a different section/area - "your export contract".

To avoid data migration/mappings in the future, on the data side, we should move these fields into a new "export contract" DB table.

This is the second step in ensuring that the next phase will not introduce or will minimise any data changes as the design and requirements evolve.

## Resolution ✔️ 
- Create new DB table `exportContract`.
- Move policy and export `DESCRIPTION` and `FINAL_DESTINATION` fields to "export contract", 
- Update application creation to create a new `exportContract` instance/relationship.
- Create UI API calls, mapping and save data functions for saving an application's `exportContract`. This is currently consumed in the policy and export's "about goods or services" controller. In the next phase, these fields will be moved into an "export contract" section.
- Update all appropriate references, mappings, test helpers and mocks including, but not limited to:
  - Summary lists.
  - UI's "get application" GQL query.
  - UI's `flattenApplicationData` function.

## Miscellaneous ➕
- Fix some typos.


src/ui/server/controllers/insurance/export-contract/save-data/index.ts
```diff
@@ -0,0 +1,35 @@
+import api from '../../../../api';
+import getDataToSave from '../../../../helpers/get-data-to-save';
+import { sanitiseData } from '../../../../helpers/sanitise-data';
+import { Application, RequestBody } from '../../../../../types';
+
+/**
+ * exportContract
+ * Strip invalid fields from submitted form data and update the application.
+ * This is used for any save functionality in the Export contract section of the application.
+ * @param {Object} Application
```
Zainzzkk, line 10: wrong type but not blocker

src/ui/server/controllers/insurance/policy-and-export/map-and-save/index.ts
```diff
@@ -5,7 +5,7 @@ import { Application, RequestBody, ValidationErrors } from '../../../../../types
 
 /**
  * mapAndSave
- * Map and save any valid Single contract policy fields
+ * Map and save any valid  policy fields
  * @param {Express.Request.body} Express request body
  * @param {Object} Application
```
Zainzzkk, line 10: same here

Zainzzkk (approved):

src/ui/server/controllers/insurance/policy-and-export/map-and-save/index.ts
ttbarnes: Discussed - we have this in many places, will update all in one go.

PR #1026, opened 2023-09-21
chore(EMS-2018): Enforce GQL operation name requirement

ttbarnes:
## Introduction ✏️ 
Our API has some logic to check if a requested operation (GQL query/mutation), is in an "allowed list". If the provided operation name is not in an allowed list, the request would be rejected.

However, this could be bypassed by not providing an `operationName` in the GQL query/mutation.

## Resolution ✔️ 
- Update all GQL queries/mutations to include an operation name.
- Update the API's Apollo operation check to specifically reject a request if an operation name is not provided.
- Update allowed GQL resolvers - renaming recently created or changed resolvers, fixing a typo.

## Miscellaneous ➕
- Simplify some "get accounts" usages/instances, fixing an issue with "get accounts" GQL query and calls in E2E tests, where the call would be rejected due to missing `operationName` in the call made via `cy.request` instead of apollo client. Needed to construct a `req.body` with `query` and `operationName`.


e2e-tests/commands/api/index.js
```diff
@@ -162,7 +162,7 @@ const queryStrings = {
       }
     `,
     getLatestHowDataWillBeUsed: () => gql`
-      query declarationHowDataWillBeUsed {
+      query declarationHowDataWillBeUseds {
```
Zainzzkk, line 165: Should it not be `declarationHowDataWillBeUsed ` and not `declarationHowDataWillBeUseds `?

ttbarnes: It is strange, but this is correct. All declarations are fetched essentially by a `findAll`. The Keystone GQL generates this available GQL query by adding an `s` to the entity type. E.g this could be `policyAndExports`. 

This is how all the other declarations are, just this declaration had an incorrect name.

Zainzzkk (approved):

Zainzzkk (approved):

abhi-markan (approved):

PR #1042, opened 2023-09-22
fix(EMS-806): Task list - remove link from eligibility task

ttbarnes:
## Introduction ✏️ 
This PR fixes an issue where the "eligibility" task in the "initial checks" group of an application's task list would display an empty link ("#").

## Resolution ✔️ 
- Remove "href" definition in `createInitialChecksTasks`.
- Update `generateTaskStatusesAndLinks` task mapping logic to optionally add a href to a task.
- Update `taskLink` function to return an empty string instead of null.
- Make a tasks href an optional type.

## Miscellaneous ➕
- Extract task mapping logic into its own function, `mapTask`.
- Fixed a typo.
- Simplified some E2E task assertions.

### DIFF
--- a/e2e-tests/insurance/cypress/e2e/journeys/all-sections.spec.js
+++ b/e2e-tests/insurance/cypress/e2e/journeys/all-sections.spec.js
@@ -53,7 +53,7 @@ context('Insurance - All sections - new application', () => {
       });
 
       describe('tasks', () => {
-        it('should render a `check eligibility` task with `completed` status', () => {
+        it('should render a `check eligibility` task with `completed` status, no link', () => {
           cy.navigateToUrl(url);
 
           const task = taskList.initialChecks.tasks.eligibility;
@@ -65,9 +65,9 @@ context('Insurance - All sections - new application', () => {
             expectedText,
           );
 
-          const expectedStatus = TASKS.STATUS.COMPLETED;
+          cy.checkText(task.status(), TASKS.STATUS.COMPLETED);
 
-          cy.checkText(task.status(), expectedStatus);
+          task.link().should('not.exist');
         });
       });
     });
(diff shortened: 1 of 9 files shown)


e2e-tests/insurance/cypress/e2e/journeys/all-sections.spec.js
```diff
@@ -51,17 +53,15 @@ context('Insurance - All sections - new application', () => {
...
-          const expectedHref = '#';
           const expectedText = TASKS.LIST.INITIAL_CHECKS.TASKS.ELIGIBILITY;
 
-          cy.checkLink(
-            task.link(),
-            expectedHref,
+          cy.checkText(
+            task.text(),
             expectedText,
           );
```
Zainzzkk, line 66: test to check link does not exist?

ttbarnes: Good shout

src/ui/server/helpers/task-list/index.ts
```diff
@@ -1,6 +1,25 @@
...
+ * @param {ApplicationFlat} Submitted application data
+ * @returns {Object} Tasks with status and optional href
+ */
+export const mapTask = (task: TaskListDataTask, submittedData: ApplicationFlat) => {
+  const mapped = {
+    ...task,
+    status: taskStatus(task, submittedData),
+  };
+
+  if (task.href) {
```
abhi-markan, line 16: Should this be `task.href` since `src/ui/types/task-list.ts` sets `href` as optional?

abhi-markan (changes requested):

src/ui/server/helpers/task-list/index.ts
ttbarnes: Yes, it's checking if a href has been provided (because it's optional). If so, it will add a href to the task.

abhi-markan (approved):

PR #1071, opened 2023-09-28
chore(EMS-2053): Improve E2E test speed

ttbarnes:
## Introduction ✏️ 
As the application has grown, we have added a lot of E2E tests and as a result, the time to run E2E tests has grown.

As we move into the next phase, where a lot more journeys and E2E tests will be added, we should reduce the time it takes to run PRs/E2E tests as much as possible.

This PR almost halves the time it takes to run E2E tests; from ~30/31 minutes to ~16.5 minutes. 🚀 

## Resolution ✔️ 
- Update `test.yml` workflow so that specific test directories, therefore creating different jobs in PR E2E tests.
- Add `createAnApplication` API call to the E2E test suite with an accompanying cypress command.
- Update/simplify `completeInsuranceEligibilitySignInAndGoToDashboard` cypress command so that it calls a new cypress command, `signInAndAssertAllSectionsUrl`.
- Update `completeSignInAndGoToApplication` cypress command to create an application with eligibility answers directly via the API to save going through the full eligibility user journey in all E2E tests.
  - This is set by default. Otherwise, if a false flag for `createApplicationViaApi` is passed, the API will not be called and the usual user journey will happen.
- Extract some cypress command functions into a new command, `createAnAccountAndApplicationAndSignIn`.
- Update "check your eligibility answers" E2E test to assert each row's value from `FIELD_VALUES` and mock country fixture (via application fixture), since the E2E eligibility fixture has been changed to have real values that are now sent to the API.
- Update all instances of `completeSignInAndGoToApplication` to use a new object param structure.

## Miscellaneous ➕
- Move `baseUrl` to the top of various E2E tests to be DRY and consistent.


e2e-tests/commands/insurance/create-an-application.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * createAnApplication
+ * Create an application directly via the API
+ * 1) Get a country
+ * 2) Create a buyer with country ID
+ * 3) Generate mock application with owner and buyer relationships
+ * 4) Create applications
+ * @param {String} Account ID for the application owner
+ * @param {Number} Amount of applications to create
```
Zainzzkk, line 14: extra param to remove?

Zainzzkk (approved):

e2e-tests/commands/insurance/create-an-application.js
ttbarnes: Discussed, will merge and address this in another PR so that we can get quicker tests in all other PRs.

PR #1083, opened 2023-09-29
fix(EMS-1759): Enable `__Host-` prefixed cookies

abhi-markan:
## Introduction ✏️ 
* Ensure `__Host-` prefix cookies can be created on `localhost`, `E2E` and GHA.
* Ensure browser based basic authentication is removed in-line with IaC roadmap

## Resolution ✔️ 

image

* Introduced `https` server which wraps around existing Express server.
* Introduced new environment variable `TLS_CERTIFICATE` and `TLS_KEY`.
* Updated hard-coded `http` protocol to `https`
* Introduced `COOKIE` constant.
* Removed `BASIC_AUTH_KEY` and `BASIC_AUTH_SECRET`.
* Removed Cypress basic authentication login.
* Removed `express-basic-auth` package.
* Improved test coverage.

## Miscellaneous ➕ 
* Constants refactorisation.
* Dependencies updates.
* Documentation updates.
* Added `ignoreDeprecations` to `src/ui/tsconfig.json`

## Note ⚠️ 
When visiting locally on `https://localhost:4000` user will receive HTTPS error due to self-signed certificate, which is to be expected.

image


e2e-tests/commands/core-page-checks.js
```diff
@@ -37,12 +37,7 @@ const checkBackLink = (currentHref, expectedHref) => {
   cy.url().should('eq', expectedUrl);
 
   // go back to current page
-  cy.visit(`${Cypress.config('baseUrl')}${currentHref}`, {
-    auth: {
-      username: Cypress.config('basicAuthKey'),
-      password: Cypress.config('basicAuthSecret'),
-    },
-  });
```
ttbarnes, line 45: This auth is required for basic-auth - without this, no page will load

src/ui/server/index.ts
```diff
@@ -111,18 +104,6 @@ ui.use(
...
-    basicAuth({
-      users: {
-        // @ts-ignore
-        [process.env.BASIC_AUTH_KEY]: process.env.BASIC_AUTH_SECRET, // @ts-ignore
-      },
-      challenge: true,
-    }),
-  );
-}
-
```
ttbarnes, line 125: Without basic auth, am I right in thinking, anyone could access e.g dev or staging environment?

abhi-markan: That is correct, however before deployment IaC will be updated to only allow VPN connections.

e2e-tests/commands/save-session.js
```diff
@@ -1,11 +1,40 @@
...
+];
+
+const flags = {
+  secure: true,
+  httpOnly: true,
+  path: '/',
+  log: false,
+};
+
+const saveSession = () => {
```
ttbarnes, line 15: Could we add some documentation and maybe link to/mention "based on preserve cookie package"?

src/ui/server/constants/application/index.ts
```diff
@@ -1,8 +1,12 @@
 import LATEST_VERSION_NUMBER from './versions/latest';
 
+export const SERVICE_NAME = 'EXIP';
+export const INSURANCE = 'Insurance';
+export const QUOTE = 'Quote';
```
ttbarnes, line 5: This might be breaking some things, the moving of `SERVICE_NAME` and `QUOTE` into `application` constants; quote is not part of the application/"insurance"

I think only `INSURANCE` should be in here. In the interests of progressing this PR, to rule out this potentially causing E2E issues, maybe move `SERVICE_NAME` and `QUOTE` back into root constants and we can extract later.

src/ui/server/index.ts
```diff
@@ -35,11 +34,6 @@ import isQuoteRoute from './helpers/is-quote-route';
 
 // @ts-ignore
 const ui = express();
-ui.disable('x-powered-by');
```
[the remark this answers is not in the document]
abhi-markan: It is still required, the middleware `security/index.ts` caters for this now.

ttbarnes (dismissed):
LGTM

ttbarnes (approved):

PR #1092, opened 2023-10-03
chore(EMS-1962-2065): Split up cypress commands, DRY field selectors

ttbarnes:
## Introduction ✏️ 
As the application has grown, we have added a lot of cypress commands, selectors and tests. As a result, some selectors are almost exactly the same and could be DRY.

As we move into the next phase, where a lot more journeys and E2E tests will be added, we should ensure the suite is using the same approaches/patterns.

## Resolution ✔️ 
- Split up all cypress commands and their relevant functions into directories, instead of having a single commands file.
- Simplify the `pages/shared/field` function.
- Update all E2E tests that check a field input/label etc, to use `pages/shared/field`, instead of having its own, same definition, but with a different field ID.
- Add some missing `-input` data-cy attributes to be aligned with other inputs.

## Miscellaneous ➕
- Remove assertions that check that labels and hints exists; No point in having this when a `checkText` function is called on the label.
- Remove some duplicate tests for submit buttons, they are already tested in `corePageChecks`.
- Update some tests to have simpler destructuring for `INSURANCE_FIELD_IDS` and `INSURANCE_ROUTES`.
- Update some tests to have a single declaration of `baseUrl`.
- Update some functions to be commands, e.g `checkChangeLinkUrl` `changeAnswerField`, instead of importing the functions inside a test.


e2e-tests/commands/insurance/complete-and-submit-about-goods-or-services-form.js
```diff
@@ -16,7 +16,8 @@ const {
 } = FIELD_IDS;
 
 export default () => {
-  cy.keyboardInput(aboutGoodsOrServicesPage[DESCRIPTION].input(), application.EXPORT_CONTRACT[DESCRIPTION]);
+  cy.keyboardInput(aboutGoodsOrServicesPage[DESCRIPTION].textarea(), application.EXPORT_CONTRACT[DESCRIPTION]);
```
Zainzzkk, line 19: any reaon why this is not field().textArea and line 21 is not field().input?

ttbarnes: So with these changes, the text areas and country inputs are the only fields that are different when using the generic `field` helper, needs some extra work, would like to clean up after.

Basically the country inputs have a single ID e.g `country`, instead of `fieldId-input` like all other fields. Same with textareas, they have a single ID due to GOV design limitation.

Want to see what else we can do to clean these up but don't want to bloat this PR anymore 🙃

e2e-tests/commands/shared-commands/index.js
```diff
@@ -0,0 +1,12 @@
+import '@cypress-audit/lighthouse/commands';
```
Zainzzkk, line 1: are we still using lighthouse?

Zainzzkk (approved):

e2e-tests/commands/shared-commands/index.js
ttbarnes: Will investigate this separately

PR #1094, opened 2023-10-04
fix(tls-escape): added additional backslash escape

abhi-markan:
## Introduction ✏️ 
Ensure TLS certificates are escaped.

## Resolution ✔️ 
* Added helper function responsible for escaping `\n` and `\` in TLS certificate and key.
* Removed `GITHUB_SHA` from Dockerfile.
* Added `NODE_ENV` to Dockerfile.
* Streamlined `UI_PORT`, `API_PORT` and `PORT` environment variable declaration.

## Miscellaneous ➕ 
* Dependencies updates


src/ui/server/helpers/sanitise-environment/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Sanitises an environment variable string by replacing occurrences of all the sequence
+ * '\\n' with a newline character ('\n'). Note this function has also replace escaped backslash (\).
+ * @param string - The input string that needs to be sanitised.
+ * @returns The sanitised version of the input string.
+ * @example `ABC\\nCBA` will be sanitised to
+ * `ABC
+ * CBA`
+ */
+export const sanitise = (string: string | undefined): string | undefined => (string ? string.replace(/\\\\n/g, '\n') : string);
```
Zainzzkk, line 10: could do `(string?: string)` so don't need undefined

Zainzzkk (approved):

PR #1095, opened 2023-10-04
fix(escape): escaping environment variables with double quotes

abhi-markan:
## Introduction
Ensure double quotes environment variables successfully escapes `\\n|\\\n|\\\\n`.

## Resolution
* Updated RegEx with values in double quotes.
* Function argument data type fix

### DIFF
--- a/src/ui/server/helpers/sanitise-environment/index.test.ts
+++ b/src/ui/server/helpers/sanitise-environment/index.test.ts
@@ -8,9 +8,9 @@ describe('helper/sanitise-environment/sanitise', () => {
     expect(result).toBe(input);
   });
 
-  // Replaces all occurrences of `\\\\n` with `\n` in the input string
-  it('should replace all occurrences of `\\\\n` with `\n` in the input string', () => {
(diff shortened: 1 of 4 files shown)


ttbarnes (dismissed):

src/ui/server/helpers/sanitise-environment/index.test.ts
```diff
@@ -22,23 +22,23 @@ describe('helper/sanitise-environment/sanitise', () => {
     expect(result).toBe('');
   });
 
-  // Returns the input string if it does not contain `\\\\n`
-  it('should return the input string if it does not contain `\\\\n`', () => {
+  // Returns the input string if it does not contain `\\\n`
+  it('should return the input string if it does not contain `\\\n`', () => {
```
ttbarnes, line 26: Not a blocker - I don't think we need a description of the test as the test describes what it's testing 😛 but lets get this merged

ttbarnes (dismissed):

Zainzzkk (approved):

PR #1102, opened 2023-10-06
fix(cookies-security): local TLS server and trust proxy for Azure

abhi-markan:
## Introduction ✏️ 
Ensure `__Host-` prefix cookies are being served across all the environments including localhost.

## Resolution ✔️ 
* Introduced `trust proxy` for Azure, where the container is being executed on HTTP protocol behind the proxy.
* Conditional server listening, when not set to `production` (localhost) then `HTTPS` server is being run.
* `NODE_ENV` moved to GitHub vars.
* Completed Dockerfile build arguments.
* `NODE_ENV` being effectively for CICD and Docker build.
* `UI_PORT` and `API_PORT` environment variable for docker build and Azure has been removed.
* Introduced `isProduction` helper function.

## Miscellaneous ➕ 
* Dependencies updates.
* Documentation updates.


src/ui/server/helpers/http/index.ts
```diff
@@ -0,0 +1,18 @@
...
+dotenv.config();
+const { PORT } = process.env;
+
+/**
+ * Sets up error handling and listening for an Express server.
+ * @param server - An instance of the Express server.
+ * @returns void
+ */
+
+// @ts-ignore
```
ttbarnes, line 14: Not a blocker - i'm sure there's a way to pass express server as types. I'll pick this up 👍

abhi-markan: Thanks, I was unable to find it, was hoping to speak with you regarding this anyways.

src/ui/server/helpers/is-production/index.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  it('should return false when NODE_ENV is set to an unexpected value', () => {
+    process.env.NODE_ENV = '';
+    expect(isProduction()).toBe(false);
+  });
+
+  it('should return false when NODE_ENV is set to an unexpected value', () => {
+    process.env.NODE_ENV = ' ';
+    expect(isProduction()).toBe(false);
+  });
```
ttbarnes, line 32: Not a blocker (in the interests of getting ready for PEN test) - duplicate tests here and the test description doesn't match the file path - I can pick this up, few improvements I have in mind can do in one PR 👍

src/ui/server/index.ts
```diff
@@ -166,4 +179,13 @@ ui.get('*', (req: Request, res: Response) => {
...
+/**
+ * When being executed on Azure only HTTP server should be running,
+ * since TLS will be stripped. However ensure localhost and GHA
+ * are executed on TLS.
+ */
+if (isProduction()) {
+  http(ui);
+} else {
+  https(ui);
+}
```
ttbarnes, line 191: Really not a fan of environment specific logic in a codebase. I think "if test/dev environment" is ok, but "if production" is risker. Not a blocker, understand we have to do this for the Azure restriction(s).

Could we add a specific comment/description about the restriction, e.g "Azure handles https so we have to execute as http inside a wrapper", instead of "only http server..."? Just if you skim read this it feels strange

src/ui/server/helpers/http/index.ts
ttbarnes: Will have a look 😊 When I see `ts-ignore` I go into investigation mode

ttbarnes (approved):

PR #1104, opened 2023-10-09
fix(EMS-2009): Application submission XLSX - fix missing final destination 

ttbarnes:
## Introduction ✏️ 
This PR fixes an issue where, after some data changes, the XLSX generated during application submission had a missing "final destination" country value.

## Resolution ✔️ 
- Update `mapPolicyAndExportIntro` function to get the value from `exportContract.finalDestination.name` instead of the (invalid) `exportContract.finalDestinationCountryCode.name`

## Miscellaneous ➕
- Create a new API test helper to generate an application with submission date: `generateSubmittedApplication`.
- Create a new API test helper to generatean object of application related IDs, `mapApplicationIds`.
- Update various API unit tests to consume `generateSubmittedApplication` instead of `mockApplication`, otherwise some data is unavailable and tests fail.
- Update `createFullApplication` API test helper to consume a policy type flag and use multiple contract policy answers if appropriate.


src/api/constants/field-ids/insurance/policy-and-exports/index.ts
```diff
@@ -26,6 +26,7 @@ const POLICY_AND_EXPORTS = {
   ABOUT_GOODS_OR_SERVICES: {
     DESCRIPTION: 'goodsOrServicesDescription',
     FINAL_DESTINATION: 'finalDestinationCountryCode',
+    FINAL_DESTINATION_OBJECT: 'finalDestinationCountry',
```
Zainzzkk, line 29: should this be country? and not object?

src/api/test-helpers/generate-submitted-application.ts
```diff
@@ -0,0 +1,29 @@
...
+import getPopulatedApplication from '../helpers/get-populated-application';
+import createFullApplication from './create-full-application';
+import mapApplicationIds from './map-application-ids';
+
+/**
+ * generateSubmittedApplication
+ * 1) Create an application.
+ * 2) Populate the application.
+ * 3) Create an object with submitted date
+ * @param {Object} Context
```
Zainzzkk, line 11: no params passed

src/api/constants/field-ids/insurance/policy-and-exports/index.ts
ttbarnes: Good shout

Zainzzkk (approved):

PR #1107, opened 2023-10-10
chore(EMS-2088): Eligibility - save "need pre-credit period cover" in application policy

ttbarnes:
## Introduction ✏️ 
In the next phase of EXIP, the "Need pre-credit period cover" field will no longer be in the "Eligibility" section; instead it will be in the "policy" section of an application.

To avoid data migration/mappings in the future, on the data side, we should move these fields into a new "export contract" DB table.

This is a pure API/DB PR, excluding a minor GQL query change in the UI.

## Resolution ✔️ 
- Update SQL dump; Also removes some fields that are no longer relevant.
- Create new default `DEFAULT_NEED_PRE_CREDIT_PERIOD_COVER` API constant.
- Update `createAnApplication` mutation to strip out `needPreCreditPeriodCover` and create a new `policy` entry with application relationship.
- Create `createAPolicy` helper function.
- Remove policy hooks in keystone hooks for an application (now created via helper function.
- Update `nullableCheckbox` keystone component to accept an optional default value.
- Update `createFullApplication` test helper to manually create and return a policy entry.
- Create a test helper to create a new policy.
- Update various types.

## Miscellaneous ➕
- Fix some typos.


src/api/keystone.test.ts
```diff
@@ -27,14 +28,27 @@ describe('Create an Application', () => {
       },
     });
 
-    // update the application with buyer ID.
+    // create policy and associate with the application.
+    const policy = await policies.create({
```
Zainzzkk, line 32: would this be better in a helper to create for us?

ttbarnes: This is calling a helper function? Same as the other create/update methods in this file:

```
applications.create(...
buyers.create(...
applications.update(...
```

Zainzzkk (approved):

PR #1119, opened 2023-10-11
fix(EMS-1736): Quote - Tell us about your policy - fix 'percentage of cover' population issue

ttbarnes:
## Introduction ✏️ 
In the quote tool's "tell us about your policy" page, if there are form submission errors, the "percentage of cover" field would not pre-populated when the form is re-rendered.

## Resolution ✔️ 
- Update the UI controller to map "percentage of cover" from `req.body` instead of payload, when there are validation errors.
- Add E2E coverage.
- Update an "insured for" test assertion - now that percentage of cover is pre-populated, the answer is different.

### DIFF
--- a/e2e-tests/quote/cypress/e2e/journeys/quote/your-quote/your-quote-change-answers-single-policy.spec.js
+++ b/e2e-tests/quote/cypress/e2e/journeys/quote/your-quote/your-quote-change-answers-single-policy.spec.js
@@ -179,7 +179,7 @@ context('Your quote page - change answers (single policy type to multiple policy
 
       const insuredFor = summaryList.field(QUOTE.INSURED_FOR);
 
-      cy.checkText(insuredFor.value(), '£84,000.00');
+      cy.checkText(insuredFor.value(), '£108,000.00');
 
       const policyLength = summaryList.field(POLICY_LENGTH);
 


Zainzzkk (dismissed):

e2e-tests/quote/cypress/e2e/journeys/quote/your-quote/your-quote-change-answers-single-policy.spec.js
```diff
@@ -179,7 +179,7 @@ context('Your quote page - change answers (single policy type to multiple policy
 
       const insuredFor = summaryList.field(QUOTE.INSURED_FOR);
 
-      cy.checkText(insuredFor.value(), '£84,000.00');
+      cy.checkText(insuredFor.value(), '£108,000.00');
```
Zainzzkk, line 182: missed this, can this be a constant?

ttbarnes: Would be great - we have about 30 instances of such assertions, for prices that are ultimately generated from the pricing grid. Constants might actually hinder these assertions when they're updated, e.g what could we call these as constants? 🤔 

```
'£11,111,110.20'
'£143,209.86'
```

I think this should be separate from this PR as there are many instances and this PR unblocks a ticket that's been in sprints for ~4 months

Zainzzkk (approved):

PR #1169, opened 2023-10-18
chore(EMS-2125-2126): API/GQL security improvements

ttbarnes:
## Introduction ✏️ 
This PR updates the API config to address some security issues reported by a PEN test.

## Resolution ✔️ 
- Add a function for custom GQL error formatting, `formatGraphQlError`. If the environment is not development, this will return a generic error message instead of exposing some information about the GQL schema, for example: _"Cannot query field "iso" on type "Country". Did you mean "isoCode"?"_
- Add `overload-protection`.
- Add some API security headers.

## Miscellaneous ➕
- Extract "apollo plugins" into an apollo directory.


src/api/middleware/headers/security/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Global middleware, ensures myriads of imperative security headers.
+ * - `HSTS` - 1 Year
+ * - `X-Frame-Options` - Clickjacking
+ * - `X-Content-Type-Options` - Content-Type headers should be followed
+ * @param {Express.Request} req Request object
+ * @param {Express.Response} res Response object
+ * @returns {Function} next()
+ */
+const security = (req: Request, res: Response, next: NextFunction) => {
+  res.setHeader('Strict-Transport-Security', 'max-age=15552000; includeSubDomains; preload');
```
abhi-markan, line 13: Should we also add remaining security headers from `ui`?

abhi-markan (changes requested):
One comment to address please.

src/api/middleware/headers/security/index.ts
ttbarnes: Good shout, have added all the same headers as the UI. Now both are aligned 👍

abhi-markan (approved):

PR #1185, opened 2023-10-20
fix(EMS-2080): Account verification resolver returning success=false when already verified

ttbarnes:
## Introduction ✏️ 
The account verification GQL resolver would return `success=false` when an account has already been verified. This was because when the resolver is called again, an account would not be found from the provided verification token, because it has been (correctly) deleted.

This ultimately results in the user being redirected to an "invalid link" page.

## Resolution ✔️ 
- Update the `verifyAccountEmailAddress` GQL resolver to:
  - Get an account from a provided account ID variable.
  - Check if the account is already verified, if so, return `success=true`.
- Update the UI's "verify email" controller to call the API with an account ID param, sent via `req.query`.
- Update the UI's `query-params` middleware to:
  - Allow a maximum of 2 query params (instead of 1).
  - Return a 400 status if the amount of query params exceeds 2.
  - Sanitise each filtered/allowed query param before passing it onto the request.
- Update `verifyAccountEmail` cypress command and other relevant functions to consume and pass an ID query param.

## Miscellaneous ➕
- FIx some invalid/outdated documentation.
- Remove previously added logs for debugging in the dev environment.
- Minor E2E test improvements.


abhi-markan (approved):
LGTM 👍🏼

abhi-markan (approved):

PR #1244, opened 2023-11-02
feat(EMS-2113): No PDF - Eligibility - Handle inactive companies, update company not found logic

ttbarnes:
## Introduction ✏️ 
This PR updates the "company house search" functionality in insurance eligibility to redirect to an exit page if a company is inactive.

## Resolution ✔️ 
- Add comapnies house number example constants so that the company types we need to test (e.g multiple SIC codes, no SIC code, no financial year end date etc) are active companies).
- Update `mapCompaniesHouseFields` API function to be simpler and map/return `isActive`.
- Update "eligibility - company search" UI GET controller to redirect if the companies house response does _not_ contain "isActive=true".
- Create UI "eligibility company inactive" GET controller, page, route etc.
- Update "company not found" logic so that if companies house returns a `404` response, we return a `notFound` flag and render a validation error.
- Add E2E test coverage.

## Miscellaneous ➕
- Various E2E test improvements, including extracting some eligibility action content checks into shared commands, since they are now used in multiple places.
- Extract some shared page content actions into a generic `COMPANIES_HOUSE_EXIT` object.
- Simplify `mapCompaniesHouseFields` helper.
- Simplify UI/API companies house validation errors - this is now only handled by the UI.
- Remove E2E test for a company that is active, but has no SIC codes. This is very edge case and seems to be for historical companies. We tried to update the companies house number we use for the test, but the company changed from "active" to "dissolved". We want to keep the logic for this, but it seems an E2E test is not possible, or would be flaky, if the company is updated.


src/api/integrations/companies-house/index.ts
```diff
@@ -26,6 +26,13 @@ const companiesHouse = {
         },
       });
 
+      if (response.status === 404) {
```
Zainzzkk, line 29: wondering if in the future, these codes should be constants

src/ui/server/controllers/insurance/eligibility/companies-house-search/index.ts
```diff
@@ -64,51 +64,59 @@ export const get = (req: Request, res: Response) =>
...
       return res.render(TEMPLATE, {
         ...insuranceCorePageVariables({
           PAGE_CONTENT_STRINGS,
           BACK_LINK: req.headers.referer,
         }),
         userName: getUserNameFromSession(req.session.user),
         ...PAGE_VARIABLES,
-        validationErrors,
+        validationErrors: formValidationErrors,
+        submittedValues: req.body,
```
Zainzzkk, line 84: this should be payload

src/ui/server/controllers/insurance/eligibility/companies-house-search/index.ts
```diff
@@ -64,51 +64,59 @@ export const get = (req: Request, res: Response) =>
...
       return res.render(TEMPLATE, {
         ...insuranceCorePageVariables({
           PAGE_CONTENT_STRINGS,
           BACK_LINK: req.headers.referer,
         }),
         userName: getUserNameFromSession(req.session.user),
         ...PAGE_VARIABLES,
-        validationErrors: response.validationErrors,
+        validationErrors: generateValidationErrors({}),
+        submittedValues: req.body,
```
Zainzzkk, line 106: payload

src/api/integrations/companies-house/index.ts
ttbarnes: Would be good, can we look into this later please. Quite a few instances of status checking, not sure what we would call them either. I.e  i don't believe eslint will allow us to create a constant starting with a number.

src/ui/server/controllers/insurance/eligibility/companies-house-search/index.ts
ttbarnes: Good spot, updated

src/ui/server/controllers/insurance/eligibility/companies-house-search/index.ts
ttbarnes: Updated

Zainzzkk (approved):

src/ui/server/controllers/insurance/eligibility/companies-house-search/index.ts
Zainzzkk: has this been changed? I can't see it

ttbarnes: Just checked, yes it's been changed. Must be github cache, happens sometimes

PR #1245, opened 2023-11-03
chore(EMS-2206): Replace wantCoverOverMaxAmount data saving with total contract value IDs

ttbarnes:
## Introduction ✏️ 
This PR updates the eligibility data saving functionality, during application creation, so that we no longer save the "Do you want cover over 500k?" question as a `wantCoverOverMaxAmount` boolean.

Instead, we now reference an ID via a new DB table, `TotalContractValue`; so the `wantCoverOverMaxAmount` field is used purely in the UI's insurance eligibility flow.

## Resolution ✔️ 
- Update DB dump to include `TotalContractValue` table.
- Add `TOTAL_CONTRACT_VALUE` and `TOTAL_CONTRACT_VALUE_ID` to field ID constants.
- Add `TOTAL_CONTRACT_VALUE` constants.
- Update eligibility test fixtures/mocks to use `TOTAL_CONTRACT_VALUE` instead of `WANT_COVER_OVER_MAX_AMOUNT`.
- Update the GQL "create an application" mutation to get a total contract value via new helper function, `getTotalContractValueByField`.
- Update XLSX mapping so that the `WANT_COVER_OVER_MAX_AMOUNT` row will continue to display a "No" answer.
- Update the `creatAnEligibility` helper function to consume a `totalContractValue` ID and return a `totalContractValue` relationship.
- Update the `getPopulatedApplication` helper function to get and return a `TotalContractValue`.
- Update keystone schema.
- Update the UI's `mapEligibilityAnswers` helper function so that `TOTAL_CONTRACT_VALUE_ID` is mapped and `WANT_COVER_OVER_MAX_AMOUNT` is not included. This mapping is used to send data to the API for application creation.
- Update the task list's `requiredFields` function for "eligibility".

## Miscellaneous ➕
- Update the "all sections" E2E test to generate an application without using the API, so its going through the full user journey.


src/api/helpers/get-total-contract-value-by-field/index.ts
```diff
@@ -0,0 +1,37 @@
+import { Context } from '../../types';
+
+const getTotalContractValueByField = async (context: Context, field: string, value: number): Promise<object | boolean> => {
```
Zainzzkk, line 3: do we need some documentation at the top?

src/ui/server/helpers/map-eligibility-answers/index.ts
```diff
@@ -1,15 +1,28 @@
...
 
 const mapEligibilityAnswers = (answers: SubmittedDataInsuranceEligibility) => {
   if (answers.buyerCountry) {
-    const { buyerCountry, ...otherAnswers } = answers;
+    const { buyerCountry, wantCoverOverMaxAmount, ...otherAnswers } = answers;
 
     const mapped = {
       ...otherAnswers,
       [BUYER_COUNTRY_ISO_CODE]: buyerCountry?.isoCode,
+      [TOTAL_CONTRACT_VALUE_ID]: mapTotalContractValue(Boolean(wantCoverOverMaxAmount)),
```
Zainzzkk, line 25: this would be better in a const and then passed to `mapTotalContractValue`

src/api/helpers/get-total-contract-value-by-field/index.ts
ttbarnes: Good spot - have moved this helper into a new directory and added documentation.

src/ui/server/helpers/map-eligibility-answers/index.ts
ttbarnes: Updated

Zainzzkk (approved):

PR #1248, opened 2023-11-03
chore(EMS-2206): Replace wantCoverOverMaxPeriod data saving with cover period IDs

ttbarnes:
## Introduction ✏️ 
This PR updates the eligibility data saving functionality, during application creation, so that we no longer save the "Do you want cover over 2 years?" question as a `wantCoverOverMaxPeriod` boolean.

Instead, we now reference an ID via a new DB table, `CoverPeriod`; so the `wantCoverOverMaxPeriod` field is used purely in the UI's insurance eligibility flow.

## Resolution ✔️ 
- Update DB dump to include `CoverPeriod` table.
- Add `COVER_PERIOD` and `COVER_PERIOD_ID` to field ID constants.
- Add `COVER_PERIOD` constants.
- Update eligibility test fixtures/mocks to use `COVER_PERIOD` instead of `WANT_COVER_OVER_MAX_AMOUNT`.
- Update the GQL "create an application" mutation to get a total contract value via new helper function, `getCoverPeriodByField`.
- Update XLSX mapping so that the `WANT_COVER_OVER_MAX_AMOUNT` row will continue to display a "No" answer.
- Update the `creatAnEligibility` helper function to consume a `totalContractValue` ID and return a `totalContractValue` relationship.
- Update the `getPopulatedApplication` helper function to get and return a `CoverPeriod`.
- Update keystone schema.
- Update the UI's `mapEligibilityAnswers` helper function so that `COVER_PERIOD_ID` is mapped and new, appropriate fields are not included. This mapping is used to send data to the API for application creation.
- Update the task list's `requiredFields` function for "eligibility".


database/exip.sql
```diff
@@ -473,6 +473,31 @@ VALUES
...
 
+# Dump of table CoverPeriod
+# ------------------------------------------------------------
+
+DROP TABLE IF EXISTS `CoverPeriod`;
+
+CREATE TABLE `CoverPeriod` (
+  `id` varchar(191) COLLATE utf8mb4_unicode_ci NOT NULL,
+  `valueId` int DEFAULT NULL,
+  `value` varchar(191) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
```
abhi-markan, line 484: Varchar set as `191`,  based on test below can we set it to `70` please as max characters are at around 16.

src/api/test-helpers/cover-period.ts
```diff
@@ -0,0 +1,35 @@
...
+      data: {
+        valueId: COVER_PERIOD.LESS_THAN_2_YEARS.DB_ID,
+        value: 'Mock value',
+      },
+      query: coverPeriodQuery,
+    })) as CoverPeriod;
+
+    return coverPeriod;
+  } catch (err) {
+    console.error(err);
```
abhi-markan, line 26: Can we add a message to `console.error` please with test `console.error('Error when creating cover period %s', error);`.

src/ui/server/helpers/map-cover-period-id/index.ts
```diff
@@ -0,0 +1,16 @@
+import { COVER_PERIOD } from '../../constants';
+
+/**
+ * mapCoverPeriodId
+ * @param {Boolean} answer: cover period answer
+ * @returns {Number} cover period DB ID
+ */
+export const mapCoverPeriodId = (answer: boolean) => {
+  if (answer) {
```
abhi-markan, line 9: Ternary for single line?

abhi-markan (changes requested):
Good work mate, only few minor comments to address please.

database/exip.sql
ttbarnes: Good shout, have added this for `CoverPeriod`and `TotalContractValue`, in the DB dump and keystone schema.

src/api/test-helpers/cover-period.ts
ttbarnes: Can we have this as tech debt? Can do after merging this PR. There's several test helpers we could do this for, would like to do in one go. Otherwise happy to do now. These are purely used for API unit tests.

src/ui/server/helpers/map-cover-period-id/index.ts
ttbarnes: I think its harder to read as a single 1 line with ~150 characters instead of a few short lines, but have made the change including to the similar `map-total-contract-value` function

src/api/test-helpers/cover-period.ts
abhi-markan: Sounds good.

abhi-markan (approved):

PR #1347, opened 2023-11-16
feat(EMS-2191): No PDF - Your business - Alternative trading address validation

Zainzzkk:
## Introduction ✏️ 
This PRs adds validation to the alternative trading address page

## Resolution ✔️ 
- Added validation to the alternative trading address page
- Updated unit and e2e tests


src/ui/server/controllers/insurance/business/alternative-trading-address/validation/rules/alternative-trading-address.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param {object} errors
+ * @returns {object} errors
+ */
+const alternativeTradingAddress = (responseBody: RequestBody, errors: object) => {
+  // if body is empty
+  if (!objectHasProperty(responseBody, FIELD_ID)) {
+    return generateValidationErrors(FIELD_ID, ERROR_MESSAGE.IS_EMPTY, errors);
+  }
+
+  // check if the field is above the maximum
```
ttbarnes, line 28: Can we use the shared validation function here? `helpers/shared-validation/max-length`, there's an example in `other-comments` validation

ttbarnes (approved):

PR #1366, opened 2023-11-21
chore(deps): Bump dependencies

ttbarnes:
## Introduction ✏️ 

## Resolution ✔️ 
- Update dependencies.


src/api/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # 1. BUILD
 ###################
 
-FROM node:20.8.1-alpine3.17 AS build
+FROM node:21.2.0-alpine3.17 AS build
```
Zainzzkk, line 5: this sometimes breaks Azure.  Will need to check and revert if it does

Zainzzkk (approved):

Zainzzkk (approved):

src/api/Dockerfile
ttbarnes: Have reverted this

PR #1425, opened 2023-11-29
feat(EMS-2279-2280): No PDF - Insurance policy - start page

ttbarnes:
## Introduction :pencil2:
This PR adds a new start page to the beginning of the "Insurancy policy" section of a no PDF application.

- Create new cypress command `startInsurancePolicySection`.
- Update all relevant "Insurance policy" E2E tests to call `startInsurancePolicySection`, instead of clicking a link in the task list.
- Add E2E test coverage for the new "Insurancy policy" start page.
- Add a new route, controller, template and content strings for the "Insurancy policy" start page.

- Replace duplicate definitions of `baseUrl` in all "Insurance policy" E2E tests.


src/ui/server/controllers/insurance/policy/index.test.ts
```diff
@@ -0,0 +1,73 @@
...
+import { mockApplication, mockReq, mockRes } from '../../../test-mocks';
+
+const {
+  INSURANCE_ROOT,
+  ALL_SECTIONS,
+  POLICY: { TYPE_OF_POLICY },
+  PROBLEM_WITH_SERVICE,
+} = INSURANCE_ROUTES;
+
+describe('controllers/insurance/policy/index', () => {
```
Zainzzkk, line 17: do we need /index here?

e2e-tests/insurance/cypress/e2e/journeys/policy/single-contract-policy/save-and-back.spec.js
```diff
@@ -142,7 +141,7 @@ context('Insurance - Policy - Single contract policy page - Save and go back', (
     });
 
     it('should have the submitted values when going back to the page', () => {
-      task.link().click();
+      cy.startInsurancePolicySection();
       submitButton().click();
```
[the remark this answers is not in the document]
ttbarnes: I don't think so, this is only applicable for the last `it` in each describe block. The previous `it` blocks should not execute this.

The last `it` is to get back to the page; the other `it` blocks do not want to go back to the page

src/ui/server/controllers/insurance/policy/index.test.ts
ttbarnes: I think so yes because its a root index in the `policy` directory, indicates it's an index file in the root of `policy`. We have the same in the `business` directory and ~10 other files like this. Happy to change if desired.

Zainzzkk (approved):

PR #1439, opened 2023-12-01
feat(EMS-2287 2347 2363): No PDF - Single/multiple policy - currency enhancements

ttbarnes:
## Introduction :pencil2:
This PR updates single/multiple contract currency support to include `JPY` and improves the UX/UI design. This is for both a no PDF application and the quote tool.

- Add `JPY` to `CURRENCIES` fixtures and `SUPPORTED_CURRENCIES` constant. Change `CURRENCIES` into an object structure.
- Create new cypress command, `changeAnswerRadioField`, similar to `changeAnswerSelectField`.
- Create shared "radios" cypress selector function.
- Update policy currency legend/label and error message copy.
- Update various E2E tests to submit/check a policy's currency as a radio option, instead of a select option.
- Simplify "map currencies" logic in the single and multiple contract policy controllers.
- Replace `mapCurrencies` function with `mapCurrenciesAsRadioOptions` and `mapCurrenciesAsSelectOptions`, since they require different mappings.
- Update application single and multiple contract policy nunjucks template to render currency radios with a new component, `currency-radio-inputs.njk`.

- Extract "get supported countries" function into its own function/directory.
- Create new helper function to generate a country/currency name with its ISO code, `nameAndIsoCodeText`. Returns eg. "the country (isoCode)".
- Add some missing E2E `POLICY` constants.
- Remove unrequired `SUPPORTED_CURRENCIES` E2E constant.
- Fix some typos.
- Remove some unnecessary nunjuck imports.


e2e-tests/commands/shared-commands/form/change-answer-radio-field.js
```diff
@@ -0,0 +1,15 @@
+import { radios, submitButton } from '../../../pages/shared';
+/**
+ * changes radio input/option
+ * @param {Object} fieldVariables
+ * @param {Function} input cypress input selector
```
Zainzzkk, line 5: only 1 param being passed.  SHould this be returns?

src/ui/server/helpers/mappings/map-currencies/as-select-options/index.ts
```diff
@@ -0,0 +1,37 @@
...
+ * mapCurrenciesAsSelectOptions
+ * Map all currencies into the required structure for GOV select component.
+ * @param {Array} Array of currency objects
+ * @param {String} Selected currency
+ * @returns {Array} Array of mapped currencies
+ */
+const mapCurrenciesAsSelectOptions = (currencies: Array<Currency>, selectedValue?: string) => {
+  const supportedCurrencies = getSupportedCurrencies(currencies);
+
+  const mapped = supportedCurrencies.map(({ name, isoCode }) => mapSelectOption(name, isoCode, true, selectedValue));
```
Zainzzkk, line 16: should we name a variable as true so we know exactly what is being set as true without going to the function?

e2e-tests/commands/shared-commands/form/change-answer-radio-field.js
ttbarnes: Typo, removed

src/ui/server/helpers/mappings/map-currencies/as-select-options/index.ts
ttbarnes: This has been around for ages 😄  but good shout, have updated

Zainzzkk (approved):

PR #1511, opened 2023-12-13
chore(EMS-1964): Move external UI currency API calls into the EXIP API 

ttbarnes:
## Introduction :pencil2:
This PR moves the last remaning external API call in the UI, into the EXIP API. This ensures that all external API calls are made from our own API, instead of the UI calling an external API.

- Remove `APIM` environment variables from the docker-compose UI service and UI env sample.
- Add `getApimCurrencies` custom GQL resolver.
- Add `SUPPORTED_CURRENCIES` API constants.
- Add currency mapping functions into the API, using a new shared `NAME` field ID to sort alphabetically.
- Add APIM currencies API call to the API.
- Remove the UI's `external` API calls directory, replace with `api.keystone.APIM.getCurrencies`.
- Update all "get currencies" calls to use `api.keystone.APIM.getCurrencies`.
- Simplify UI currency mapping functions so that only needs to map the currencies into a structure for GOV select component options.


src/api/helpers/map-currencies/index.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Currency } from '../../types';
+
+/**
+ * getSupportedCurrencies
+ * Get all supported currencies
+ * @param {Array} Array of all possible currencies
+ * @returns {Array} Array of currencies that EXIP supports
+ */
+export const getSupportedCurrencies = (currencies: Array<Currency>) => {
+  const supported = currencies.filter((currency) => SUPPORTED_CURRENCIES.find((currencyCode: string) => currency.isoCode === currencyCode));
```
Zainzzkk, line 12: do we need the const or could we just return the filter directly?

src/api/helpers/map-currencies/index.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Map and sort currencies.
+ * 1) Filter supported currencies.
+ * 2) Sort the currencies alphabetically.
+ * @param {Array} Array of currency objects
+ * @returns {Array} Array supported currencies
+ */
+const mapCurrencies = (currencies: Array<Currency>) => {
+  const supportedCurrencies = getSupportedCurrencies(currencies);
+
+  const sorted = sortArrayAlphabetically(supportedCurrencies, 'name');
```
Zainzzkk, line 28: should the 'name' be a const since is a sorting type?

src/api/helpers/map-currencies/index.ts
ttbarnes: I did think about this, I think it's good to have the const so that debugging supported currencies is easier. Also if we remove the const it'd be a very long line of code, harder to read IMO

src/api/helpers/map-currencies/index.ts
ttbarnes: Good shout, have created a new, shared `NAME` field ID constant.

Zainzzkk (approved):

PR #1517, opened 2023-12-15
chore(EMS-2414): Add enctype to all forms

ttbarnes:
## Introduction :pencil2:
This PR updates all forms to include an `enctype` attribute with a value of `application/x-www-form-urlencoded`.

Note that we do not currently have any file upload functionality, so the `enctype` attribute does not need to differ.

- Update all instances of html forms.

- Replace all unused form ID attributes with `data-cy`.
- Update E2E `corePageChecks` to have a `hasAForm` parameter, instead of `assertSubmitButton` and check form attributes.

### DIFF
--- a/e2e-tests/commands/core-page-checks.js
+++ b/e2e-tests/commands/core-page-checks.js
@@ -1,5 +1,10 @@
 import { BUTTONS, LINKS, ORGANISATION } from '../content-strings';
-import { backLink as backLinkSelector, heading, submitButton } from '../pages/shared';
+import {
+  backLink as backLinkSelector,
+  form,
+  heading,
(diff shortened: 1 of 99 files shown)


abhi-markan (dismissed):
Looks good, one minor comment:

* Should we have a component test for `<form>` element to ensure the `enctype` property exists and value is never changed to `plain/text`?

ttbarnes:
Thanks @abhi-markan great shout - we actually don't have nunjuck component tests setup - heavier on the E2E side.

I think we can easily update our "core page checks" E2E command to check for form attributes, having a look

Zainzzkk (dismissed):

PR #1569, opened 2023-12-21
chore(EMS-2418): No PDF - Upgrade GOV design package

ttbarnes:
## Introduction :pencil2:
This PR updates the No PDF version of EXIP to use the latest GOV design version, [5.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0), which contains breaking changes.

- Bump GOV design package.
- Update the phase banner to be capitalized.
- Remove client side JS checks for `govuk-button--disabled`, since this class has been deprecated.
- Update, simplify `js-enabled` script.
- Update static express paths and nunjucks config to include new GOV `dist` directory.
- Remove `$govuk-new-link-styles` SCSS declaration, this is now enabled by default.
- Update, simplify favicons as per new guidelines and new file names.
- Replace all instances of `type="text/javascript"` with `type="module"`, as per new guidelines.
- Update integrity hashes for client side JS.


src/ui/public/js/govukFrontend.js.LICENSE.txt
```diff
@@ -0,0 +1,84 @@
+/**
```
ttbarnes, line 1: Not sure if we should keep this in our repo? It's auto generated since the latest version.

Zainzzkk: shall we remove it and see if everything works

Zainzzkk (approved):

PR #1665, opened 2024-01-08
feat(EMS-2522-2523): No PDF - Single contract policy - Total contract value

ttbarnes:
## Introduction :pencil2:

This PR moves the "total contract value" field from the "single contract policy" form, into it's own form, as part of the "single contract policy" flow of a no PDF application.

Additionally, this means that when a user is checking application answers and decides to change a multiple policy to a single policy, we need to redirect to the "total contract value", before continuing to check application answers and ultimately, submit the application.

- Move `TOTAL_CONTRACT_VALUE` field  ID, content strings etc into the `CONTRACT_POLICY.SINGLE` object, instead of being in `EXPORT_VALUE`.
- Create new cypress commands `completeTotalContractValueForm`, `completeAndSubmitTotalContractValueForm`.
- Update various E2E tests and commands to complete and submit the "total contract value" form (with or without `policyMaximumValue`), after submitting the "single contract policy" form, as per the user journey.
- Create new routes, controllers and templates for "total contract value". Move content strings.
- Move existing "check and change total contract value" functionality to work with the new form/endpoints.
- Update "Single contract policy" POST controller to redirect to "total contract value" if the application does not have `TOTAL_CONTRACT_VALUE` data.
- Update `fieldset-legend` nunjucks component to accept a `isPageHeading` flag. If passed, the legend will render with a page heading.

- Various minor E2E and destructuring improvements.
- Remove some unecessary `isPageHeading` flags in nunjucks templates.


e2e-tests/insurance/cypress/e2e/journeys/policy/broker/save-and-back.spec.js
```diff
@@ -103,6 +104,8 @@ context('Insurance - Policy - Broker page - Save and back', () => {
       submitButton().click();
       // submit single contract policy form
       submitButton().click();
+      // submit total contract value form
+      submitButton().click();
```
Zainzzkk, line 108: I've had an idea.  What if we make a cypress function with a for loop and we loop the submitButton click so we don't need to paste it many times

e2e-tests/constants/routes/insurance/policy.js
```diff
@@ -11,6 +11,9 @@ export const POLICY = {
   SINGLE_CONTRACT_POLICY_SAVE_AND_BACK: `${ROOT}/single-contract-policy/save-and-go-back`,
   SINGLE_CONTRACT_POLICY_CHANGE: `${ROOT}/single-contract-policy/change`,
   SINGLE_CONTRACT_POLICY_CHECK_AND_CHANGE: `${ROOT}/single-contract-policy/check-and-change`,
+  SINGLE_CONTRACT_POLICY_TOTAL_CONTRACT_VALUE: `${ROOT}/single-contract-policy/total-contract-value`,
+  SINGLE_CONTRACT_POLICY_TOTAL_CONTRACT_VALUE_CHANGE: `${ROOT}/single-contract-policy/total-contract-value/change`,
+  SINGLE_CONTRACT_POLICY_TOTAL_CONTRACT_VALUE_CHECK_AND_CHANGE: `${ROOT}/single-contract-policy/total-contract-value/check-and-change`,
```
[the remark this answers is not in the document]
ttbarnes: Yes I agree, i've added a TODO to split up this file and make the constants shorter. Can do after this PR please?

src/ui/server/controllers/insurance/policy/single-contract-policy/total-contract-value/index.ts
```diff
@@ -0,0 +1,183 @@
...
+      return res.redirect(`${INSURANCE_ROOT}/${referenceNumber}${CHECK_YOUR_ANSWERS}`);
+    }
+
+    if (isCheckAndChangeRoute(req.originalUrl)) {
+      return res.redirect(`${INSURANCE_ROOT}/${referenceNumber}${CHECK_AND_CHANGE_ROUTE}`);
+    }
+
+    return res.redirect(`${INSURANCE_ROOT}/${referenceNumber}${NAME_ON_POLICY}`);
+  } catch (err) {
+    console.error('Error updating application - policy - single contract policy %O', err);
```
Zainzzkk, line 179: should we add total contract value to the error?

e2e-tests/insurance/cypress/e2e/journeys/policy/broker/save-and-back.spec.js
ttbarnes: Great idea, I was thinking of creating something like `partiallySubmitXSection({ ... })` but a submission counter loop sounds loads better.

Can we do this as a piece of tech debt, we have lots of instances of submit buttons would need to go through a lot of files.

Zainzzkk (approved):

PR #1681, opened 2024-01-10
feat(EMS-2572): No PDF - Policy - Pre-credit period - Form submission, redirection flow

ttbarnes:
## Introduction :pencil2:
This PR updates the Policy form submission redirects to that the "Pre-credit period" form is part of the flow.

Note - there is no form validation yet, this is purely updating the flow.

- Update the `NAME_ON_POLICY` and `DIFFERENT_NAME_ON_POLICY` forms to redirect to the `PRE_CREDIT_PERIOD` page, instead  of `BROKER`.
- Add a`PRE_CREDIT_PERIOD` POST endpoint redirect to the `BROKER` page.
- Create new cypress command `completeAndSubmitPreCreditPeriodForm`. This will eventually be updated to actually complete the form.
- Update all relevant E2E tests and commands to submit the "Pre-credit period" form before submitting the "broker" form.

- Moved the `completeAndSubmitBrokerForm` cypress command into the policy directory.

### DIFF
--- a/e2e-tests/insurance/cypress/e2e/journeys/policy/complete-policy-as-multiple-contract-policy.spec.js
+++ b/e2e-tests/insurance/cypress/e2e/journeys/policy/complete-policy-as-multiple-contract-policy.spec.js
@@ -31,6 +31,7 @@ context('Insurance - Policy - Complete the entire section as a multiple contract
       cy.completeAndSubmitMultipleContractPolicyForm();
       cy.completeAndSubmitExportValueForm({ policyType });
       cy.completeAndSubmitNameOnPolicyForm({});
+      cy.completeAndSubmitPreCreditPeriodForm();
       cy.completeAndSubmitBrokerForm({});
 
       // go back to the all sections page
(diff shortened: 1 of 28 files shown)


e2e-tests/insurance/cypress/e2e/journeys/policy/complete-policy-as-multiple-contract-policy.spec.js
```diff
@@ -31,6 +31,7 @@ context('Insurance - Policy - Complete the entire section as a multiple contract
       cy.completeAndSubmitMultipleContractPolicyForm();
       cy.completeAndSubmitExportValueForm({ policyType });
       cy.completeAndSubmitNameOnPolicyForm({});
+      cy.completeAndSubmitPreCreditPeriodForm();
```
Zainzzkk, line 34: should these go in a function since repeated a few times?

ttbarnes: I don't think so, for now at least. This is to get to the middle of a flow. There will be probably 5-7 other forms after this, for the policy section.

We have this everywhere - tricky one. Could have e.g `submitUpToXForm`, but then there would be lots of these commands for each stage of a section and if one of the forms is moved, it would be very troublesome to update/maintain. Unless there are any other ideas, I think it's better to have as we have now.

Zainzzkk (approved):

PR #1696, opened 2024-01-11
fix(EMS-2556-2570): No PDF - Multiple contract policy - Total months of cover - decimal places

ttbarnes:
## Introduction :pencil2:
This PR fixes an issue where the "Total months of cover" field of a no PDF application, would allow decimal points and cause a "problem with service" issue.

- Add a validation rule so that "Total months of cover" cannot be submitted with decimal points.
- Add E2E test coverage.

- Bump depedencies.
- Minor destructuring improvements.

### DIFF
--- a/e2e-tests/insurance/cypress/e2e/journeys/policy/multiple-contract-policy/validation/multiple-contract-policy-validation-total-months-of-cover.spec.js
+++ b/e2e-tests/insurance/cypress/e2e/journeys/policy/multiple-contract-policy/validation/multiple-contract-policy-validation-total-months-of-cover.spec.js
@@ -89,6 +89,18 @@ context('Insurance - Policy - Multiple contract policy page - form validation - 
     });
   });
 
+  describe('when total months of cover contains a decimal', () => {
+    it('should render a validation error', () => {
+      cy.submitAndAssertFieldErrors(
+        field,
+        '7.5',
+        errorIndex,
+        expectedErrors,
+        CONTRACT_ERROR_MESSAGES[TOTAL_MONTHS_OF_COVER].INCORRECT_FORMAT,
+      );
+    });
+  });
+
   describe('when total months of cover is below the minimum', () => {
     it('should render a validation error', () => {
       cy.submitAndAssertFieldErrors(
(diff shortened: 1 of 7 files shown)


e2e-tests/insurance/cypress/e2e/journeys/policy/multiple-contract-policy/validation/multiple-contract-policy-validation-total-months-of-cover.spec.js
```diff
@@ -89,6 +89,18 @@ context('Insurance - Policy - Multiple contract policy page - form validation -
     });
   });
 
+  describe('when total months of cover contains a decimal', () => {
+    it('should render a validation error', () => {
+      cy.submitAndAssertFieldErrors(
+        field,
+        '7.5',
```
Zainzzkk, line 96: constant?

src/ui/server/controllers/insurance/policy/multiple-contract-policy/validation/rules/total-months-of-cover.test.ts
```diff
@@ -56,6 +54,18 @@ describe('controllers/insurance/policy/multiple-contract-policy/validation/rules
     });
   });
 
+  describe('when a value contains a decimal', () => {
+    it('should return a validation error', () => {
+      mockBody[FIELD_ID] = '7.5';
```
Zainzzkk, line 59: does 7.5 have to be a test mock?

e2e-tests/insurance/cypress/e2e/journeys/policy/multiple-contract-policy/validation/multiple-contract-policy-validation-total-months-of-cover.spec.js
ttbarnes: What kind of a constant? I don't think we shouldn't constants for plain numbers?

Zainzzkk: for example, I have no idea why it's 7.5.  but like const MONTH_WITH_DECIMAL would explain better?

ttbarnes: Is the test description not enough? "when total months of cover contains a decimal"

We have decimal tests like this everwhere with hard coded decimal values, e.g search for "contains a decimal"

Thing is, 7.5 is not month specific, it's just a number and so if we want this (and values with decimals) to be a constants, I suppose they should be generic, in which case how would a number constant be named?

Zainzzkk (approved):

PR #1698, opened 2024-01-11
feat(EMS-2558): No PDF - Policy - Pre-credit period - Save data

ttbarnes:
## Introduction :pencil2:
This PR adds data saving functionality for when the "Pre-credit period" form is submitted.

- Rename `PRE_CREDIT_PERIOD_DESCRIPTION` to `CREDIT_PERIOD_WITH_BUYER`. Now aligned with the existing database.
- Update the "Pre-credit period" POST controller to call an existing `mapAndSave` function.
- Update the "Pre-credit period" GET controller to map the application data and pass a `applicationAnswer` variable to the template.
- Add E2E test coverage.


src/ui/templates/partials/insurance/credit-period-with-buyer-description.njk
```diff
@@ -0,0 +1,21 @@
...
+{% endset %}
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-three-quarters-from-desktop">
+
+    {{ govukDetails({
+      summaryText: contentStrings.INTRO,
+      html: descriptionHtml,
+      attributes: {
+        "data-cy": "credit-period-with-buyer-description"
```
Zainzzkk, line 16: can the `credit-period-with-buyer` come from a fieldId?

ttbarnes: Updated - the only downside to this is that if this partial is used in another place (like we do with other partials), we may need to revert back as a relevant field ID maybe not be available.

Zainzzkk (approved):

PR #1713, opened 2024-01-16
fix(EMS-2538-2654-2658): No PDF - Get a quote currencies ordering, content updates

ttbarnes:
## Introduction :pencil2:
This PR updates the "currency as select options" mapping so that they are specifically ordered as per design, instead of being alphabetical.

Have also fixed/updated various copy to align with the No PDF design.

- Update content strings.
- Update mock and supported currencies fixtures and constants to export currency codes e.g `GBP`, `EUR` etc.
- Create a new UI helper function to map and specifically sort currencies for a select drop down, `mapAndSortCurrencies`.
- Update `mapCurrenciesAsSelectOptions` to consume new `mapAndSortCurrencies`.

- Rename some E2E test descriptions.
- Documentation improvements.

### DIFF
--- a/src/ui/server/constants/supported-currencies.ts
+++ b/src/ui/server/constants/supported-currencies.ts
@@ -1,1 +1,6 @@
 export const SUPPORTED_CURRENCIES = ['EUR', 'GBP', 'JPY', 'USD'];
+
+export const EUR = 'EUR';
+export const GBP = 'GBP';
+export const JPY = 'JPY';
+export const USD = 'USD';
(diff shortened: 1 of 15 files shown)


src/ui/server/constants/supported-currencies.ts
Zainzzkk, line 1: could we not get these from APIM?

ttbarnes: We do, but we need to have a specific hard coded list of supported currencies so that we can filter the APIM response. Also used for testing

Zainzzkk (approved):

PR #1761, opened 2024-01-23
feat(EMS-2535): No PDF - Policy - Change from single to multiple - enhancements

ttbarnes:
## Introduction :pencil2:
This PR updates the ability to change a "single contract policy" to a "multiple contract policy", so that when this occurs, a 3rd multiple policy related form is required to be completed before a user is redirected back to a "check or change your answers" page.

This can occur from:
- Policy - check your answers
- Insurance - check "policy" answers.

As part of this PR, have also update some policy data mapping so that if a policy is changed from single to multiple (or vice versa), single or multiple specific fields will be wiped.

- Add "check and change" UI routes for the `EXPORT_VALUE` (multiple policy specific field).
- Update "multiple contract policy" POST controller to redirect to `EXPORT_VALUE` if the route is a "change" or "check and change" route.
- Update the policy `mapSubmittedData` function to wipe single/multiple specific fields.
- Update `sanitiseFormField` to return a `null` if the provided value is `null`. Otherwise, it's not possible to nullify date values.
- Update E2E test coverage.

- Simplify some conditions and documentation in the "single contract policy" POST controller.

### DIFF
--- a/src/ui/server/controllers/insurance/policy/multiple-contract-policy/index.ts
+++ b/src/ui/server/controllers/insurance/policy/multiple-contract-policy/index.ts
@@ -18,7 +18,13 @@ import { Request, Response } from '../../../../../types';
 
 const {
   INSURANCE_ROOT,
-  POLICY: { MULTIPLE_CONTRACT_POLICY_SAVE_AND_BACK, MULTIPLE_CONTRACT_POLICY_EXPORT_VALUE, CHECK_YOUR_ANSWERS },
+  POLICY: {
+    MULTIPLE_CONTRACT_POLICY_SAVE_AND_BACK,
+    MULTIPLE_CONTRACT_POLICY_EXPORT_VALUE,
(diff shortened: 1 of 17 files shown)


src/ui/server/controllers/insurance/policy/multiple-contract-policy/index.ts
```diff
@@ -123,6 +132,8 @@ export const post = async (req: Request, res: Response) => {
     return res.redirect(PROBLEM_WITH_SERVICE);
   }
 
+  const { policy } = application;
```
Zainzzkk, line 135: do we need to have a check where if no policy, then it goes to PROBLEM_WITH_SERVICE

ttbarnes: No, `policy` object will always exist; It's created when the application is created.

If an application does not exist, server middleware will redirect the user before it reaches this controller 👍

PR #1779, opened 2024-01-24
fix(EMS-2610): No PDF - Policy - Single/multiple contracts - Currency prefixes

ttbarnes:
## Introduction :pencil2:
This PR fixes 2x issues with single and multiple contract policies where, after submitting a policy currency - the sequent monetary input fields would have a default GBP currency symbol prefix.

Now, if the submitted currency is "standard" currency, the appropriate currency symbol prefix will render Otherwise, if an alternative currency has been submitted, no prefix will be rendered.

- Update `completeAndSubmitMultipleContractPolicyForm` and `completeAndSubmitSingleContractPolicyForm` commands to accept an `alternativeCurrency` and `isoCode` params. If provided, an alternative currency will be submitted. These commands now need to be called with an object param structure and so, all appropriate tests have been updated.
- Add `SYMBOLS` constants/fixtures.
- Add E2E test coverage for:
  - Single and Multiple contract policy - when submitting an alternative currency, the sequent monetary input fields should NOT have a prefix.
  - Single and Multiple contract policy - when submitting a standard, non-GBP currency, the sequent monetary input fields should have a non-GBP prefix.
- Replace hard coded GBP currency symbols with `SYMBOLS` constant.
- Update the "accessible autocomplete" client side script to not consume the `alternativePolicyCurrency` ID.
- Update "total contract value" and "export value" controllers `pageVariables` to generate and return a `CURRENCY_PREFIX_SYMBOL`.
- Update single and multiple contract policy controllers to use `mapRadioAndSelectOptions` instead of `mapCurrenciesAsRadioOptions`.
- Create a new `monetaryValueInput` nunjucks components. Update all (3x) GOV inputs that have a prefix, to use this component.
- Replace `ALTERNATIVE_POLICY_CURRENCY_CODE` consumption with `ALTERNATIVE_CURRENCY_CODE`.

- Create new `autocompleteKeyboardInput` cypress command.
- Create new `clickAlternativeCurrencyRadioOption` cypress command.
- Minor improvemenst to the `keyboardInput` cypress command.
- Various destructuring improvements.
- Fix some typos.
- Remove some unnecessary unit tests for the multiple contract policy controller.


e2e-tests/commands/autocomplete-keyboard-input.js
```diff
@@ -0,0 +1,19 @@
+import { countryInput } from '../pages/shared';
+
+/**
+ * autocompleteKeyboardInput
+ * Type text into an autocomplete input.
+ * For example, country or currency autocomplete fields.
+ * @param {Function} fieldId: Autocomplete field ID
```
Zainzzkk, line 7: string?

e2e-tests/commands/shared-commands/click-events/click-alternative-currency-radio-option.js
Zainzzkk, line 1: could we not use assert-alternative-currency-form to do this?

e2e-tests/insurance/cypress/e2e/journeys/policy/multiple-contract-policy/export-value/multiple-contract-policy-export-value-non-gbp-currency.spec.js
Zainzzkk, line 1: could we combine this and the alternative-currency test since they're both short?

ttbarnes: I did think about this - think it's clearer as separate tests, rather than longer individual tests. 

These are also at the very start of the application so they are pretty quick to run.

Happy to change though but I think it's nicer as is?

e2e-tests/commands/shared-commands/click-events/click-alternative-currency-radio-option.js
ttbarnes: `assert-alternative-currency-form` has been updated to use this new command since it was repeated 3x times 😎

Zainzzkk (approved):

PR #1806, opened 2024-01-29
fix(EMS-2716-2717-2694): No PDF - Policy/Business - Name errors, check/change answer links, hint/legend issue

ttbarnes:
## Introduction :pencil2:

This PR fixes a few issues where:
- "Policy - name on policy" had some incorrect error messages.
- "Multiple policy - check/change your answers" had some incorrect URLs for the`TOTAL_SALES_TO_BUYER` and `MAXIMUM_BUYER_WILL_OWE` fields.
- "Business - turnover" was rendering a legend inside a legend. Also had a hint that is no longer required.

- Update error messages.
- Remove now unrequired Business - turnover legend.
- Re-enable and update some E2E tests for "Multiple policy - check/change your answers".
- Fix/update policy summary list field URLs.
- Update `fieldset-legend` nunjucks component to NOT render a legend if `isPageHeading` is false. Otherwise, a legend renders inside of a legend.

### DIFF
--- a/e2e-tests/insurance/cypress/e2e/journeys/check-your-answers/policy/change-your-answers/change-your-answers-policy-type-change-single-to-multiple-policy-type-summary-list.spec.js
+++ b/e2e-tests/insurance/cypress/e2e/journeys/check-your-answers/policy/change-your-answers/change-your-answers-policy-type-change-single-to-multiple-policy-type-summary-list.spec.js
@@ -11,7 +11,6 @@ const {
   POLICY: {
     TYPE_OF_POLICY_CHECK_AND_CHANGE,
     MULTIPLE_CONTRACT_POLICY_EXPORT_VALUE_CHECK_AND_CHANGE,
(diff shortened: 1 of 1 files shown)


src/ui/templates/components/fieldset-legend.njk
```diff
@@ -10,7 +10,7 @@
     </legend>
 
     {% else %}
-    <legend class="govuk-label--m" id="{{ id }}-legend" data-cy="{{ id }}-legend">{{ legendText }}</legend>
+    <span class="govuk-label--m" id="{{ id }}-legend" data-cy="{{ id }}-legend">{{ legendText }}</span>
```
Zainzzkk, line 13: could we pass the class from constants?

ttbarnes: Can we do as a tech improvement? We have over 20 instances of "govuk-label--X" that could come from constants.

Zainzzkk (approved):

PR #1827, opened 2024-01-31
chore(EMS-2726-2734): No PDF - Checked radio cypress commands

ttbarnes:
## Introduction :pencil2:
This PR introduces some new cypress commands to easily assert checked radios, therefore reducing the need to import specific radio input selectors in individual tests.

As the No PDF application grows, radio assertions will become even more frequent and so we should have commands to quickly/easily assert checked radios.

  - `assertRadioOptionIsChecked`
  - `assertYesRadioOptionIsChecked`
  - `assertNoRadioOptionIsChecked`
- Update all E2E tests.
- Add E2E test coverage for "alternative currency" pages having the GBP radio option checked by default (`assertGbpCurrencyCheckedByDefault`)


e2e-tests/commands/shared-commands/assertions/assert-yes-radio-option-is-checked.js
```diff
@@ -0,0 +1,11 @@
+import { yesRadioInput } from '../../../pages/shared';
+
+/**
+ * assertYesRadioOptionIsChecked
+ * Assert that a "yes" radio option is checked.
+ */
+const assertYesRadioOptionIsChecked = () => {
```
Zainzzkk, line 7: how will this work for more than one yes/no radio on the page

Zainzzkk: what if we pass the number, so by default it's input, but if you pass a number, then it does `yesRadioInput.eq(number)`? I think all yes no radios should come to this command rather than some going straight to `assertRadioOptionIsChecked` and some to here?
Opinion?

ttbarnes:
```js
cy.assertRadioOptionIsChecked(yesRadioInput().first());
cy.assertRadioOptionIsChecked(yesRadioInput().eq(1));
```

ttbarnes: @Zainzzkk Great idea, have added `index` and `eq()` logic to `cy.assertRadioOptionIsChecked()`.

Zainzzkk (approved):

PR #1833, opened 2024-01-31
fix(EMS-2709-2732): No PDF - Policy - Dynamic page titles with currency names

ttbarnes:
## Introduction :pencil2:
This PR fixes 2x issues in the "Policy" section of a no PDF application where, in the "export value" and "total contract value" pages, the page titles would render an "undefined" currency name, if an alternative currency name was selected in a previous part of the flow.

This was happening due to 2x issues:
- The policy `mapSubmittedData` function was incorrectly saving an alternative currency as the field ID instead of the actual field value.
- The GET controllers were consuming and passing only supported currencies to the page title genereation, therefore an alternative currency would never be found.

- Update "Business - Turnover - Alternative currency" copy (EMS-2732).
- Update GQL "Get APIM currencies" resolver to return an additional`allCurrencies` array.
- Extract "Business - Turnover - Alternative currency" `currency mapping logic into a new helper function, `mapCurrencyCodeFormData` since the logic is now required in 2 places
- Update 2x policy controllers to:
  - Call `pageVariables` with `allCurrencies`, instead of supported/mock currencies.
  - Include `ALTERNATIVE_CURRENCY_CODE` in `FIELD_IDS`, otherwise, submitted data mapping does not pick up the alternative currency code value.
- Add E2E test coverage.

- Create new cypress command to assert a page title with a currency name and optional question mark, `assertHeadingWithCurrencyName`.
- Update `policyCurrencyCode` to have a VarChar of 3 instead of 1000.
- Create a DRY UI test mock for an empty currencies response, `mockCurrenciesEmptyResponse`.
- Replace some hard coded UI unit test values with values from test mocks.


src/ui/server/controllers/insurance/business/turnover/map-submitted-data/index.ts
```diff
@@ -29,15 +31,13 @@ const mapSubmittedData = (formBody: RequestBody): object => {
...
-      populatedData[TURNOVER_CURRENCY_CODE] = populatedData[CURRENCY_CODE];
-    }
+  // map "currency code" and "alternative currency code" fields.
+  populatedData = mapCurrencyCodeFormData(populatedData);
 
+  // map the resulting "currency code" into a single "Turnover currency code" field
+  if (objectHasProperty(populatedData, CURRENCY_CODE)) {
+    populatedData[TURNOVER_CURRENCY_CODE] = populatedData[CURRENCY_CODE];
     delete populatedData[CURRENCY_CODE];
-    delete populatedData[ALTERNATIVE_CURRENCY_CODE];
```
Zainzzkk, line 40: is this not set if you choose an alternative currency?

src/ui/server/controllers/insurance/policy/map-submitted-data/policy/index.test.ts
```diff
@@ -161,7 +163,7 @@ describe('controllers/insurance/policy/map-submitted-data/policy', () => {
   describe(`when ${POLICY_TYPE} is multiple`, () => {
     it('should return an object with wiped single policy specific fields', () => {
       const mockBody = {
-        ...mockApplicationMultiplePolicy.policy,
+        [POLICY_TYPE]: mockApplicationMultiplePolicy.policy.policyType,
```
Zainzzkk, line 166: not a blocker, should we dereference? since used more than once

Zainzzkk (dismissed):

Zainzzkk (dismissed):

Zainzzkk (approved):

src/ui/server/controllers/insurance/business/turnover/map-submitted-data/index.ts
ttbarnes: Sorry only just seen this - no it's not, 

For all currency fields, it will always be set to `CURRENCY_CODE`.

So if the "alternative" option is selected, `CURRENCY_CODE` will equal "alternative".

the "alternative" field is not in the actual data it's purely for the UI, so we need to delete it (moved to a new helper function in this PR)

PR #1847, opened 2024-02-02
feat(EMS-2722): No PDF - Single/Multiple contract policy - Currency validation updates. Refactor policy currency ID

ttbarnes:
## Introduction :pencil2:
This PR updates both the Single and Multiple contract policy forms to render an error the "alternative currency" option is selected, but no alternative currency is chosen.

This also refactors the policy currency ID. Previously, `POLICY_CURRENCY_CODE` was used for field rendering, validation and data saving. Now, this is only used for data saving and application mapping. All other policy currency instances use `CURRENCY_CODE` - the same as all other currency areas.

Have also replaced the `ALTERNATIVE_POLICY_CURRENCY_CODE` field ID. with `ALTERNATIVE_CURRENCY_CODE`.

- Update some E2E currency assertion functions to:
  - Have object structured params.
  - Accept a `completeNonCurrencyFields` argument, If passed, this function will be called immediately.
  - Accept an `errorIndex` param.
  - This is required so that if a form has currency fields and other non-currency related fields, the other non-currency fields can be filled in before asserting the currency fields.
- Create `completeSingleContractPolicyForm` cypress command.
- Create `completeMultipleContractPolicyForm` cypress command.
- Update policy `mapSubmittedData` and `mapApplicationToFormFields` functions to map `POLICY_CURRENCY_CODE` to `CURRENCY_CODE`.
- Create new `policyAlternativeCurrencyCodeRule` shared validation rules.
- Add `alternativeCurrencyRule` validation rule to Single and Multiple policy validaiton rules.

- Replace `POLICY_CURRENCY_CODE` instances with `CURRENCY_CODE`.
- Update Single and Multiple contract policy E2E tests to use `assertCurrencyFormFields`.
- Remove now unused `checkPolicyCurrencyCodeInput` E2E helper.
- Remove now unused `ALTERNATIVE_POLICY_CURRENCY_CODE` field ID.
- Add some missing documentation.
- Simplify some alternative currency code validation unit tests.
- Remove some unnecessary policy E2E tests for validating currency code fields.


e2e-tests/constants/field-ids/insurance/policy/index.js
```diff
@@ -9,7 +9,6 @@ export const SHARED_CONTRACT_POLICY = {
   REQUESTED_START_DATE_MONTH: `${REQUESTED_START_DATE}-month`,
   REQUESTED_START_DATE_YEAR: `${REQUESTED_START_DATE}-year`,
   POLICY_CURRENCY_CODE: 'policyCurrencyCode',
```
Zainzzkk, line 11: has this one been removed?

src/ui/server/controllers/insurance/policy/single-contract-policy/index.test.ts
```diff
@@ -103,9 +103,9 @@ describe('controllers/insurance/policy/single-contract-policy', () => {
             ID: CONTRACT_COMPLETION_DATE,
             ...FIELDS.CONTRACT_POLICY.SINGLE[CONTRACT_COMPLETION_DATE],
           },
-          POLICY_CURRENCY_CODE: {
-            ID: POLICY_CURRENCY_CODE,
-            ...FIELDS.CONTRACT_POLICY[POLICY_CURRENCY_CODE],
+          CURRENCY_CODE: {
```
Zainzzkk, line 106: we are using this and then mapping to `POLICY_CURRENCY_CODE`.  Is it not worth using `POLICY_CURRENCY_CODE `here instead?

src/ui/server/helpers/mappings/map-application-to-form-fields/index.ts
```diff
@@ -74,6 +76,13 @@ const mapApplicationToFormFields = (application?: Application): Application | ob
       };
     }
 
+    if (application?.policy?.[POLICY_CURRENCY_CODE]) {
```
Zainzzkk, line 79: do we need this line as won't it always be set to a default value of GBP so will always exist?

src/ui/server/controllers/insurance/policy/single-contract-policy/index.test.ts
ttbarnes: But then it would be unaligned with all currency form fields. I think all currency form fields should use the same ID, `CURRENCY_CODE`?

Zainzzkk (dismissed):

Zainzzkk (approved):

PR #1879, opened 2024-02-08
chore(EMS-2825): No PDF - Task status cypress commands

ttbarnes:
## Introduction :pencil2:
This PR updates all instances of E2E task assertions to use new, DRY cypress commands.

Therefore reducing the need to import partials, `TASKS` strings and define local variables in individual tests.

- Create new generic cypress commands for possible task status:
  - `checkTaskStatusCannotStart`
  - `checkTaskStatusNotStartedYet`
  - `checkTaskStatusInProgress`
  - `checkTaskStatusCompleted`
- Create a cypress command for each possible task status, for every application section. For example, the "business" section has the following commands available:
  - `checkTaskBusinessStatusIsNotStartedYet`
  - `checkTaskBusinessStatusIsComplete`
  - `checkTaskBusinessStatusIsInProgress`
- Update every E2E test to call the new commands, instead of importing partials and/or content strings.

- Improve assertions commands structure, imports/exports.


e2e-tests/insurance/cypress/e2e/journeys/declarations/how-your-data-will-be-used/save-and-back.spec.js
```diff
@@ -66,8 +63,8 @@ context('Insurance - Declarations - How your data will be used page - Save and g
       cy.assertUrl(expected);
     });
 
-    it(`should retain the status of task 'declarations' as '${IN_PROGRESS}'`, () => {
-      cy.checkTaskStatus(task, IN_PROGRESS);
+    it('should retain the status of task `declarations` as `in progress`', () => {
```
Zainzzkk, line 66: should the name of the task status still not come from the constant?

ttbarnes: For task checking I don't think it should, they're very unlikely to change and if we use constants for the task names in test descriptions, each test with task assertions has to import and destructure them

What do you think?

Zainzzkk (approved):

PR #1880, opened 2024-02-09
feat(EMS-2514): No PDF - Your buyer - Buyer financial information - Data saving and save and back

Zainzzkk:
## Introduction :pencil2:
This PR adds data saving and save and back for the buyer financial information page

* Added data saving to controller
* Added save and back controller
* Updated routes
* Updated tests

- Fixed `completeAndSubmitBuyerFinancialInformationForm` which had wrong form to complete


src/ui/server/controllers/insurance/your-buyer/buyer-financial-information/index.ts
```diff
@@ -133,6 +135,13 @@ export const post = async (req: Request, res: Response) => {
       });
     }
 
+    // if no errors, then runs save api call
```
ttbarnes, line 138: Not a blocker but FYI the try/catch on such controllers only needs to start from just before the `mapAndSave` await call. I.e the try doesn't need to be at the top of the controller

ttbarnes (approved):

ttbarnes (approved):

src/ui/server/controllers/insurance/your-buyer/buyer-financial-information/index.ts
Zainzzkk: will do in next PR

PR #1925, opened 2024-02-15
feat(EMS-2826): No PDF - Nominated loss payee - Setup DB fields/schema

ttbarnes:
## Introduction :pencil2:
This PR sets up the initial DB fields and GQL schema for the "Nominated loss payee" feature of a No PDF application.

- Create `NominatedLossPayee` table with some fields and relationship to:
  - `Application`
  - `LossPayeeFinancialInternational`
  - `LossPayeeFinancialUk`
- Create `LossPayeeFinancialInternational` table with some fields and relationship to `NominatedLossPayee`.
- Create `LossPayeeFinancialUk` table with some fields and relationship to `NominatedLossPayee`.
- Create new types.
- Update DB dump.
- Update `createAnApplication` GQL mutation to generate a nominated loss payee.
- Create new helper functions:
  - `createANominatedLossPayee`
  - `createALossPayeeFinancialInternational`
  - `createALossPayeeFinancialUk`
- Update the UI's "Get application" GQL query to include some simple `nominatedLossPayee` fields.

- Re-order some assertions, queries and types to be ordered alphabetically.


src/ui/server/graphql/queries/application.ts
```diff
@@ -38,6 +38,11 @@ const applicationQuery = gql`
           }
           validExporterLocation
         }
+        nominatedLossPayee {
+          isAppointing
+          locatedInUk
+          name
+        }
```
Zainzzkk, line 45: do we need to also query the UK and international tables?

ttbarnes: Eventually yes, but we don't need any of the UI GQL queries right now. Just put this in as a place holder as it'll be the first part of the whole section that will need to be queried.

Also need to think about how the sensitive financial data will be returned after encryption.

Zainzzkk (approved):

PR #1934, opened 2024-02-15
feat(EMS-2854): No PDF - Policy - Confirm broker address - Multi line support

ttbarnes:
## Introduction :pencil2:
This PR updates the "Confirm broker address" functionality of a no PDF application so that if a "full broker address" (textarea) is submitted with line breaks, it will be played back to the user exactly as it was entered.

- Update `completeBrokerDetailsForm` and `completeAndSubmitBrokerDetailsForm` cypress commands to accept custom values.
- Create `FULL_ADDRESS` E2E constants/examples.
- Add E2E test for submitting and playing back an address with line breaks in the "Confirm broker address" page.
- Update `mapApplicationToFormFields` to map `FULL_ADDRESS`.
- Update "Confirm broker address" GET controller to consume/pass `mapApplicationToFormFields`.
- Update "Confirm broker address" nunjucks component to render the inset text value as raw (safe) HTML.

### DIFF
--- a/e2e-tests/commands/insurance/complete-and-submit-broker-details-form.js
+++ b/e2e-tests/commands/insurance/complete-and-submit-broker-details-form.js
@@ -1,9 +1,12 @@
 /**
  * completeAndSubmitBrokerDetailsForm
  * Complete and submit "broker details" form
+ * @param {String} name: Broker name
+ * @param {String} email: Broker email
+ * @param {String} fullAddress: Broker's full address
  */
-const completeAndSubmitBrokerDetailsForm = () => {
-  cy.completeBrokerDetailsForm();
+const completeAndSubmitBrokerDetailsForm = ({ name, email, fullAddress }) => {
+  cy.completeBrokerDetailsForm({ name, email, fullAddress });
 
   cy.clickSubmitButton();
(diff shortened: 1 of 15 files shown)


e2e-tests/commands/insurance/complete-and-submit-broker-details-form.js
```diff
@@ -1,9 +1,12 @@
 /**
  * completeAndSubmitBrokerDetailsForm
  * Complete and submit "broker details" form
+ * @param {String} name: Broker name
+ * @param {String} email: Broker email
+ * @param {String} fullAddress: Broker's full address
  */
-const completeAndSubmitBrokerDetailsForm = () => {
-  cy.completeBrokerDetailsForm();
+const completeAndSubmitBrokerDetailsForm = ({ name, email, fullAddress }) => {
```
Zainzzkk, line 8: do we need default values here as well?

ttbarnes: No, it'll always go to `completeBrokerDetailsForm` which has default values. Having them here would just be duplicated. We have this pattern in several places.

PR #1935, opened 2024-02-15
chore(EMS-2883): E2E - Declare and use baseUrl

Zainzzkk:
## Introduction :pencil2:
This PR standardises use of baseUrl in E2E tests.

* Declared `const baseUrl` in all files which did not use it.
* Change to `${baseUrl}` in urls.


ttbarnes (approved):
Great work 😍

PR #1938, opened 2024-02-16
fix(EMS-2881): No PDF - Your Buyer - Change your answers - Alternative currency

Zainzzkk:
## Introduction :pencil2:
This PR fixes not being redirected to check your answers after submitting with an alternative currency

* Added flag in get controller in trading-history for if change or check-and-change route
* Constructed url for alternative currency in nunjucks template if url is change/check-and-change
* Added e2e tests


src/ui/templates/partials/insurance/buyer-outstanding-payment-conditional-yes-html.njk
```diff
@@ -1,5 +1,19 @@
...
+{#
+  sets url for alternative currency link
+  if isChange - then adds /change to url
+  if isCheckAndChange - then adds /check-and-change to url
+  else sets it to PROVIDE_ALTERNATIVE_CURRENCY_URL
+#}
+{% if isChange === true %}
+  {% set alternativeCurrencyUrl = PROVIDE_ALTERNATIVE_CURRENCY_URL + "/change"%}
+{% elif isCheckAndChange === true %}
+  {% set alternativeCurrencyUrl = PROVIDE_ALTERNATIVE_CURRENCY_URL + "/check-and-change"%}
```
ttbarnes, line 12: Could we pass the URLs via the controller so the template just has e.g `CHANGE_URL`, `CHECK_AND_CHANGE_URL`? Like we do for the "save and back" URL

e2e-tests/insurance/cypress/e2e/journeys/your-buyer/change-your-answers/change-your-answers-alternative-currency.spec.js
```diff
@@ -0,0 +1,149 @@
...
+
+  beforeEach(() => {
+    cy.saveSession();
+  });
+
+  after(() => {
+    cy.deleteApplication(referenceNumber);
+  });
+
+  describe(`changing ${CURRENCY_CODE} to Euros`, () => {
```
ttbarnes, line 58: Could we use `SYMBOLS.EUR` instead of `EUR_CURRENCY_CODE` and "Euros" text?

src/ui/server/controllers/insurance/your-buyer/trading-history/index.test.ts
```diff
@@ -151,11 +151,59 @@ describe('controllers/insurance/your-buyer/trading-history', () => {
         userName: getUserNameFromSession(req.session.user),
         ...pageVariables(mockApplication.referenceNumber, mockCurrencies, currencyValue),
         application: mapApplicationToFormFields(mockApplication),
+        isChange: undefined,
+        isCheckAndChange: undefined,
```
ttbarnes, line 155: Could this just not be provided instead of having `undefined` values?

ttbarnes (approved):

PR #1949, opened 2024-02-19
feat(EMS-2721-2480): No PDF - Single/Multiple policy - "Other currency" validation, status tag design update

ttbarnes:
## Introduction :pencil2:
This PR ensures that the "single" and "multiple" contract policy form validation renders a validation error when the "other currency" radio option is selected, but no other currency is selected.

This functionality is already in place, however we did not have an E2E test for this.

Have also updated the status tag styles to match `no-pdf` design.

- Update "single" and "multiple" contract policy E2E tests to consume the existing `selectAltRadioButNoAltCurrency` assertion function.
- Update `selectAltRadioButNoAltCurrency` to have object structured params.
- Update "status tag" nunjucks component.
- Create new `checkClassName` cypress command.
- Update all relevant "check task status" commands to check the class name.

- Move `ATTRIBUTES` constants into their own file, outside of `TEMPLATES`, since they are no longer template specific. Add to E2E constants.
- Create new `checkTaskStatusSubmitted` cypress command.
- Update `task-status` nunjucks component to consume tag classes from constants.
- Fix some typos.


src/ui/templates/components/status-tag.njk
```diff
@@ -17,11 +17,11 @@
   {% endif %}
 
   {% if status === "Completed" %}
-    {% set classes = "govuk-tag--green" %}
+    {% set classes = "govuk-tag--turquoise" %}
```
abhi-markan, line 20: Is there a test which reflects this change?

abhi-markan (changes requested):

Zainzzkk (approved):

src/ui/templates/components/status-tag.njk
Zainzzkk: could do an e2e with for class test or Caroline to test it

ttbarnes: Thanks guys, good shout.

Typically we don't have E2E tests for "visual CSS things", since 99.9% of it is handled by the GOV design library and we don't have much/any customisation of the visual aspects.

However for status tags, I think it's important to have this.

I've got a new cypress command working for 1x scenario. `cy.checkClassName()`

Need to create or update dashboard and task lists E2E tests to consume this new command 👍

Zainzzkk (dismissed):

abhi-markan (approved):

src/ui/templates/components/status-tag.njk
ttbarnes: Have updated RE the above with the latest commits.

Zainzzkk (approved):

PR #1950, opened 2024-02-19
chore(EMS-2908): No PDF - Rename countryInput cypress selector

ttbarnes:
## Introduction :pencil2:
This PR renames a shared cypress selector to be generic.

Since changing currency fields to be an autocomplete field, all tests were using the `countryInput` selector.

Since the selector is used for multiple country and currency fields, it should have a generic name.

- Rename `countryInput` to `autoCompleteField`.
- Simplify `autoCompleteField` structure to it can be called immediately, instead of having to call `.field`.

- Align some E2E test descriptions for "save and back" assertions.


e2e-tests/pages/shared/autoCompleteField.js
```diff
@@ -0,0 +1,17 @@
+import fieldSelector from './field';
+
+/**
+ * autoCompleteField
+ * "Autocomplete" cypress selector.
+ * This field is used for any autocomplete fields, e.g country and currency fields.
+ * @param {String} fieldId: Field ID
+ * @returns
```
abhi-markan, line 8: * @returns {Object} autocompleted `input`, `results` and `noResults` Cypress selectors?
* Do we require `noResults` here given it is not being constructed with `fieldId`?

abhi-markan (changes requested):
Minor comment to address please.

e2e-tests/commands/insurance/assert-currency-form-fields.js
```diff
@@ -85,24 +85,24 @@ const assertCurrencyFormFields = ({
...
     cy.clickAlternativeCurrencyRadioOption();
 
-    checkAutocompleteInput.rendersNoResultsMessage(countryInput.field(alternativeCurrencyFieldId), 'not a currency');
+    checkAutocompleteInput.rendersNoResultsMessage(autoCompleteField(alternativeCurrencyFieldId), 'not a currency');
     // should not render radio values in alternate currency input
-    checkAutocompleteInput.rendersNoResultsMessage(countryInput.field(alternativeCurrencyFieldId), GBP.isoCode);
-    checkAutocompleteInput.rendersNoResultsMessage(countryInput.field(alternativeCurrencyFieldId), USD.isoCode);
-    checkAutocompleteInput.rendersNoResultsMessage(countryInput.field(alternativeCurrencyFieldId), JPY.isoCode);
-    checkAutocompleteInput.rendersNoResultsMessage(countryInput.field(alternativeCurrencyFieldId), EUR.isoCode);
+    checkAutocompleteInput.rendersNoResultsMessage(autoCompleteField(alternativeCurrencyFieldId), GBP.isoCode);
```
Zainzzkk, line 96: is it worth doing `const autoCompleteField= autoCompleteField(alternativeCurrencyFieldId)` so less complicated line?

ttbarnes: Good shout, i've made some updates so it's use a `field` variable.

Also noticed that there is some dupication for these assertions and another assertion file, made a note to address this later.

Zainzzkk (approved):

ttbarnes:
@abhi-markan Can you please re-review

abhi-markan (approved):

PR #2034, opened 2024-03-06
feat(EMS-2830): No PDF - Policy - Check your answers - Contact Email

Zainzzkk:
## Introduction :pencil2:
This PR adds contact email row if same name on policy without a change link

* Added logic to summary list helper
* Updated tests
* Updated e2e tests

### DIFF
--- a/e2e-tests/commands/shared-commands/assertions/assert-summary-list-row-value.js
+++ b/e2e-tests/commands/shared-commands/assertions/assert-summary-list-row-value.js
@@ -4,8 +4,8 @@
  * E.g assertSummaryListRowValue({ rowA: {...}, rowB: {...} }, 'startDate', '01/02/2023')
  * @param {Object} Summary list cypress selectors from pages/partials.
  * @param {String} Field ID in the summary list to assert
- * @param {Number} index eg the index of identical rows
  * @param {String} Expected value
+ * @param {Number} index eg the index of identical rows
  */
 const assertSummaryListRowValue = (summaryList, fieldId, expected, index) => {
   if (index) {
--- a/e2e-tests/commands/shared-commands/assertions/assert-summary-list-row-value.js
+++ b/e2e-tests/commands/shared-commands/assertions/assert-summary-list-row-value.js
@@ -4,13 +4,13 @@
  * E.g assertSummaryListRowValue({ rowA: {...}, rowB: {...} }, 'startDate', '01/02/2023')
  * @param {Object} Summary list cypress selectors from pages/partials.
  * @param {String} Field ID in the summary list to assert
- * @param {Number} fieldNumber eg the first or second answer of identical field_id'd rows
+ * @param {Number} index eg the index of identical rows
  * @param {String} Expected value
  */
-const assertSummaryListRowValue = (summaryList, fieldId, expected, fieldNumber) => {
-  if (fieldNumber) {
+const assertSummaryListRowValue = (summaryList, fieldId, expected, index) => {
+  if (index) {
     cy.checkText(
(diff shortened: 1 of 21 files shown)


e2e-tests/commands/shared-commands/assertions/assert-summary-list-row-value.js
```diff
@@ -4,13 +4,21 @@
  * E.g assertSummaryListRowValue({ rowA: {...}, rowB: {...} }, 'startDate', '01/02/2023')
  * @param {Object} Summary list cypress selectors from pages/partials.
  * @param {String} Field ID in the summary list to assert
+ * @param {Number} fieldNumber eg the first or second answer of identical field_id'd rows
```
ttbarnes, line 7: `field_id'd` ?

ttbarnes: What's the need for this, why would there be multiple identital field IDs rows, shouldn't summary lists have only have unique IDs?

e2e-tests/insurance/cypress/e2e/journeys/check-your-answers/policy/change-your-answers/change-your-answers-change-broker.spec.js
```diff
@@ -196,7 +196,8 @@ context('Insurance - Change your answers - Policy - Broker - Summary list', () =
 
       it('should render the new answer and retain a `completed` status tag', () => {
         fieldVariables.newValue = fieldVariables.newValueInput;
-        cy.checkChangeAnswerRendered(fieldVariables);
+        // 1 as is the second email field on the page
+        cy.checkChangeAnswerRendered(fieldVariables, 1);
```
ttbarnes, line 200: Are the 2 email fields - 1 for broker and 1 for policy contact using the same `EMAIL` ID?

I'm wondering if it'd be better to select the the fields via the child summary list (for X section), rather than "second email field in the whole page".

Otherwise not sure if should be tech debt, but `checkChangeAnswerRendered` would be nicer in an object structure so `1` has more context and then no need for a comment

Zainzzkk: The functions not able to do this yet I think and it would be a whole other PR to get it working in that way.  I think it's better to do it in another PR as Caroline is blocked on testing by these PRs

ttbarnes: It seems that all `checkChangeAnswerRendered` commands (~20) could be updated very easily to support this with object structure. 5min job?

Any existing commands e.g `cy.checkChangeAnswerRendered(fieldVariables);` would still work with new structure. just in this new pattern it would be nicer and not introducing tech debt or additional code comments 

e.g
```js
cy.checkChangeAnswerRendered({ ...fieldVariables, index: 1 });
```

e2e-tests/commands/shared-commands/assertions/assert-summary-list-row-value.js
```diff
@@ -4,13 +4,21 @@
  * E.g assertSummaryListRowValue({ rowA: {...}, rowB: {...} }, 'startDate', '01/02/2023')
  * @param {Object} Summary list cypress selectors from pages/partials.
  * @param {String} Field ID in the summary list to assert
+ * @param {Number} index eg the index of identical rows
```
ttbarnes, line 7: Documentation is the wrong way around

Can fix in next PR

ttbarnes (approved):

PR #2045, opened 2024-03-08
feat(EMS-2768): No PDF - Policy - Loss Payee Bank Details - Page setup

Zainzzkk:
## Introduction :pencil2:
This PR introduces the loss payee bank details page

* Added routes and constants for loss payee bank details
* Added controller
* Added tests and e2e tests
* Added commands and mocks


e2e-tests/content-strings/fields/insurance/policy/index.js
```diff
@@ -276,4 +277,18 @@ export const POLICY_FIELDS = {
       },
     },
   },
+  BANK_DETAILS: {
```
ttbarnes, line 280: Can we call this and all respective directories, files, ID etc, `LOSS_PAYEE_BANK_DETAILS`, this is specific to a loss payee.

I'm wondering if we should call this and all respective directories, files etc, `LOSS_PAYEE_UK_BANK_DETAILS` so it's clear, otherwise we have a "hidden assumption" that `X_BANK_DETAILS` is UK - wouldn't be obvious for a new developer. Thoughts?

src/ui/templates/insurance/policy/bank-details.njk
```diff
@@ -0,0 +1,151 @@
...
+          id: FIELDS.ACCOUNT_NUMBER.ID,
+          name: FIELDS.ACCOUNT_NUMBER.ID,
+          value: submittedValues[FIELDS.ACCOUNT_NUMBER.ID] or application.policy[FIELDS.ACCOUNT_NUMBER.ID],
+          errorMessage: validationErrors.errorList[FIELDS.ACCOUNT_NUMBER.ID] and {
+            text: validationErrors.errorList[FIELDS.ACCOUNT_NUMBER.ID].text,
+            attributes: {
+              "data-cy": FIELDS.ACCOUNT_NUMBER.ID + "-error-message"
+            }
+          }
+        }) }}
```
ttbarnes, line 114: Have been wanting to create a new `textInput` type nunjucks component because all instances are almost exactly the same except for the field ID and e.g custom classes.

Not a blocker, I think we should create a ticket that blocks the page setup for international bank details. Until we have this component, the more pages that are built, the longer the task will take

e2e-tests/fixtures/application.js
```diff
@@ -139,6 +141,9 @@ const application = {
     [MAXIMUM_BUYER_WILL_OWE]: '500',
     [NEED_PRE_CREDIT_PERIOD]: 'false',
     [CREDIT_PERIOD_WITH_BUYER]: 'Mock description',
+    [ACCOUNT_NUMBER]: '123456',
+    [SORT_CODE]: '123456',
+    [BANK_ADDRESS]: 'Mock address',
```
ttbarnes, line 146: Can we wrap this in an `LOSS_PAYEE_FINANCIAL_UK` object

Also can we update the account number with maximium characters and have "Mock bank UK address" instead of "Mock address"? We have "Mock address" in a few places

ttbarnes (approved):

PR #2075, opened 2024-03-14
feat(EMS-2769-2770-2771): No PDF - Policy - Loss payee - UK financial details - Validation

Zainzzkk:
## Introduction :pencil2:
This PR adds validation to the loss payee financial page

* Added validation for the three fields
* Added new helpers for minimum maximum length for whole numbers
* Added helper for hyphen or space validation
* Added tests

* Made REGEX constants
* Changed instances of UK to Uk


e2e-tests/content-strings/fields/insurance/policy/index.js
```diff
@@ -280,14 +280,18 @@ export const POLICY_FIELDS = {
...
       LABEL: 'Account number',
       HINT: 'Must be between 6 and 8 digits long',
+      MINIMUM: 6,
+      MAXIMUM: 8,
     },
   },
   FINANCIAL_ADDRESS: {
     LABEL: 'Bank address',
-    MAXIMUM: 300,
+    MAXIMUM: 500,
```
ttbarnes, line 295: We can remove this, if the full address validation is using `fullAddress` from `shared-validation`, it's handled in there

e2e-tests/content-strings/fields/insurance/policy/index.js
```diff
@@ -280,14 +280,18 @@ export const POLICY_FIELDS = {
...
       LABEL: 'Sort code',
       HINT: 'Must be 6 digits long',
+      MINIMUM: 6,
+      MAXIMUM: 6,
     },
     [ACCOUNT_NUMBER]: {
       LABEL: 'Account number',
       HINT: 'Must be between 6 and 8 digits long',
+      MINIMUM: 6,
+      MAXIMUM: 8,
```
ttbarnes, line 290: Should we move these (and for `SORT_CODE`), into `constants/validation`? Cleaner than having in content strings IMO, since they're also used in validation

e2e-tests/insurance/cypress/e2e/journeys/policy/loss-payee-financial-uk/validation/sort-code-validation.spec.js
```diff
@@ -0,0 +1,136 @@
...
+
+    cy.submitAndAssertFieldErrors(field, value, errorIndex, numberOfExpectedErrors, ERROR.INCORRECT_FORMAT);
+  });
+
+  it(`should render validation errors when ${FIELD_ID} has a special character`, () => {
+    const value = '11-22-3!';
+
+    cy.submitAndAssertFieldErrors(field, value, errorIndex, numberOfExpectedErrors, ERROR.INCORRECT_FORMAT);
+  });
+
```
ttbarnes, line 102: Can we add test for "Has a letter and a special character"?

src/ui/server/constants/regex.ts
```diff
@@ -0,0 +1,5 @@
+export const REGEX = {
+  ALPHA_CHARACTERS: /^[a-zA-Z ]*$/,
```
ttbarnes, line 2: This is actually alpha characters and an empty space, same with the below regexes

Shall we add some comments to each regex? I think quite helpful, e.g

```js
/**
  * ALPHA_CHARACTERS:
  * Regex that allows only:
  * - a-z or A-Z characters.
  * - an empty space.
 */
  ALPHA_CHARACTERS: /^[a-zA-Z ]*$/,
```

src/ui/server/controllers/insurance/policy/loss-payee-financial-uk/validation/rules/financial-address.ts
```diff
@@ -0,0 +1,24 @@
...
+export const MAXIMUM = Number(POLICY_FIELDS.FINANCIAL_ADDRESS.MAXIMUM);
+
+/**
+ * financialAddressRule
+ * validates financial address field
+ * checks if response has been provided and not above maximum
+ * Returns providedAndMaxLength if there are any errors.
+ * @param {RequestBody} formBody
+ * @param {Object} errors
+ * @returns {Object} errors
```
ttbarnes, line 20: Same as above

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      const result = numberHyphenSpacesOnlyValidation(mockBody, FIELD_ID, mockErrorMessagesObject, mockErrors, MINIMUM, MAXIMUM);
+
+      expect(result).toEqual(mockErrors);
+    });
+  });
+
+  describe('when the provided value does not have hyphens or spaces', () => {
+    it('should return provided errors object', () => {
+      mockBody[FIELD_ID] = '112233';
```
ttbarnes, line 87: Should this be valid?

Zainzzkk: yes, it can either be a number, number with hyphen or number with spces

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  // check if the field is below the minimum
+  if (minimum && replaced.length < minimum) {
+    return generateValidationErrors(fieldId, errorMessage.BELOW_MINIMUM, errors);
+  }
+
+  // check if the field is above the maximum
+  if (maximum && replaced.length > maximum) {
+    return generateValidationErrors(fieldId, errorMessage.ABOVE_MAXIMUM, errors);
+  }
```
ttbarnes, line 62: Could we use these helper functions for the 2x conditions above?
```js
isNumberBelowMinimum
isNumberAboveMaximum
```

src/ui/server/shared-validation/whole-number-minimum-maximum-length/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  // check if the field is below the minimum length
+  if (minimum && numberWithoutCommas.length < minimum) {
+    return generateValidationErrors(fieldId, errorMessage.BELOW_MINIMUM, errors);
+  }
+
+  // check if the field is above the maximum length
+  if (maximum && numberWithoutCommas.length > maximum) {
+    return generateValidationErrors(fieldId, errorMessage.ABOVE_MAXIMUM, errors);
+  }
+
```
ttbarnes, line 47: Same as above, maybe can use `isNumberBelowMinimum` and `isNumberAboveMaximum`?

Should the minimum and maximum variables here be required? I can't see why they'd be optional

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.test.ts
ttbarnes: It's returning validation errors if pure numbers?

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.ts
Zainzzkk: I am not sure as we are checking the legnth here while those check the number provided is above or below minimum

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.test.ts
Zainzzkk: no, 12345, 12 34 5 and 1-2-3-4-5 are all valid

ttbarnes: After looking through the code, I understand now.

It wasn't clear to me that this test in particular was returning the provided errors object (which has no errors).

Can we update the describe blocks so that the valid and invalid scenarios are in a parent describe block?

src/ui/server/controllers/insurance/policy/loss-payee-financial-uk/validation/rules/financial-address.ts
```diff
@@ -0,0 +1,24 @@
...
+ * financialAddressRule
+ * validates financial address field
+ * checks if response has been provided and not above maximum
+ * Returns providedAndMaxLength if there are any errors.
+ * @param {RequestBody} formBody
+ * @param {Object} errors
+ * @returns {Object} providedAndMaxLength errors
+ */
+const financialAddressRule = (formBody: RequestBody, errors: object) => providedAndMaxLength(formBody, FIELD_ID, ERROR_MESSAGES_OBJECT, errors, MAXIMUM);
+
```
ttbarnes, line 23: This function can just call `fullAddress` (which calls `providedAndMaxLength` with the `MAXIMUM`)

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  const fieldValue = formBody[fieldId];
+
+  const joiString = Joi.string();
+
+  /**
+   * Regex that allows only
+   * - numbers
+   * - hyphens
+   * - spaces
+   */
```
ttbarnes, line 41: I think this comment is redundant since we have constants with comments 😎

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  // replaces and removes hyphens and spaces
+  const replaced = fieldValue.replaceAll(REGEX.SPACE_AND_HYPHEN, '');
+
+  // check if the field is below the minimum
+  if (minimum && isNumberBelowMinimum(replaced.length, minimum)) {
+    return generateValidationErrors(fieldId, errorMessage.BELOW_MINIMUM, errors);
+  }
+
+  // check if the field is above the maximum
+  if (maximum && isNumberAboveMaximum(replaced.length, maximum)) {
```
ttbarnes, line 61: Seems we can remove the `minimum &&` and `maximum &&` checks now because they're required params in this function?

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        const result = numberHyphenSpacesOnlyValidation(mockBody, FIELD_ID, mockErrorMessagesObject, mockErrors, MINIMUM, MAXIMUM);
+
+        const expected = generateValidationErrors(FIELD_ID, mockErrorMessagesObject.ABOVE_MAXIMUM, mockErrors);
+
+        expect(result).toEqual(expected);
+      });
+    });
+  });
+
+  describe('valid inputs', () => {
```
ttbarnes, line 77:
```js
describe('invalid values', () => {
describe('valid inputs', () => {
```

Feels a bit strange (values vs inputs), how about something like this?

```js
describe('invalid field value', () => {
describe('valid field value', () => {
```

src/ui/server/shared-validation/number-hyphen-and-spaces-only/index.ts
ttbarnes: Same in the `wholeNumberMinimumMaximumLength` function

ttbarnes (approved):

PR #2077, opened 2024-03-15
feat(EMS-2766): No PDF - Policy - Loss payee details - Save and back

Zainzzkk:
## Introduction :pencil2:
This PR adds save and back functionality to the loss payee details page

* Added save and back controller
* Updated routes
* Updated SAVE_AND_BACK_URL
* Updated and added tests.


src/ui/server/controllers/insurance/policy/save-data/nominated-loss-payee/index.ts
```diff
@@ -1,8 +1,19 @@
...
 
+import POLICY_FIELD_IDS from '../../../../../constants/field-ids/insurance/policy';
+
+const { NAME } = POLICY_FIELD_IDS.LOSS_PAYEE_DETAILS;
+
+/**
+ * string fields which are exempt from being stripped by stripEmptyFormFields
+ * for example when a string field needs to be set to an empty string or null
+ */
+export const NULL_OR_EMPTY_STRING_FIELDS = [NAME];
```
ttbarnes, line 15: Why is `NAME` exepmt from being set to empty/null - what happens if a user removes all text from the input and submits?

Zainzzkk: if you remove text from NAME and press save and back, it won't save your changes (remove the value for NAME) without this exemption

ttbarnes (approved):

PR #2092, opened 2024-03-18
fix(EMS-2969): No PDF - Your Buyer - Check your answers - Trading history fields

Zainzzkk:
## Introduction :pencil2:
This PR fixes an issue where `TRADING_HISTORY` fields were not wiped if an answer is changed from yes to no.

* Added to `mapSubmittedData` to wipe all trading history fields if `TRADED_WITH_BUYER` is false
* Added fields to `NULL_OR_EMPTY_STRING_FIELDS`
* Updated tests

### DIFF
--- a/e2e-tests/insurance/cypress/e2e/journeys/your-buyer/change-your-answers/change-your-answers-trading-history.spec.js
+++ b/e2e-tests/insurance/cypress/e2e/journeys/your-buyer/change-your-answers/change-your-answers-trading-history.spec.js
@@ -206,4 +206,42 @@ context('Insurance - Your buyer - Change your answers - Trading history - As an 
       });
     });
   });
+
+  describe(`changing ${TRADED_WITH_BUYER} from yes to no`, () => {
+    const fieldId = TRADED_WITH_BUYER;
+
+    describe('when clicking the `change` link', () => {
+      it(`should redirect to ${TRADED_WITH_BUYER_CHANGE}`, () => {
+        cy.navigateToUrl(url);
+
+        summaryList.field(OUTSTANDING_PAYMENTS).changeLink().click();
+        cy.completeAndSubmitTradingHistoryWithBuyerForm({ outstandingPayments: true, failedToPay: true });
+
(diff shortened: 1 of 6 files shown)


e2e-tests/insurance/cypress/e2e/journeys/your-buyer/change-your-answers/change-your-answers-trading-history.spec.js
```diff
@@ -206,4 +206,42 @@ context('Insurance - Your buyer - Change your answers - Trading history - As an
       });
     });
   });
+
+  describe(`changing ${TRADED_WITH_BUYER} from yes to no`, () => {
```
ttbarnes, line 210: Can we move this test into it's own E2E test? We have this pattern in other places, easier for debugging and reduces file size, e.g `change-your-answers-pre-credit-period-yes-to-no.spec`

src/ui/server/controllers/insurance/your-buyer/map-submitted-data/buyer-trading-history/index.test.ts
```diff
@@ -101,4 +101,58 @@ describe('controllers/insurance/your-buyer/map-submitted-data/buyer-trading-hist
...
+        ...expectedBody,
+        [FAILED_PAYMENTS]: null,
+      };
+
+      expect(result).toEqual(expected);
+    });
+  });
+
+  describe(`when ${TRADED_WITH_BUYER} is set to false`, () => {
+    it('should return mockFormBody without _csrf and relevant fields set to "null"', () => {
```
ttbarnes, line 140: I think the 2x new test descriptions could be simpler, maybe just 

```js
it(`should return the body with ${OUTSTANDING_PAYMENTS} set to null", () => {
it(`should return the body with ${FAILED_PAYMENTS} set to null`, () => {
it('should return the relevant TRADED_WITH_BUYER fields set to null', () => {
```

ttbarnes (approved):

PR #2098, opened 2024-03-18
fix(EMS-3079): No PDF - Policy - Loss Payee details - Wrong name validation

Zainzzkk:
## Introduction :pencil2:
This PR fixes an issue where the wrong validation was used for the name field on loss payee details

* Used providedAndMaxLength validation for name field
* Updated tests

### DIFF
--- a/e2e-tests/insurance/cypress/e2e/journeys/policy/loss-payee-details/validation/loss-payee-details-validation.spec.js
+++ b/e2e-tests/insurance/cypress/e2e/journeys/policy/loss-payee-details/validation/loss-payee-details-validation.spec.js
@@ -4,6 +4,7 @@ import { ACCOUNT_FIELDS } from '../../../../../../../content-strings/fields/insu
 import { INSURANCE_ROUTES } from '../../../../../../../constants/routes/insurance';
 import { INSURANCE_FIELD_IDS } from '../../../../../../../constants/field-ids/insurance';
 import partials from '../../../../../../../partials';
+import mockNameWithSpecialCharacters from '../../../../../../../fixtures/name-with-special-characters';
 
 const ERRORS = ERROR_MESSAGES.INSURANCE.POLICY;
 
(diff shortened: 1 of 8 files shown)


e2e-tests/insurance/cypress/e2e/journeys/policy/loss-payee-details/validation/loss-payee-details-validation.spec.js
```diff
@@ -87,6 +88,14 @@ context('Insurance - Policy - Loss Payee Details - Validation', () => {
 
       cy.submitAndAssertFieldErrors(field, value, errorIndex, numberOfExpectedErrors, ERROR.ABOVE_MAXIMUM);
     });
+
+    it(`should not render validation errors when ${FIELD_ID} contains numbers and special characters`, () => {
+      cy.keyboardInput(fieldSelector(FIELD_ID).input(), 'Name 123!');
```
ttbarnes, line 93: Can we use `mockNameWithSpecialCharacters` for this value

ttbarnes: I think we should test that when going back to the page, special characters are rendered as expected, we have similar tests e.g

```js
submit-policy-textarea-fields-with-special-characters.spec.js
submit-name-fields-with-special-characters/*.spec.js
```

src/ui/server/controllers/insurance/policy/loss-payee-details/validation/rules/name.ts
```diff
@@ -1,21 +1,24 @@
...
 
+export const MAXIMUM = Number(POLICY_FIELDS.LOSS_PAYEE_DETAILS[FIELD_ID].MAXIMUM);
+
 /**
  * validates name field
  * checks if response has been provided
  * @param {RequestBody} formBody
  * @param {Object} errors
- * @returns {Object} errors
+ * @returns {Object} providedAndMaxLength errors
```
ttbarnes, line 20: Should be 

```js
@returns {Function} providedAndMaxLength
```

src/ui/server/content-strings/fields/insurance/policy/index.ts
```diff
@@ -251,6 +251,7 @@ export const POLICY_FIELDS = {
   LOSS_PAYEE_DETAILS: {
     [LOSS_PAYEE_NAME]: {
       LABEL: 'Name of the loss payee',
+      MAXIMUM: 100,
```
ttbarnes, line 254: This could be in the new `constants/validation`, not a blocker since all name field validation rules will be re-visited soon, can be done then

ttbarnes (approved):

PR #2130, opened 2024-03-21
feat(EMS-2824-2827): No PDF - Account forms - Align maximum allowed email characters

ttbarnes:
## Introduction :pencil2:
This PR updates all account forms so that the email field renders errors if the total amount of provided characters is above a certain threshold.

For more context RE "Account sign in" validation:
- Before this change, If there is any validation error - even if an email is formatted incorrectly, we throw a generic validation for both the email and password fields, i.e "invalid credentials".
- Now, this remains exactly the same, however the email now has a custom rule so that if the email has over X amount of characters, a new, specific error message will be rendered. Otherwise, the email and password fields render the same generic "invalid credentials" validation error.

Note about email/apssword "show errors fro both"

- Update `MAXIMUM_CHARACTERS` constants to have an `EMAIL` definition.
- Update "Account - sign in" E2E test to call `assertEmailFieldValidation` and therefore all email validation assertions are run.
- Remove the now redundant `assertMaximumLength` flag from `assertEmailFieldValidation`, since all emails now have maximum length validation.
- Update `assertEmailFieldValidation` to accept and consume 2x new params/flags:
  - `isGenericErrorMessage`
  - `assertErrorWhenCorrectlyFormatted`
  - This allows for the DRY assertions to be assert a different error message or 1x particular scenario, depending on the context. I.e, in "account sign in" validation, we should **not** check that there is no validation error, when the email is valid. This is because the sign in validation intentionally renders an email validation error, if any part of the user provided credentials are invalid.
- Update the "Account sign in" POST controller to call `emailAndPasswordIncorrectValidationErrors` instead of `generateValidationErrors` with an empty object.
- Update "Account sign in" `emailRules` and `passwordRules` validation to run their own validations, instead of calling `emailAndPasswordValidation`.
  - This is now required because we need a new, custom validation rule specific to the email field (maximum characters).
- Create a new DRY helper function, `isValidEmail`.
- Rename/repurpose `emailAndPasswordValidation` to be `emailAndPasswordIncorrectValidationErrors`.

- Simplify `assertAllFieldErrors` E2E assertion function.
- Simplify a `describe` block in the "Different name on Policy" E2E test.
- Add new UI test mock, `mockValidEmail`, re-order some test mock imports.


e2e-tests/shared-test-assertions/email-field-validation/index.js
```diff
@@ -10,7 +10,8 @@ import partials from '../../partials';
  * @param {Object} errorMessages: Email error messages
  * @param {Integer} totalExpectedErrors: Total expected errors in the form
  * @param {Integer} totalExpectedOtherErrorsWithValidEmail: Total expected errors in the form when an email is valid.
- * @param {Boolean} assertMaximumLength: Flag for whether to assert maximum length (currently, some fields do not have this).
+ * @param {Integer} isGenericErrorMessage: Email has a generic error message for incorrect formatting errors.
```
avaitonis, line 13: Type for new fields should be boolean

avaitonis (approved):
LGTM

PR #2139, opened 2024-03-22
feat(EMS-2562): No PDF - Export contract - How you will get paid - Page setup

ttbarnes:
## Introduction :pencil2:
This PR adds the initial setup for the "How you will get paid"  page in the "Export contract" section of a No PDF application.

Note: This is purely setting up the page, no form validation, data saving or UI/UX flow updates. These will be added in other PRs.

- Update `assertTextareaRendering` cypress command to have an optional label check; Some pages do not have a specific textarea label because the page heading is the "label".
- Add `PAYMENT_TERMS_DESCRIPTION` field ID, content strings and `MAXIMUM_CHARACTERS` constant definition.
- Add initial `HOW_WILL_YOU_GET_PAID` UI routes.
- Add E2E test.
- Create new UI GET controller, route and nunjucks template.

- Improve some export contract imports/exports.


src/api/constants/field-ids/insurance/export-contract/index.ts
```diff
@@ -1,12 +1,15 @@
...
     DESCRIPTION: 'goodsOrServicesDescription',
     FINAL_DESTINATION_KNOWN: 'finalDestinationKnown',
     FINAL_DESTINATION: 'finalDestinationCountryCode',
   },
+  HOW_WILL_YOU_GET_PAID: {
+    PAYMENT_TERMS_DESCRIPTION: 'paymentTermsDescription',
+  },
 };
 
 export default EXPORT_CONTRACT;
```
avaitonis, line 15: Do we need second `default` export?

ttbarnes: Good question - at the moment yes.

In the main `field-ids` index file, we import via the default export.

But in E2E tests, we import via destructuring so without the default export.

Seems we should align this, but wouldn't be quick as should apply to all other field ID directories

avaitonis (approved):
LGTM

PR #2144, opened 2024-03-25
feat(EMS-2563): No PDF - Export contract - How will you get paid - Form validation

ttbarnes:
## Introduction :pencil2:
This PR updates the "How will you get paid" POST controller so that the `PAYMENT_TERMS_DESCRIPTION` field must be provided and less than a certain amount of characters.

- Update E2E test to include form validation.
- Create new UI validation rules.
- Update the UI POST controller to run validation error checks.

- Add missing directory to `test.yml`.
- Simplify some validation rules documentation.


src/ui/server/shared-validation/full-address/index.ts
```diff
@@ -12,7 +12,7 @@ import { RequestBody, ErrorMessageObject } from '../../../types';
  * @param {String} fieldId: Field ID
  * @param {String} errorMessage: Error message message
  * @param {Object} errors: Other validation errors for the same form
- * @returns {Object} providedAndMaxLength
+ * @returns {Function} providedAndMaxLength
```
avaitonis, line 15: We call function `providedAndMaxLength` but we don't return function itself, function return would look like `return providedAndMaxLength;`.

Return type might be `ValidationErrors` from '../../../types';

ttbarnes: Might be a misread, we do return `providedAndMaxLength`, on line 17

```js
return providedAndMaxLength( ...
```

src/ui/server/shared-validation/full-address/index.ts
```diff
@@ -12,7 +12,7 @@ import { RequestBody, ErrorMessageObject } from '../../../types';
...
  */
 const fullAddress = (formBody: RequestBody, fieldId: string, errorMessages: ErrorMessageObject, errors: object) => {
   return providedAndMaxLength(formBody, fieldId, errorMessages, errors, MAXIMUM_CHARACTERS.FULL_ADDRESS);
```
ttbarnes, line 18: @avaitonis Here is where it's returned

src/ui/server/shared-validation/full-address/index.ts
avaitonis: We have () on line 17, so we are calling the function and returning it's result, not function itself.

```
// return function
return Math.abs;

// return result of function, number
return Math.abs(-1);
```

src/ui/server/shared-validation/full-address/index.ts
avaitonis: yep, looks like function call and then return.

Most likely return is {ValidationErrors} or {Object}
For async it would be {Promise<ValidationErrors>} or {Promise<Object>}

src/ui/server/shared-validation/full-address/index.ts
ttbarnes: Good spot, understood. We had a few instances of this, i've updated them all so it's now e.g

```js
* @returns {ValidationErrors} fullAddressValidation
```

Commit => https://github.com/UK-Export-Finance/exip/pull/2144/commits/46fe277d1e658f38d4a28a2dc0d3d7e8f7bc0148

PR #2169, opened 2024-03-28
chore(EMS-3116): No PDF - Simplify referenceNumber instances, documentation improvements

ttbarnes:
## Introduction :pencil2:
This PR introduces 2x improvements and fixes where:
- Various validation functions were referencing the `formBody` param as `Express.Response.body`, where it should be `RequestBody`.
- Various controllers were obtaining the reference number from `req.params` when it could be simpler and avoid `Number(referenceNumber)` instances by simply using `application.referenceNumber`.

- Align `formBody` documentation.
- Update various UI controllers to use `referenceNumber` from `res.locals.application`, instead of params.


avaitonis (approved):
LGTM

PR #2176, opened 2024-04-02
fix(EMS-3078): No PDF - Single/multiple contract policy - Alternative currency validation - Pre-selected radio

ttbarnes:
## Introduction :pencil2:
This PR fixes an issue in the single and multiple contract policy forms where, if the "alternative currency" radio option is selected, but there are other validation errors, the "alternative currency" radio option  would not be pre-selected.

- Update the DRY, shared E2E currency form field assertions (`assertCurrencyFormFields`), to check that the "alternative currency" radio option is checked when there are validation errors.
- Update single and multiple contract policy POST controllers to conditionally use the `POLICY_CURRENCY_CODE` from the submtited values during validation errors.

- Remove some unnecessary `submittedValues` properties  from 2x GET controllers.
- Fix a UI date validation unit test that was failing due to the current time of year.
- Fix an E2E test that was failing due to an end date being generated as before the start date, due to the current time of year.

### DIFF
--- a/e2e-tests/commands/insurance/assert-currency-form-fields.js
+++ b/e2e-tests/commands/insurance/assert-currency-form-fields.js
@@ -115,6 +115,10 @@ const assertCurrencyFormFields = ({
 
     cy.clickSubmitButton();
 
+    const option = currencyRadio({ fieldId: alternativeCurrencyFieldId });
+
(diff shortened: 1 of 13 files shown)


avaitonis (approved):
LGTM

PR #2249, opened 2024-04-12
feat(EMS-2860): No PDF - Policy - Check your answers - Other company to be insured

Zainzzkk:
## Introduction :pencil2:
This PR adds the check your answers section for other company to be insured

* Added summary list section for other company to be insured
* Added countries call to controller for policy check your answers
* Added tests 
* Added e2e tests


src/ui/server/controllers/insurance/check-your-answers/policy/index.ts
```diff
@@ -60,6 +60,7 @@ export const get = async (req: Request, res: Response) => {
     const { isUsingBroker } = broker;
 
     const { supportedCurrencies } = await api.keystone.APIM.getCurrencies();
+    const countries = await api.keystone.countries.getAll();
```
ttbarnes, line 63: Can we add a check just like below please, 

```js
if (!isPopulatedArray(countries)) {
  return res.redirect(PROBLEM_WITH_SERVICE);
}
```

src/ui/server/controllers/insurance/check-your-answers/policy/index.test.ts
```diff
@@ -159,6 +162,19 @@ describe('controllers/insurance/check-your-answers/policy', () => {
           expect(res.redirect).toHaveBeenCalledWith(PROBLEM_WITH_SERVICE);
         });
       });
+
```
ttbarnes, line 165: Can we add a test for calling the API? Seems it's missing for currencies as well (probably my fault, sorry)

```js
it('should call api.keystone.countries.getAll', async () => {
it('should call api.keystone.currencies.getAll', async () => {
``

File: src/ui/server/controllers/insurance/policy/check-your-answers/index.test.ts [hunk 3]
```diff
@@ -132,6 +135,19 @@ describe('controllers/insurance/policy/check-your-answers', () => {
           expect(res.redirect).toHaveBeenCalledWith(PROBLEM_WITH_SERVICE);
         });
       });
+
```
ttbarnes, line 138: Same as above

src/ui/server/controllers/insurance/check-your-answers/policy/index.ts
```diff
@@ -60,8 +60,9 @@ export const get = async (req: Request, res: Response) => {
...
 
-    if (!isPopulatedArray(supportedCurrencies)) {
+    if (!isPopulatedArray(supportedCurrencies) || !isPopulatedArray(countries)) {
```
ttbarnes, line 65: Need to add 2x tests, happy to do in the next PR e.g:

```js
describe('when the get currencies response does not return a populated array', () => {
```

PR #2335, opened 2024-04-30
chore(EMS-3191): No PDF - Export contract - Agent details - Validation - Improve E2E test coverage

ttbarnes:
## Introduction :pencil2:
This PR updates the E2E test coverage for the "Agent details" form, in the Export contract section of a No PDF application.

A bug was raised where the `COUNTRY_CODE` field in this page would not render a previously submitted value, if other fields have errors.

Unable to replicate this, but have added E2E test coverage for all fields, to ensure that every field retains a previously submitted value, when there are validation errors for other fields.

- Update E2E test coverage.


Zainzzkk (approved):
massive PR.  Great work.  Took hours to review ;)

PR #2336, opened 2024-04-30
fix(EMS-3194): No PDF - Buyer - Trading history - Validation - Missing/stripped submitted values

ttbarnes:
## Introduction :pencil2:
This PR fixes an issue in the "Buyer - Trading history" form, where if a monetary field is submitted as "0" and there are validation errors, the previously submitted "0"'s, would not be rendered in the input.

- Update E2E test coverage.
- Remove `sanitiseData` from the POST controller's `submittedValues`.
  - This was not only (correctly) transforming a zero from a string into a number, but was also stripping previously submitted special characters (E.g commas), which should remain intact during validation errors - so that we playback to a user exactly what was entered.

- Fix a typo.


Zainzzkk (approved):
Another big one!

PR #2350, opened 2024-04-30
feat(EMS-1175): No PDF - CRON jobs - Inactive application helper

Zainzzkk:
## Introduction :pencil2:
This PR adds a helper to find an update the status of inactive applications

* Added `updateInactiveApplications` to check for inactive applications and update status
* Added status to find inactive applications and to map and update them
* Added tests


src/api/constants/application/index.ts
```diff
@@ -38,6 +38,7 @@ export const APPLICATION = {
   STATUS: {
     IN_PROGRESS: 'In progress',
     SUBMITTED: 'Submitted to UKEF',
+    ABANDONED: 'Abandoned',
```
ttbarnes, line 41: Are the E2E and UI constants also updated?

src/api/custom-resolvers/mutations/update-inactive-applications/index.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Application, Context } from '../../../types';
+
+const { IN_PROGRESS, ABANDONED } = APPLICATION.STATUS;
+
+describe('custom-resolvers/update-inactive-applications', () => {
+  let context: Context;
+  let application: Application;
+
+  const now = new Date();
+  const twoMonthsAgo = new Date(now.setDate(now.getDate() - 60));
```
ttbarnes, line 14: Can this be in the constants like other date generations, e.g

```js
DATE_24_HOURS_FROM_NOW
DATE_24_HOURS_IN_THE_PAST
DATE_30_MINUTES_FROM_NOW
```

src/api/custom-resolvers/mutations/update-inactive-applications/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import getInactiveApplications from '../../../helpers/get-inactive-applications';
+import mapAndUpdateInactiveApplications from '../../../helpers/map-and-update-inactive-applications';
+
+/**
+ * getInactiveApplications
+ * Gets inactive applications which have not been updated for 30 days
+ * Sets their status to Abandoned
+ * returns success flag
+ * @param {Object} GraphQL root variables
+ * @param {Object} KeystoneJS context API
```
ttbarnes, line 12: {Context}

src/api/custom-resolvers/mutations/update-inactive-applications/index.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+const updateInactiveApplicationsMutation = async (root: any, context: Context): Promise<SuccessResponse> => {
+  try {
+    console.info('Getting and updating inactive applications');
+
+    /**
+     * gets inactive applications based on status and updatedAt
+     * returns array of application ids and statuses
+     */
+    const applications = await getInactiveApplications(context);
```
ttbarnes, line 23: Maybe should change the comment to simply "get inactive applications"

Just thinking, if `getInactiveApplications` changes to something different, this comment could become misleading

src/api/helpers/date/index.ts
```diff
@@ -42,3 +42,16 @@ export const getYesterdayDay = (): number => {
 
   return result;
 };
+
+/**
+ * getThirtyDaysBefore
+ * Get 30 days before now and return the timestamp
+ * @returns {Date} 30 days before today
+ */
+export const getThirtyDaysBefore = (): Date => {
```
ttbarnes, line 51: Could we rename this to be `getThirtyDaysBeforeNow`, otherwise feels like "before" could be a param

src/api/custom-resolvers/mutations/update-inactive-applications/index.test.ts
ttbarnes: Or in the helpers, not sure which is best off the top of my head.

src/api/helpers/get-inactive-applications/index.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  beforeEach(async () => {
+    await createFullApplication(context);
+    await createFullApplication(context);
+    await createFullApplication(context);
+    application4 = await createFullApplication(context);
+    application5 = await createFullApplication(context);
+    application6 = await createFullApplication(context);
+  });
```
ttbarnes, line 33: Can we create a test helper for this, i.e `createMultipleFullApplications`, so it's just a one liner

```js
const applications = await createMultipleFullApplications();
```

src/api/helpers/get-inactive-applications/index.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      await context.db.Application.updateOne({
+        where: { id: application4.id },
+        data: updateData,
+      });
+
+      await context.db.Application.updateOne({
+        where: { id: application5.id },
+        data: updateData,
+      });
+    });
```
ttbarnes, line 54: Could we use the existing `applications.update` test helper?

src/api/helpers/get-inactive-applications/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ * returns array of application ids and statuses
+ * @param {Context} context
+ * @returns {Promise<Application[]>} Array of application ids and status
+ */
+const getInactiveApplications = async (context: Context): Promise<Application[]> => {
+  try {
+    console.info('Getting inactive applications - getInactiveApplications helper');
+
+    // timestamp of 30 days ago
+    const thirtyDaysLimit = getThirtyDaysBefore();
```
ttbarnes, line 19: Code is self explanatory, no need for the comment IMO

ttbarnes: Should probably have a session, maybe this should go in `constants/application`, or the application version definitions - incase it changes

Am also wondering how/if the "30 days" check will work against the existing deadline definition that a user would see, `SUBMISSION_DEADLINE_IN_MONTHS`

src/api/helpers/map-and-generate-inactive-applications-save-array/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * mapAndGenerateInactiveApplicationSaveArray
+ * generates array of objects for updating the db
+ * object contains where object with id of application to be updated
+ * contains data object with fields which are updated
+ * @param {Array<InactiveApplicationUpdate>} applications: array of application ids and statuses
+ * @returns {Array<InactiveApplicationUpdate>}: Array of objects for updating db
+ */
+const mapAndGenerateInactiveApplicationSaveArray = (applications: Array<Application>) => {
```
ttbarnes, line 13: Not sure about the naming, how about

`mapInactiveApplications`

Could we use a `.map` instead of `.forEach`?

Maybe a `.map` with a `mapInactiveApplication` for each object in the array 🤔

src/api/helpers/map-and-update-inactive-applications/index.ts
```diff
@@ -0,0 +1,30 @@
+import { Application, Context } from '.keystone/types'; // eslint-disable-line
+import mapAndGenerateInactiveApplicationsSaveArray from '../map-and-generate-inactive-applications-save-array';
+
+/**
+ * mapAndUpdateInactiveApplications
+ * maps and generates save object for inactive applications
+ * updates application statuses to abandoned
+ * @param {Array<Application>} applications: array of application ids and statuses
```
ttbarnes, line 8: I think just `array of applications` is simpler, what do you think?

src/api/helpers/map-and-update-inactive-applications/index.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Context} context
+ */
+const mapAndUpdateInactiveApplications = async (applications: Array<Application>, context: Context) => {
+  try {
+    console.info('Mapping and updating inactive applications - mapAndUpdateInactiveApplications');
+
+    /**
+     * generates data array for database saving
+     * contains where and data (status change)
+     */
```
ttbarnes, line 18: Just thought - could you check the application's `updatedAt` fields are updated after calling `context.db.Application.updateMany`?

It might be doing so automatically; If not we should update it in the mapping function

src/api/types/cron/index.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface ApplicationDataUpdate {
+  status: string;
+  previousStatus: string;
+}
+
+export interface InactiveApplicationUpdate {
+  where: ApplicationWhereUpdate;
+  data: ApplicationDataUpdate;
+}
```
ttbarnes, line 13: Should we maybe prefix these types with `Cron` e.g 

```js
CronApplicationWhereUpdate
CronApplicationDataUpdate
CronApplicationInactiveupdate
```

Just thinking so it's clear that it's specifically for a CRON job

src/api/helpers/map-and-update-inactive-applications/index.ts
Zainzzkk: its done automatically

src/api/helpers/map-and-generate-inactive-applications-save-array/index.ts
Zainzzkk: done

src/api/custom-resolvers/mutations/update-inactive-applications/index.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    return {
+      success: true,
+    };
+  } catch (err) {
+    console.error('Error getting and updating inactive applications %O', err);
+    throw new Error(`Error getting and updating inactive applications ${err}`);
+  }
+};
+
```
ttbarnes, line 39: Can we change this to be a helper - if it's not going to be added to the GQL schema, it's not a GQL mutation

I imagine this will be called from middleware rather than GQL.

src/api/custom-resolvers/mutations/update-inactive-applications/index.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+      const expected = await context.query.Application.findOne({
+        where: { id: application.id },
+        query: 'id status previousStatus',
+      });
+
+      expect(expected.status).toEqual(ABANDONED);
+      expect(expected.previousStatus).toEqual(IN_PROGRESS);
+    });
+  });
```
ttbarnes, line 46: Could we add a test for the `updatedAt`? Could do e.g

```js
expect(latestApplication.updatedAt).not.toEqual(application.updatedAt);
expect(latestUpdatedAtTimeStamp).toBeGreaterThan(timestamp);
```

src/api/helpers/map-and-update-inactive-applications/index.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+      await mapAndUpdateInactiveApplications(applications, context);
+
+      const expected = await context.query.Application.findOne({
+        where: { id: application.id },
+        query: 'id status previousStatus',
+      });
+
+      expect(expected.status).toEqual(ABANDONED);
+      expect(expected.previousStatus).toEqual(IN_PROGRESS);
```
ttbarnes, line 37: Same as above RE `updatedAt`

src/api/helpers/map-inactive-applications/index.test.ts
```diff
@@ -0,0 +1,50 @@
...
+            status: ABANDONED,
+            previousStatus: applications[0].status,
+          },
+        },
+        {
+          where: { id: applications[1].id },
+          data: {
+            status: ABANDONED,
+            previousStatus: applications[1].status,
+          },
```
ttbarnes, line 43: Could destructure `applications` here so easier to read and array indexing e.g

```js
const [application0, application1] = applications;

// ...

where: { id: application0.id },
data: {
  status: ABANDONED,
  previousStatus: application0.status,
},
```

src/api/test-helpers/create-multiple-full-applications.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param {Object} KeystoneJS context API
+ * @returns {Array<Application>} Applications array
+ */
+export const createMultipleFullApplications = async (context: Context) => {
+  const application1 = await createFullApplication(context);
+  const application2 = await createFullApplication(context);
+  const application3 = await createFullApplication(context);
+  const application4 = await createFullApplication(context);
+  const application5 = await createFullApplication(context);
+  const application6 = await createFullApplication(context);
```
ttbarnes, line 16: New `createMultipleFullApplications` test helper? Something like 


```js
const createMultipleFullApplications = async (count: number) => {
  const arr = new Array(count).fill();

  arr.forEach(async () => {
    await createFullApplication();
  });
};
```

`Promise.all()` would be even better but i'm not sure if that could work right now without refactoring `createFullApplication`.

src/api/helpers/map-inactive-applications/index.ts
```diff
@@ -0,0 +1,36 @@
...
+        status: APPLICATION.STATUS.ABANDONED,
+        previousStatus: application.status,
+        updatedAt: new Date(),
+      },
+    } as CronApplicationInactiveUpdate;
+
+    return update;
+  });
+
+  return updateArray;
```
ttbarnes, line 33: Not a blocker, can be done after this PR, i'd use `mapped` instead of `update` since it's not actually updating anything, is purely mapping

src/api/helpers/update-inactive-applications/index.test.ts
```diff
@@ -0,0 +1,55 @@
+import updateInactiveApplicationsMutation from '.';
```
ttbarnes, line 1: `updateInactiveApplications`

src/api/helpers/update-inactive-applications/index.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * updateInactiveApplications
+ * Gets inactive applications which have not been updated for 30 days
+ * Sets their status to Abandoned
+ * returns success flag
+ * @param {Object} GraphQL root variables
+ * @param {Context} KeystoneJS context API
+ * @returns {Promise<SuccessResponse>} success flag
+ */
+const updateInactiveApplications = async (root: any, context: Context): Promise<SuccessResponse> => {
```
ttbarnes, line 15: Can remove the `root` variable since it's not a GQL resolver

src/api/constants/index.ts
```diff
@@ -82,6 +82,22 @@ export const DATE_ONE_MINUTE_IN_THE_PAST = () => {
...
+ * Generate a date that is 2  months in the past
+ * @returns {Date}
+ */
+export const DATE_2_MONTHS_IN_THE_PAST = () => {
+  const now = new Date();
+
+  const day = now.getDate();
+
+  // subtract 1 day from the current day.
+  const twoMonths = new Date(now.setDate(day - 60));
```
ttbarnes, line 96: 1 day vs `twoMonths`, `day - 60`?

ttbarnes (approved):
Minor documentation comment, happy for this to be resolved in the next PR.

src/api/test-helpers/create-multiple-full-applications.ts
Zainzzkk: this will be done later, was not working

PR #2351, opened 2024-04-30
chore(EMS-3113): No PDF - Replace MOJ password reveal with GOV.UK component. Bump dependencies

ttbarnes:
## Introduction :pencil2:
GOV.UK has recently introduced a new password reveal component.

Therefore, we can remove MOJ from EXIP and use the new GOV.UK component.

- Update E2E tests.
- Remove MOJ depedency.
- Remove MOJ SRI constant.
- Update `password-input` nunjucks component.
- Remove MOJ client side script instances.
- Remove MOJ script from webpack config.

- Bump dependencies.


e2e-tests/commands/insurance/account/assert-password-reveal-button.js
```diff
@@ -8,13 +8,16 @@ const {
 
 const FIELD_STRINGS = ACCOUNT_FIELDS[FIELD_ID];
 
-const expectedShowText = `${FIELD_STRINGS.REVEAL.SHOW} password`;
-const expectedHideText = `${FIELD_STRINGS.REVEAL.HIDE} password`;
+const EXPECTED_ARIA_LABELS = {
```
Zainzzkk, line 11: not a blocker, should these be constants instead?

src/ui/package.json
```diff
@@ -22,10 +22,9 @@
     "test-watch": "jest --watchAll --notify --coverage --verbose --config=jest.config.js --detectOpenHandles"
   },
   "dependencies": {
-    "@babel/core": "^7.24.4",
+    "@babel/core": "^7.24.5",
     "@babel/plugin-transform-runtime": "^7.24.3",
-    "@babel/preset-env": "^7.24.4",
-    "@ministryofjustice/frontend": "^2.1.2",
```
Zainzzkk, line 28: bye bye MOJ :(

e2e-tests/commands/insurance/account/assert-password-reveal-button.js
ttbarnes: Good shout, although currently, it's only relevant to this E2E test - this text is automatically generated by the password component.

I think we should keep this in here for now. I can't see a scenario where this specific password reveal copy would be used somewhere else.

PR #2359, opened 2024-05-02
feat(EMS-2772): No PDF - Task list - "Insurance policy" - Loss payee - Financial details UK

ttbarnes:
## Introduction :pencil2:
This PR updates the  "Insurance policy" task list group so that the "Loss payee - Financial details UK" fields are required.

- Update `completeLossPayeeFinancialDetailsUkForm` cypress command to conditionally clear all inputs.
- Create new `LOSS_PAYEE_NAME` and `LOSS_PAYEE_FINANCIAL_ADDRESS` field IDs.
- Update E2E tests.
- Update `ALL_SECTIONS` GET controller and task list helper functions to pass `isAppointingLossPayee` and `lossPayeeIsLocatedInUk` flags to `generateGroupsAndTasks`.
- Update `flattenApplicationData` helper function to map Loss payee details.
- Update required policy fields helper function to map Loss payee field IDs.
- Update "Get application" GQL query to return encrypted financial details.

- Rename `appointingLossPayee` params to `isAppointingLossPayee`.
- Update API field IDs.
- Simplify some documentation.
- Update UI application types.


src/ui/server/graphql/queries/application.ts
```diff
@@ -46,9 +46,15 @@ const applicationQuery = gql`
           name
           financialUk {
             id
+            accountNumber
```
Zainzzkk, line 49: I thought these already existed for playback?

Zainzzkk (approved):

src/ui/server/graphql/queries/application.ts
ttbarnes: Yes, but not in the "All sections" task list context.

The other GQL queries that have this are for showing encrypted data.

For "All sections", we need to check the data exists, but it doesn't have to be decrypted. So we can simply update the standard "Get application" GQL query to return the encrypted) data 👍

PR #2363, opened 2024-05-02
chore(EMS-3302): No PDF - Account - Split Account table into Account and AccountStatus 

Zainzzkk:
## Introduction :pencil2:
This PR splits the Account table with AccountStatus

* Added AccountStatus to schema
* Added AccountStatus to SQL dump
* Updated usage of AccountStatus data
* Updated GQL calls
* Added test helpers
* Updated tests

- Updated test.yml to correctly split check-your-answers/export-contract tests


src/api/schema.ts
```diff
@@ -522,9 +515,24 @@ export const lists = {
...
+    fields: {
+      account: relationship({ ref: 'Account.accountStatus' }),
+      isVerified: checkbox({ defaultValue: false }),
+      /**
+       * isBlocked flag will only be true if the account has:
+       * - repeatedly attempted sign in
+       * - repeatedly attempted password reset request
+       */
+      isBlocked: checkbox({ defaultValue: false }),
+      isInactivated: checkbox({ defaultValue: false }),
```
ttbarnes, line 532: Can this be structured to be dynamic instead of individual status columns? If we have e.g 6-10 statuses, we'd have to refactor.

Could be just like `TotalContractValue` or `CoverPeriod`,

src/api/helpers/get-account-by-field/index.ts
```diff
@@ -27,7 +27,23 @@ const getAccountByField = async (context: Context, field: string, value: string)
...
+    /**
+     * constructs full account object
+     * adds accountStatus data
+     */
+    const fullAccount = {
+      ...account,
+      accountStatus: {
+        ...accountStatus,
+      },
+    };
```
ttbarnes, line 44: We could do `status` instead of `accountStatus`, so when references in other places its just `account.status` rather than `account.accountStatus`, what do you think? Would also make this simpler

```js
const fullAccount = {
  ...account,
  status: accountStatus,
};
```

src/api/custom-resolvers/mutations/create-an-account/index.ts
```diff
@@ -58,6 +57,17 @@ const createAnAccount = async (root: any, variables: AccountCreationVariables, c
       data: accountData,
     });
 
+    // creates account status with links to account
```
ttbarnes, line 60: `creates account status relationship`?

src/api/helpers/update-account/index.test.ts
```diff
@@ -60,4 +60,42 @@ describe('helpers/update-account', () => {
...
+    await update.accountStatus(context, account.accountStatus.id, accountUpdate);
+
+    account = await accounts.get(context, account.id);
+
+    const { accountStatus } = account;
+
+    expect(accountStatus.isBlocked).toEqual(true);
+    expect(accountStatus.isVerified).toEqual(true);
+    expect(accountStatus.isInactivated).toEqual(false);
+  });
```
ttbarnes, line 79: Can we add a `lastUpdated` field and unti test?

src/api/helpers/update-account/index.test.ts
```diff
@@ -60,4 +60,42 @@ describe('helpers/update-account', () => {
...
+        await update.accountStatus(context, account.accountStatus.id, {});
+      } catch (err) {
+        const expectedKeystoneError = 'Access denied: You cannot update that AccountStatus - it may not exist';
+
+        const expected = new Error(`Updating account status ${expectedKeystoneError}`);
+
+        expect(err).toEqual(expected);
+      }
+    });
+  });
```
ttbarnes, line 100: Could we use `rejects.toThrow` here?

src/api/helpers/update-account/index.ts
```diff
@@ -27,8 +27,36 @@ const account = async (context: Context, accountId: string, updateData: object)
   }
 };
 
+/**
+ * accountStatus
+ * Update an accountStatus
+ * @param {Object} KeystoneJS context API
+ * @param {String} AccountStatus ID
+ * @param {Object} AccountStatus update data
```
ttbarnes, line 35: typo with `Object` param, also should this have the param names e.g

```js
* @param {Object} context: KeystoneJS context API
* @param {String} accountStatusId: Account status ID
* @param {Object} updateData: Update data
```

src/api/schema.ts
ttbarnes: Had a team chat about this - the approach we have here feels the best, simplest approach.

If there are more statuses added this could become troublesome and should probably be split up more into 2x different tables related to an account. However, it is extremely unlikey for us to require any additional statuses.

At this stage, other approaches feel like over engineering.

If we need to have a different, more split approach, this shouldn't be to cumbersome to achieve in the future.

src/api/test-helpers/account-status.ts
```diff
@@ -0,0 +1,125 @@
...
+    throw new Error(`Getting and deleting account status (test helpers) ${err}`);
+  }
+};
+
+/**
+ * create account status test helper
+ * Create an account with mock accountStatus data and any connections to account
+ * @param {Object} KeystoneJS context API, accountStatus data, accountId and deleteAccountStatus flag
+ * @returns {TestHelperAccountStatusCreate} Created accountStatus
+ */
```
ttbarnes, line 36: Missing some documentation here

src/api/test-helpers/account-status.ts
```diff
@@ -0,0 +1,125 @@
+import { mockAccount } from '../test-mocks';
+import { AccountStatus, AccountStatusUpdate, Context, TestHelperAccountStatusCreate } from '../types';
+
+/**
+ * deleteAll test helper
+ * Get all account-status and delete them.
+ * @param {Object} KeystoneJS context API
```
ttbarnes, line 7: `Context`?

src/api/types/account/index.ts
```diff
@@ -87,6 +87,19 @@ export interface AccountSignInVariables {
...
+  isBlocked?: boolean;
+  isInactivated?: boolean;
+}
+
+export interface AccountStatusUpdate {
+  isVerified?: boolean;
+  isBlocked?: boolean;
+  isInactivated?: boolean;
+}
+
```
ttbarnes, line 102: Could have an `AccountCore` interface that gets extended to be DRY e.g

```ts
export interface AccountStatus extends AccountStatusCore {
  id: string;
}
```

database/exip.sql
```diff
@@ -67,6 +67,19 @@ CREATE TABLE `_AuthenticationRetry_account` (
   CONSTRAINT `_AuthenticationRetry_account_B_fkey` FOREIGN KEY (`B`) REFERENCES `AuthenticationRetry` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
 
+# Dump of table AccountStatus
```
ttbarnes, line 70: Can we update the "generation time" at the top of this file

src/api/custom-resolvers/mutations/create-an-account/index.ts
```diff
@@ -58,6 +57,17 @@ const createAnAccount = async (root: any, variables: AccountCreationVariables, c
...
+    await context.db.AccountStatus.createOne({
+      data: {
+        account: {
+          connect: {
+            id: creationResponse.id,
+          },
+        },
+      },
+    });
+
```
ttbarnes, line 70: Can we update the documentation at the top of the file please

src/api/custom-resolvers/mutations/verify-account-reactivation-token/index.test.ts
```diff
@@ -54,17 +57,17 @@ describe('custom-resolvers/verify-account-reactivation-token', () => {
     const reactivationExpiry = ACCOUNT.REACTIVATION_EXPIRY();
 
     const unverifiedAndBlockedAccount = {
-      ...mockAccount,
-      [IS_VERIFIED]: false,
-      [IS_BLOCKED]: true,
+      ...mockAccountUpdate,
```
ttbarnes, line 60: `IS_VERIFIED` and `IS_BLOCKED` have been removed here and replaced with `mockAccountUpdate`, however the flags are different:

In the mocks:
```ts
isVerified: true,
isBlocked: false,
```

Which is the reverse of the original test.

Not sure if it should still pass or if the original flags are required (off the top of my head), can we check please?

src/api/package.json
```diff
@@ -19,7 +19,7 @@
     "lint": "eslint ./",
     "lint:fix": "eslint ./ --fix",
     "start": "keystone start --no-ui",
-    "test": "jest --runInBand --verbose --silent --coverage --config=jest.config.js"
+    "test": "jest --runInBand --verbose --coverage --config=jest.config.js"
```
ttbarnes, line 22: Revert 😛

src/api/test-helpers/account-status.ts
```diff
@@ -0,0 +1,129 @@
+import { mockAccount } from '../test-mocks';
+import { AccountStatus, AccountStatusCore, Context, TestHelperAccountStatusCreate } from '../types';
+
+/**
+ * deleteAll test helper
+ * Get all account-status and delete them.
```
ttbarnes, line 6: `AccountStatus` instead of `account-status`?

src/api/test-helpers/account-status.ts
```diff
@@ -0,0 +1,129 @@
...
+  } catch (err) {
+    console.error('Error creating account status %O', err);
+    return err;
+  }
+};
+
+/**
+ * get account status test helper
+ * Get an accountStatus by ID
+ * @param {Object} KeystoneJS context API
```
ttbarnes, line 78: `Context` like above 😛

src/api/test-helpers/account-status.ts
```diff
@@ -0,0 +1,129 @@
...
+  } catch (err) {
+    console.error(err);
+    throw new Error(`Getting an account status by ID (test helpers) ${err}`);
+  }
+};
+
+/**
+ * update account-status test helper
+ * Updates an accountStatus by accountStatus ID
+ * @param {Object} KeystoneJS context API
```
ttbarnes, line 101: Same as above

ttbarnes (commented):
Great work, few minor comments

src/api/custom-resolvers/mutations/verify-account-reactivation-token/index.test.ts
Zainzzkk: the status is removed at the top of the file and added on line 66 to the correct table.  line 66 has `[IS_VERIFIED]: false, [IS_BLOCKED]: true`

ttbarnes (approved):

PR #2389, opened 2024-05-08
fix(EMS-3312-3327): No PDF - Broker name validation, minor content updates/fixes

ttbarnes:
## Introduction :pencil2:
This PR fixes 3x issues in a No PDF application where:
- The "Broker name" field would error if more than 191 characters are submitted. It should allow up to 800 characters.
- The `TOTAL_CONTRACT_VALUE` error message had a double "the the" and missing copy.
- The `TOTAL_VALUE_INSURED` hint text had "polices", when it should be "policies".

- Update SQL dump.
- Update field content strings.


Zainzzkk (approved):
can we get my PR in first please though

Zainzzkk (approved):

PR #2394, opened 2024-05-10
feat(EMS-1175): No PDF - CRON jobs - Unverified account helper

Zainzzkk:
## Introduction :pencil2:
This PR adds a helper to find and update the `isInactive` flag for account status.

* Added `updateUnverifiedAccounts` to check for unverified accounts over 24hrs and update `isInactive` flag
* Added tests


src/api/helpers/get-unverified-accounts/index.ts
```diff
@@ -0,0 +1,41 @@
...
+     */
+    const accounts = (await context.query.Account.findMany({
+      where: {
+        AND: [
+          { verificationExpiry: { lt: new Date() } },
+          { status: { isVerified: { equals: false } } },
+          { status: { isBlocked: { equals: false } } },
+          { status: { isInactive: { equals: false } } },
+        ],
+      },
```
ttbarnes, line 28: If the function is called `getUnverifiedAccounts`, why does it care about `isBlocked` and `isInactive`?

src/api/helpers/map-and-update-unverified-accounts/index.ts
```diff
@@ -0,0 +1,40 @@
+import { Context } from '.keystone/types'; // eslint-disable-line
+import mapUnverifiedAccounts from '../map-unverified-accounts';
+import { Account } from '../../types';
+
+/**
+ * mapAndUpdateUnverifiedAccounts
+ * maps and generates save object for unverified accounts
+ * updates AccountStatus' isInactive flag to true and updates AccountStatus and Account updatedAt
+ * @param {Array<Account>} accounts: array of applications
+ * @param {Context} context
```
ttbarnes, line 10: `* @param {Context} context: KeystoneJS context API`

src/api/helpers/map-unverified-accounts/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import { mockAccount } from '../../test-mocks';
+
+describe('helpers/map-unverified-accounts', () => {
+  describe('when an empty array is passed', () => {
+    it('should return an empty account array', () => {
+      const result = mapUnverifiedAccounts([]);
+
+      expect(result.account).toEqual([]);
+    });
+    it('should return an empty accountStatus array', () => {
```
ttbarnes, line 11: Missing empty line

ttbarnes: Maybe these two tests could just be one "should return empty arrays"

src/api/helpers/map-unverified-accounts/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        ...mockAccount,
+        id: '1',
+      },
+      {
+        ...mockAccount,
+        id: '2',
+      },
+    ];
+
+    it('should return a populated account array', () => {
```
ttbarnes, line 30: mapped* instead of populated?

src/api/helpers/map-unverified-accounts/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        {
+          where: { id: accounts0.id },
+          data: {
+            updatedAt: new Date(),
+          },
+        },
+        {
+          where: { id: accounts1.id },
+          data: {
+            updatedAt: new Date(),
```
ttbarnes, line 45: Will we not have the same issue as fixed in another file, RE the dates.

src/api/helpers/map-unverified-accounts/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+          },
+        },
+      ];
+
+      expect(result.account).toEqual(expectedAccounts);
+    });
+
+    it('should return a populated accountStatus array', () => {
```
ttbarnes, line 53: mapped*

src/api/helpers/map-unverified-accounts/index.ts
```diff
@@ -0,0 +1,47 @@
+import { Account, CronAccountStatusUnverifiedUpdate, CronAccountUpdate } from '../../types';
+
+/**
+ * mapInactiveApplications
```
ttbarnes, line 4: typo

src/api/helpers/map-unverified-accounts/index.ts
```diff
@@ -0,0 +1,47 @@
...
+ * generates array of objects for updating the db
+ * object contains where object with id of application to be updated
+ * contains data object with fields which are updated
+ * @param {Array<CronAccountStatusUnverifiedUpdate>} applications: array of application ids and statuses
+ * @returns {Array<CronAccountStatusUnverifiedUpdate>}: Arrays of objects for updating db
```
ttbarnes, line 9: Documentation seems to be copied/pasted

src/api/helpers/update-unverified-accounts/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * if unverified accounts
+     * then maps through them to create update objects
+     * and updates account status to isInactive
+     */
+    if (accounts.length) {
+      await mapAndUpdateUnverifiedAccounts(accounts, context);
+    }
+
```
ttbarnes, line 29: Could we add some logs here, for when the array is populated, and when it's not?

src/api/helpers/update-unverified-accounts/index.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Sets inactive flag to true and updates updatedAta
+ * returns success flag
+ * @param {Context} KeystoneJS context API
+ * @returns {Promise<SuccessResponse>} success flag
+ */
+const updateUnverifiedAccounts = async (context: Context): Promise<SuccessResponse> => {
+  try {
+    console.info('Getting and updating unverified accounts');
+
+    // gets inactive applications
```
ttbarnes, line 18: typo

src/api/test-helpers/accounts.ts
```diff
@@ -97,10 +97,35 @@ const create = async ({ context, data, deleteAccounts = true }: TestHelperAccoun
...
+
+    const account = (await context.db.Account.updateOne({
+      where: { id: accountId },
+      data,
+    })) as Account;
+
+    return account;
+  } catch (err) {
+    console.error(err);
+    throw new Error(`Getting an account by ID (test helpers) ${err}`);
```
ttbarnes, line 120: "Getting" should be "Updating"

src/api/test-helpers/create-multiple-accounts.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Account, Context } from '../types';
+
+/**
+ * createMultipleAccounts
+ * Create multiple accounts for unit testing
+ * @param {Object} KeystoneJS context API
+ * @returns {Promise<Array<Account>>} Accounts array
+ */
+export const createMultipleAccounts = async (context: Context): Promise<Array<Account>> => {
+  const account1 = (await accounts.create({ context })) as Account;
```
ttbarnes, line 11: What's the reason for _not_ deleteing accounts in the first call here like the others?

ttbarnes: Could `accounts.create` have a default `deleteAccounts` param of false? Save passing `false` to it multiple times here.

src/api/helpers/get-unverified-accounts/index.ts
Zainzzkk: If an account is already inActive, then it should not be returned as will waste resources.  And also I thought if an account is blocked, then should not also be set to isInactive

src/api/test-helpers/create-multiple-accounts.ts
Zainzzkk: accounts.create is being used in 40+ other places and I think they need a delete to be set to default true

src/api/helpers/update-unverified-accounts/index.ts
Zainzzkk: If populated, there is a console.info in `mapAndUpdateUnverifiedAccounts`

src/api/helpers/get-unverified-accounts/index.ts
ttbarnes: `isVerified` and `isInactive` makes sense, I don't think `isBlocked` should be included here.

A user could become blocked by trying to sign in a few times, and then abandon trying to sign in. CRON job could handle this

src/api/test-helpers/create-multiple-accounts.ts
ttbarnes: Understood, makes sense as it is. This seems to be the only place where we need to _not_ delete accounts 👍

src/api/test-helpers/create-multiple-accounts.ts
```diff
@@ -0,0 +1,21 @@
...
+  const account2 = (await accounts.create({ context, deleteAccounts: false })) as Account;
+  const account3 = (await accounts.create({ context, deleteAccounts: false })) as Account;
+  const account4 = (await accounts.create({ context, deleteAccounts: false })) as Account;
+  const account5 = (await accounts.create({ context, deleteAccounts: false })) as Account;
+  const account6 = (await accounts.create({ context, deleteAccounts: false })) as Account;
+
+  return [account1, account2, account3, account4, account5, account6];
```
ttbarnes, line 18: Can we have 0 based index instead of 1?

src/api/helpers/get-unverified-accounts/index.ts
ttbarnes: Not a blocker - can be done in the next PR, could we add some documentation about why we check that `isVerified` is false?

ttbarnes (approved):

PR #2396, opened 2024-05-13
fix(EMS-3299): No PDF - Export contract - Change your answers - Using agent - Yes to no

ttbarnes:
## Introduction :pencil2:
This PR fixes an issue where, when changing the "Using agent" question from "yes" to "no", some agent data would not be deleted.

- Create new cypress commands to assert empty agent field values:
   - `assertEmptyAgentChargesFieldValues`
   - `assertEmptyAgentDetailsFieldValues`
   - `assertEmptyAgentServiceFieldValues`
- Update E2E tests to go through all agent forms (after changing from "yes" to "no") and run the above commands.
- Update `mapAndSave.exportContractAgent` to conditionally nullify agent service and agent charges data.
- Update `save.exportContractAgentService` to _not_ strip empty form fields.
- Create `nullifyAgentServiceData` helper function.
- Create `shouldNullifyAgentServiceData` helper function.

- Bump `cspell` dependency.
- Fix some typos.

### DIFF
--- a/src/ui/server/controllers/insurance/export-contract/map-and-save/export-contract-agent/index.ts
+++ b/src/ui/server/controllers/insurance/export-contract/map-and-save/export-contract-agent/index.ts
@@ -1,11 +1,18 @@
 import hasFormData from '../../../../../helpers/has-form-data';
 import mapSubmittedData from '../../map-submitted-data/agent';
-import save from '../../save-data/export-contract-agent';
+import saveAgent from '../../save-data/export-contract-agent';
+import saveAgentService from '../../save-data/export-contract-agent-service';
+import saveAgentServiceCharge from '../../save-data/export-contract-agent-service-charge';
+import shouldNullifyAgentServiceData from '../../../../../helpers/should-nullify-agent-service-data';
+import nullifyAgentServiceData from '../../../../../helpers/nullify-agent-service-data';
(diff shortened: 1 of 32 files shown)


src/ui/server/controllers/insurance/export-contract/map-and-save/export-contract-agent/index.ts
```diff
@@ -14,20 +21,49 @@ import { Application, RequestBody, ValidationErrors } from '../../../../../../ty
...
 
       if (!saveResponse) {
         return false;
       }
 
+      /**
+       * If AGENT_SERVICE data should be nullified,
+       * Nullify and save all AGENT_CHARGES and AGENT_SERVICE data.
+       */
+      if (shouldNullifyAgentServiceData(isUsingAgent, application.exportContract.agent)) {
```
Zainzzkk, line 44: could this go in a helper? will make this file shorter

Zainzzkk: as in the bit which sets service and serviceCharge to null

ttbarnes: Good shout. Moved into a new function.

Need to do the same for "Loss payee" - one for later.

Zainzzkk (approved):

PR #2461, opened 2024-05-22
feat(EMS-3346): no PDF - application submission - XLSX - key information, eligibility

ttbarnes:
## Introduction :pencil2:

This PR updates the XLSX mapping sent during application submission so that the following sections are aligned with the latest no PDF/XLSX design:
- Key information
- Eligibility

- Rename and move some `XLSX_CONFIG` `INDEXES` definitions into their own functions:
  - `TITLE_INDEXES`
  - `INDEXES`
- Update XLSX field content strings.
- Update API field and links content strings.
- Remove now unused helper functions previously used for some eligbility XLSX field mappings.
- Re-order `mapApplicationToXLSX` sections. Eligibility is now more towards the top.
- Update `mapEligibility` so that:
  - The order amatches the latest design.
  - A Companies house number is included.
  - The `TOTAL_CONTRACT_VALUE` answer is now mapped via `MORE_THAN_250K`.
  - The `HAS_END_BUYER` question/answer is now included.
- Rename `mapKeyInformation` to `mapIntroduction`.
- Remove now unused `mapSecondaryKeyInformation`.
- Add `APPLICANT_ROLE` question/answer to `mapIntroduction`.

- Update E2E eligibility fixture to have a `false` value for `HAS_END_BUYER`. The previous `true` value was incorrect and not representitive of the real world user flow. Note: this fixture is used to quickly create an application during some E2E tests.


src/api/content-strings/XLSX.ts
Zainzzkk, line 1: Are some of these the same as other content strings?

ttbarnes: They're not, these are specific content strings for the XLSX, that differ from the field content strings.

The XLSX uses a mixture of both. Some field titles/labels need to have different content for internal teams, compared to when a user is completing the fields.

src/api/generate-xlsx/map-application-to-XLSX/map-key-information/index.test.ts
```diff
@@ -1,27 +1,43 @@
...
+  },
+  POLICY: {
+    TYPE_OF_POLICY: { POLICY_TYPE },
+  },
+} = FIELD_IDS;
 
 describe('api/generate-xlsx/map-application-to-xlsx/map-key-information', () => {
-  it('should return an array of mapped key information', () => {
-    const result = mapPolicy(mockApplication);
+  it('should return an array of mapped buyer fields', () => {
```
Zainzzkk, line 30: is this right?

ttbarnes: Good spot, updated

src/api/content-strings/XLSX.ts
ttbarnes: I've added some documentation to explain this 👍

Zainzzkk (approved):

PR #2471, opened 2024-05-23
feat(EMS-3345): no PDF - application submission - XLSX - exporter business

ttbarnes:
## Introduction :pencil2:

This PR updates the XLSX mapping sent during application submission so that the "Exporter business" section is aligned with the latest no PDF/XLSX design:

- Update various cypress commands to consume and pass a `differentTradingName` param.
- Add new application submission E2E tests for single and multiple policies for the following scenarios:
  - Using a broker, different trading address.
  - Using a broker, different trading name.
  - Using a broker, different trading address, different trading name.
  - Not using a broker, different trading address.
  - Not using a broker, different trading name.
  - Not using a broker, different trading address, different trading name.
- Extract `XLSX_CONFIG` `TITLE_INDEXES` and `INDEXES` into their own files.
- Create new `incrementIndexes` function.
- Update XLSX_CONFIG` to conditionally increment indexes if:
  - "Exporter business" has different name.
  - "Exporter business" has different trading address.
  - Or both of above.
- Update XLSX content strings.
- Update `mapApplicationToXLSX` ordering so that the "Exporter business" section is towards the top of the XLSX.
- Update `getPopulatedApplication` to conditionally create a `CompanyDifferentTradingAddress` relationship.
- Create new `companyScenarios` test mocks.
- Rename `mapExporter` XLSX function to `mapExporterBusiness` and move all mapping logic into their own functions.

- Align API `EXPORER_BUSINESS` field IDs.
- Split up `getPopulatedApplication` error handling unit tests into their own test.
- Fix some linting issues.
- Fix some typos.


src/api/generate-xlsx/map-application-to-XLSX/map-exporter-business/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    xlsxRow(FIELDS[HAS_CREDIT_CONTROL], mapYesNoField(business[HAS_CREDIT_CONTROL])),
+
+    xlsxRow(FIELDS[COMPANY_SIC], mapSicCodes(companySicCodes)),
+
+    xlsxRow(CONTENT_STRINGS[FINANCIAL_YEAR_END_DATE].SUMMARY?.TITLE, financialYearEndDate),
+
+    ...mapBroker(application),
+  ];
+
```
ttbarnes, line 143: TODO: address comments in this file. Split up functions.

src/api/constants/XLSX-CONFIG/index.ts
```diff
@@ -1,78 +1,67 @@
...
     indexes.BUYER_CONTACT_DETAILS += 1;
   }
 
   if (broker[USING_BROKER]) {
+    indexes.TITLES.POLICY += 3;
     indexes.TITLES.BUYER += 3;
     indexes.TITLES.DECLARATIONS += 3;
+    indexes.BUYER_ADDRESS += 3;
+
+    indexes.BROKER_ADDRESS = 48;
```
Zainzzkk, line 50: could this be a constant?

src/api/generate-xlsx/map-application-to-XLSX/map-exporter-business/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  NATURE_OF_YOUR_BUSINESS: { GOODS_OR_SERVICES, YEARS_EXPORTING, EMPLOYEES_UK },
+  TURNOVER: { ESTIMATED_ANNUAL_TURNOVER, PERCENTAGE_TURNOVER },
+  HAS_CREDIT_CONTROL,
+} = FIELD_IDS;
+
+/**
+ * mapExporterBusiness
+ * Map an application's exporter business fields into an array of objects for XLSX generation
+ * @param {Application}
+ * @returns {Array<object>} Array of objects for XLSX generation
```
Zainzzkk, line 38: capital O in object

src/api/generate-xlsx/map-application-to-XLSX/map-exporter-business/map-different-trading-address/index.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * mapDifferentTradingAddress
+ * Generate an XLSX row if a company has a "different trading address" value.
+ * @param {ApplicationCompany} company: Application company
+ * @returns {Object | undefined} xlsxRow
+ */
+const mapDifferentTradingAddress = (company: ApplicationCompany) => {
+  const { differentTradingAddress } = company;
+
+  const differentTradingAddressValue = differentTradingAddress[FULL_ADDRESS];
```
Zainzzkk, line 21: do we need  this line or could we just do `if ( differentTradingAddress[FULL_ADDRESS]) {`

ttbarnes: It's to be DRY, otherwise we'd have `differentTradingAddress[FULL_ADDRESS]` in two places (line 23 and 24)

src/api/constants/XLSX-CONFIG/index.ts
ttbarnes: I'm not sure about this. I think it would make it more confusing, because it's a conditional index, what could the constants be called? There's a few instances of this.

E.g  I think it would be harder to read/follow, compared to having it hard coded here. What do you think?

```ts
indexes.BROKER_ADDRESS = USING_BROKER_BROKER_ADDRESS_INDEX;
indexes.ALTERNATIVE_TRADING_ADDRESS = HAS_DIFFERENT_TRADING_ADDRESS_ALTERNATIVE_ADDRESS_INDEX;

// another condition
indexes.ALTERNATIVE_TRADING_ADDRESS = HAS_DIFFERENT_TRADING_NAME_AND_ADDRESS_ALTERNATIVE_ADDRESS_INDEX;
```

Zainzzkk: Just thought would be easier where would only need to change it in one place.  But this willbe fine as is

ttbarnes: More than likely to be other instances of this in other sections in the next few tickets, can have another think about it then

PR #2500, opened 2024-05-31
fix(EMS-3385): no PDF -  policy - nullify data when changing policy type

ttbarnes:
## Introduction :pencil2:

This PR fixes an issue where, when changing the policy type from "single" to "multiple" (or vice versa), certain "single" or "multiple" specific policy data would not be wiped.

  - `changePolicyTypeToSingleAndSubmitContractPolicyForm`
  - `assertCurrencyFormFieldsAreEmpty`
  - `assertEmptyContractCompletionDateFieldValues`
  - `assertEmptyRequestedStartDateFieldValues`
- Add an E2E test for changing the policy type from "multiple" to "single", then back to "multiple".
- Add an E2E test for changing the policy type from "single" to "multiple" then back to "single".
- Update `mapAndSave.policy` to pass `application` to `mapSubmittedData`.
- Split up policy data saving mapping functions into their own files/functions:
  - `mapDateFields`
  - `nullifyGenericContractPolicyFields`
  - `nullifyMultipleContractPolicyFields`
  - `nullifySingleContractPolicyFields`
- Update `saveData.policy` to not strip empty form fields.
- Create new helper function to determine if a policy type has changed, `hasPolicyTypeChanged`.

- Fix some linting issues.


src/ui/server/controllers/insurance/policy/map-submitted-data/policy/date-fields/index.test.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    day = date.getDate();
+    month = date.getMonth() + 1;
+    year = date.getFullYear();
+  });
+
+  describe(`when ${REQUESTED_START_DATE} day, month and year fields are provided`, () => {
+    it(`should return an object with ${REQUESTED_START_DATE} as a timestamp`, () => {
+      const mockBody = {
+        [`${REQUESTED_START_DATE}-day`]: '1',
```
Zainzzkk, line 29: is there a reason this is 1 and not `day`?

ttbarnes: Updated to use `day` - will push once another PR is merged and this can be updated with `no-pdf`

Zainzzkk (approved):

PR #2501, opened 2024-05-31
feat(EMS-3428):  no PDF - CRON - Application - submissionDeadline

Zainzzkk:
## Introduction :pencil2:

This PR switches to submissionDeadline instead of updatedAt to check if application should be abandoned 

* Changed query to consume submissionDeadline
* Updated test


src/api/helpers/get-inactive-applications/index.ts
```diff
@@ -15,15 +14,13 @@ const getInactiveApplications = async (context: Context): Promise<Application[]>
...
-    const thirtyDaysLimit = getThirtyDaysBeforeNow();
-
     /**
      * queries in progress applications which are older than 30 days from the db
      * returns array of id and status
      */
     const applications = (await context.query.Application.findMany({
       where: {
-        AND: [{ status: { in: [IN_PROGRESS] } }, { updatedAt: { lt: thirtyDaysLimit } }],
+        AND: [{ status: { in: [IN_PROGRESS] } }, { submissionDeadline: { lt: new Date() } }],
```
ttbarnes, line 23: Can we update the documentation to e.g

```ts
/**
  * queries "in progress"/${IN_PROGRESS} applications which have a submissionDeadline that are....
  * returns array of ids and statuses
  */
```

ttbarnes: The "elapsed" part needs more of a description 😛 

Do we need the `getThirtyDaysBeforeNow` helper anymore?

ttbarnes: Nice but sorry the documentation isn't clear, can we mention `IN_PROGRESS` status with all caps, also "has is in the ..." ? i think something like this would be nice:

```ts
/**
 * Queries applications that:
 * - have a status of IN_PROGRESS
 * - have a submission deadline in the past
 */
```

src/api/test-helpers/applications.ts
```diff
@@ -71,10 +71,38 @@ const update = async ({ context, applicationId, data }: TestHelperApplicationUpd
   }
 };
 
+/**
+ * deleteAll test helper
+ * Get all accounts and delete them.
+ * @param {Object} KeystoneJS context API
+ * @returns {Array} Accounts that have been deleted
```
ttbarnes, line 78: Not a blocker, can fix in the next PR - this says accounts instead of applications

ttbarnes (approved):

PR #2636, opened 2024-06-25
fix(EMS-3515): no PDF - XLSX - Policy - Pre credit period

Zainzzkk:
## Introduction :pencil2:

This PR adds a missing row for credit period in the "policy" section of the XLSX sent during application submission.

* Added row for when `NEED_PRE_CREDIT_PERIOD` is true
* Modified indexes
* Added tests


src/api/constants/XLSX-CONFIG/index.ts
```diff
@@ -124,6 +125,16 @@ export const XLSX_ROW_INDEXES = (application: Application): XLSXRowIndexes => {
...
+  if (needPreCreditPeriodCover) {
+    indexes.BROKER_ADDRESS += 1;
+    indexes.BUYER_ADDRESS += 1;
+    indexes.LOSS_PAYEE_ADDRESS += 1;
+
+    indexes.TITLES.BUYER += 3;
+    indexes.TITLES.DECLARATIONS += 1;
+    indexes.TITLES.EXPORT_CONTRACT += 1;
+  }
+
```
ttbarnes, line 137: Did you run any tests and check the XLSX for any other scenarios? It looks like we also need to bump the index for `AGENT_ADDRESS`

Would be good to look at the XLSX with these changes, when fully populated, with a single and multiple policy

ttbarnes (approved):

PR #2641, opened 2024-06-25
fix(EMS-3520): no pdf - dashboard - policy currency code

ttbarnes:
## Introduction :pencil2:

This PR fixes an issue in the dashboard where the "value" column always render with a GBP currency code, when it should use the currency selected when completing the policy section.

- Update some cypress commands to consume and pass a `chooseCurrency` param.
- Add a new E2E test.
- Update "get applications" GQL query to return the policy currency code.
- Update `mapValue` to use the policy currency code. Otherwise, default to GBP.


e2e-tests/insurance/cypress/e2e/journeys/dashboard/dashboard-populated-application-non-gbp-currency.spec.js
```diff
@@ -0,0 +1,113 @@
...
+const {
+  CONTRACT_POLICY: {
+    SINGLE: { TOTAL_CONTRACT_VALUE },
+  },
+  EXPORT_VALUE: {
+    MULTIPLE: { MAXIMUM_BUYER_WILL_OWE },
+  },
+} = FIELD_IDS;
+
+context('Insurance - Dashboard - populated application - non GBP currency', () => {
```
Zainzzkk, line 27: could this say alternative currency instead of non GBP?

ttbarnes: added "alternative", want to keep "non GBP" so that it's clear. We do this in a few places

Zainzzkk (approved):

PR #2675, opened 2024-07-03
fix(EMS-3508-3543): no pdf - application submission - XLSX - agent values

ttbarnes:
## Introduction :pencil2:

This PR fixes 2x issues in the XLSX where:

- The "Agent country" field was rendering a country code instead of the country name.
- The "Agent charge amount" field was not rendering a currency code.

- Update `mapAgentChargeAmount` to:
  - Consume a list of countries
  - Call `getCountryByIsoCode`.
  - Call `formatCurrency` with `FIXED_SUM_AMOUNT` as a number (it's saved as a string).

- Various application submission E2E test improvements.

### DIFF
--- a/e2e-tests/commands/insurance/complete-export-contract-section.js
+++ b/e2e-tests/commands/insurance/complete-export-contract-section.js
@@ -54,7 +54,7 @@ const completeExportContractSection = ({
       });
 
       if (alternativeCurrency) {
-        cy.go('back');
+        cy.clickBackLink();
 
         cy.clickProvideAlternativeCurrencyLink();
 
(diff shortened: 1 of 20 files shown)


e2e-tests/commands/insurance/complete-export-contract-section.js
```diff
@@ -50,6 +52,16 @@ const completeExportContractSection = ({
         fixedSumAmount: agentChargeFixedSumAmount,
         percentageMethod: agentChargeMethodPercentage,
       });
+
+      if (alternativeCurrency) {
+        cy.go('back');
```
Zainzzkk, line 57: cy.clickBackButton?

ttbarnes: We don't have such a command; I think we tried before and causes some strange cypress issues when called within other commands.

ttbarnes: Updated to use `cy.clickBackLink();`

Zainzzkk (approved):

PR #2676, opened 2024-07-03
chore(mysql-upgrade): upgrade mysql version

Zainzzkk:
## Introduction :pencil2:

On the mysql update to 9.0, `--mysql-native-password=ON` is deprecated

* Remove the line


abhi-markan (commented):
Can we tag to a specific version to avoid new MySQL image breaking future releases please.
```bash
image: mysql:9.0.0
```

PR #2680, opened 2024-07-04
feat(EMS-3548): no pdf - post data migration - company data

ttbarnes:
## Introduction :pencil2:

This PR introduces some new redirection and saving logic so that if an MVP application is in progress and is migrated to no-PDF - when starting the "Business" section of the application, the user is redirected to a new "Enter Companies House Number" form.

Without this in place, an MVP application could become "stuck" in no-PDF and unable to be completed.

This is because:
- MVP does not enforce collection of a companies house number before an application is created.
- No PDF enforces collection of a companies house number before an application is created.
- Therefore, an MVP application could be in progress without a Companies House number, which is required in no-PDF.

- Create new API GraphQL resolver, `updateCompanyPostDataMigration`.
- Create new `EXPORTER_BUSINESS` route, controller and template, `ENTER_COMPANIES_HOUSE_NUMBER`.
- Update the `EXPORTER_BUSINESS` "Start" GET controller to render a different `START_NOW_ROUTE` if an application has been migrated and does not have company data.
- Create new `saveData.companyDetailsPostMigration` function.
- Update `mapCompaniesHouseData` helper function to strip `apiError` and `notFound` flags.

- Minor logging improvement.


e2e-tests/content-strings/pages/insurance/business/index.js
```diff
@@ -7,18 +7,14 @@ export const ROOT = {
...
+    ITEMS: ['what your estimated annual turnover is for this year', 'if you have any credit management processes'],
   },
   OUTRO: 'It should only take a few minutes to complete.',
 };
 
-export const COMPANIES_HOUSE_NUMBER = {
+export const ENTER_COMPANIES_HOUSE_NUMBER = {
   ...SHARED,
   PAGE_TITLE: 'Enter your Companies House registration number (CRN)',
-  NO_COMPANIES_HOUSE_NUMBER: 'I do not have a UK Companies House registration number',
```
Zainzzkk, line 21: will this not be removed from the eligibility page?

src/ui/server/graphql/mutations/update-application/companyPostDataMigration.ts
```diff
@@ -0,0 +1,11 @@
+import gql from 'graphql-tag';
+
+const updateCompanyPostDataMigrationMutation = gql`
+  mutation updateCompanyPostDataMigration($id: String!, $company: CompanyInput!) {
+    updateCompanyPostDataMigration(id: $id, company: $company) {
+      success
```
Zainzzkk, line 6: do you not need to also return  apiError, notFound?

Zainzzkk (commented):
Do we have a way to select wrong company selected and to go back tot he search? like we have on eligibility company details?

src/ui/server/graphql/mutations/update-application/companyPostDataMigration.ts
ttbarnes: No, this is a mutation for updating company data in the DB (with data obtained from Companies House)

The Companies house search is already setup in a different query

e2e-tests/content-strings/pages/insurance/business/index.js
ttbarnes: This is an old content string left over in the E2E "business section" content strings specifically; It is not used.

It still remains in eligibility.

PR #2684, opened 2024-07-08
feat(EMS-3533): no pdf - start page/URL redirect

ttbarnes:
## Introduction :pencil2:

This PR updates the "insurance - eligibility" section of a no-pdf application so that the service has a GOV start page, instead  of hosting our own.

- Update various E2E tests and commands to call `cy.navigateToCheckIfEligibleUrl();` instead of `cy.completeStartForm();`
- Remove `START` nunjucks, template, content strings and POST route/controller.
- Update `START` GET controller to redirect to the GOV start page.
- Update `CHECK_IF_ELIGIBLE` GET controller to create a new `insuranceEligibility` session.


src/ui/server/routes/insurance/index.ts
```diff
@@ -27,7 +27,6 @@ const insuranceRouter = express.Router();
 /* eslint-enable @typescript-eslint/ban-ts-comment */
 
 insuranceRouter.get(INSURANCE_ROUTES.START, startGet);
```
Zainzzkk, line 29: do we need to remove the get?

Zainzzkk (approved):

PR #2705, opened 2024-07-10
feat(EMS-3547): Azure MySQL Flexible server version set

abhi-markan:
## Introduction :pencil2:
Ensure MySQL server on Azure version can be explicitly specified and matches the local docker tagged image version.

* Added `--version` flag to `az mysql flexible-server create` command, with value set to latest supported version by [Azure](https://learn.microsoft.com/en-us/azure/mysql/concepts-version-policy).

### DIFF
--- a/.github/workflows/infrastructure.yml
+++ b/.github/workflows/infrastructure.yml
@@ -261,6 +261,7 @@ jobs:
             --storage-size ${{ vars.DB_STORAGE_SIZE_GB }} \
             --sku-name ${{ env.DB_SKU }} \
             --tier ${{ env.DB_TIER }} \
+            --version ${{ vars.DB_VERSION }} \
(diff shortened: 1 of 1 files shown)


.github/workflows/infrastructure.yml
```diff
@@ -261,6 +261,7 @@ jobs:
             --storage-size ${{ vars.DB_STORAGE_SIZE_GB }} \
             --sku-name ${{ env.DB_SKU }} \
             --tier ${{ env.DB_TIER }} \
+            --version ${{ vars.DB_VERSION }} \
```
Zainzzkk, line 264: is this set to 8.0.36 where the vars comes from?

abhi-markan: Yes, it comes from GitHub

Zainzzkk (approved):

PR #2708, opened 2024-07-10
chore(remove-script-prefix): package.json - Remove --prefix

Zainzzkk:
## Introduction :pencil2:

This PR removes --prefix from package.json and replaces it with cd commands

* Removed --prefix from package.json scripts
* Added cd to directories instead
* Fixed linting error in api directory


src/api/types/test-helpers/index.ts
```diff
@@ -1,11 +1,13 @@
 import Context from '@keystone-6/core/types';
+/* eslint-disable */
```
abhi-markan, line 2: Can you please add a comment why we are done this and a `TODO` on a long term fix on the ticket please.

abhi-markan (changes requested):
One minor comment please.

src/api/types/test-helpers/index.ts
Zainzzkk: done - no TODO as not sure if it will ever be fixed

ttbarnes (changes requested):
Can we please hold off of merging this until we have `no-pdf` merged into `main`.

Otherwise will have to merge `no-pdf` into `main` again. Will be quicker to have this branch/PR rebased, or apply the changes in a new branch off of `main`

ttbarnes (approved):

abhi-markan (approved):

PR #2720, opened 2024-07-11
fix(EMS-3560): update az cli to v2

abhi-markan:
## Introduction :pencil2:
Update Azure CLI GHA plugin to `v2`

* Updated `azure/cli` to `v2`.
* Refactored `lint.yml` and `sca.yml`.
* Added two distinct DNS domains names for public Azure DNS Zones.

* Minor lint fixes
* Updated `szenius/set-timezone` plugin to `v2`


.github/workflows/lint.yml
```diff
@@ -125,7 +125,64 @@ jobs:
...
+      - name: Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ vars.NODE_VERSION }}
+
+      - name: Dependencies
+        working-directory: ./
+        run: npm run ci:all
+
+      # TODO: Enable post TSC error resolution
```
ttbarnes, line 157: Do we need a ticket for this?

abhi-markan: Yes we do, would you like me to add ticket numbers to all TODO?

.github/workflows/deployment.yml
```diff
@@ -73,21 +73,21 @@ jobs:
           creds: ${{ secrets.AZURE_CREDENTIALS }}
 
       - name: Defaults ✨
-        uses: azure/cli@v2.0.0
+        uses: azure/cli@v2
```
Zainzzkk, line 76: could this not run the risk of updating to the latest version which could have breaking changes? my mysql?

abhi-markan: It shouldn't do as this is just Azure CLI version update, if there is a breaking change then pipelines will pick it up.

.github/workflows/lint.yml
abhi-markan: Have added ticket numbers to `TODO`s. Thanks

Zainzzkk (approved):

PR #2723, opened 2024-07-12
feat(EMS-3083): iterations - Loss Payee International - BIC Swift Code - updated validation

Zainzzkk:
## Introduction :pencil2:
This PR updates the BIC_SWIFT_CODE validation to allow alpha numeric, alpha only but NOT numeric only

* Updated validation to use joi
* Matches alpha numeric
* Matches numberic only
* Rejects if error for alpha numeric or if numeric only has no errors
* Updated tests


src/ui/server/constants/regex.ts
```diff
@@ -14,6 +14,13 @@ export const REGEX = {
    */
   INCLUDES_NUMERICAL_CHARACTERS: /\d+/g,
 
+  /**
+   * NUMERICAL_CHARACTERS:
+   * Regex that checks a string includes 0-9 characters without the global flag.
+   * Note: \d is exactly the same as [0-9]
+   */
+  NUMERICAL_CHARACTERS: /^\d+$/,
```
ttbarnes, line 22: I think this should be called `ONLY_NUMERICAL_CHARACTERS`, and documentation updated to say "only" instead of "includes", because this only checks if a string has pure numerical characters rather than if a string includes numerical characters

src/ui/server/shared-validation/alpha-numerical-characters-only/index.ts
```diff
@@ -17,33 +18,29 @@ import generateValidationErrors from '../../helpers/validation';
...
 
   /**
-   * Otherwise, the provided fieldValue is invalid.
-   * Return validation errors.
+   * if the field value does not match the alpha numerical characters validation
+   * or if the field value does not contain an error for only numerical values (string is numbers only)
+   * then return generate and return validation errors
    */
-  return generateValidationErrors(fieldId, errorMessage, errors);
+  if (validationAlphaNum.error || !validationNumbersOnly.error) {
```
ttbarnes, line 39: Why does `validationNumbersOnly` have a `!`?

Zainzzkk: in the documentation above.  If validationNumbersOnly does not return a number - means that value is a number so invalid

ttbarnes: Finding this hard to read/understand even with the comments and re-reading, any way we can make it clearer?

ttbarnes: I think it's this part that could be clearer:

```ts
!validationNumbersOnly.error
```

It reads "if there is no error" ... return an error

ttbarnes (approved):

PR #2744, opened 2024-07-16
chore(EMS-3559): iterations - SonarCloud intentionality errors

Zainzzkk:
## Introduction :pencil2:
This PR fixes Sonarcloud intentionality errors

* Various changes to codebase for intentionality


src/ui/server/helpers/string/index.test.ts
```diff
@@ -100,15 +100,15 @@ describe('server/helpers/string', () => {
...
+        const result = stringsAreDefined('test');
 
         expect(result).toEqual(false);
       });
     });
 
     describe('both strings are undefined', () => {
       it('should return false', () => {
-        const result = stringsAreDefined(undefined, undefined);
+        const result = stringsAreDefined();
```
ttbarnes, line 111: Seems that we can remove the test on line 49 with this change

src/ui/types/index.d.ts
```diff
@@ -85,6 +85,10 @@ import {
 import { InsuranceFeedbackVariables } from './feedback';
 import { SanitiseValueObjParams } from './sanitise-value';
 
+type FieldsWithValues = {
+  [key: string]: string;
+};
```
ttbarnes, line 90: Could we use `ObjectType` instead? Few instances of `ObjectType` declared, will be a nice and quick DRY change 😎

src/ui/server/helpers/string/index.test.ts
Zainzzkk: that is a different function?

src/ui/types/index.d.ts
ttbarnes: Can we move this into its own file like the other types?

Happy to do after this is merged

ttbarnes (approved):

PR #2756, opened 2024-07-17
chore(EMS-3467): iterations - move total contract value threshold flag into the API

ttbarnes:
## Introduction :pencil2:
This PR moves the `totalContractValueOverThreshold` flags/conditions into the API.

This means that there is one single instance of this logic and is owned by the "get populated application" API helper function.

- Update various XLSX mapping functions to consume an application instead of nested application objects.
- Move `mapTotalContractValueOverThreshold` UI helper function into the API, simplify the logic.
- Update API types and test mocks.
- Update the API's `getPopulatedApplication` function to return a `totalContractValueOverThreshold` flag.
- Remove the UI's "get application" GQL query and middleware. Now, the UI only uses "get application by reference number".

- Fix/update a unit test description.


src/api/generate-xlsx/map-application-to-XLSX/map-buyer/map-previous-cover-with-buyer/index.test.ts
```diff
@@ -47,7 +49,11 @@ describe('api/generate-xlsx/map-application-to-xlsx/map-buyer/map-previous-cover
           exporterHasPreviousCreditInsuranceWithBuyer: false,
         };
 
-        const result = mapPreviousCoverWithBuyer(eligibility, mockRelationship);
+        const application = mockApplicationSinglePolicyTotalContractValueOverThreshold;
+
+        application.buyer.relationship = mockRelationship;
```
Zainzzkk, line 54: could this and the line above go in a beforeEach?

src/api/generate-xlsx/map-application-to-XLSX/map-export-contract/map-private-market/index.test.ts
```diff
@@ -49,7 +46,11 @@ describe('api/generate-xlsx/map-application-to-xlsx/map-private-market', () => {
           [ATTEMPTED]: false,
         };
 
-        const result = mapPrivateMarket(mockPrivateMarket, mockTotalContractValue);
+        const application = mockApplicationSinglePolicyTotalContractValueOverThreshold;
+
+        application.exportContract.privateMarket = mockPrivateMarket;
```
Zainzzkk, line 51: same as above

src/api/generate-xlsx/map-application-to-XLSX/map-buyer/map-previous-cover-with-buyer/index.test.ts
ttbarnes: We can only move the `application` const above to be DRY. The `relationship` is specific to each test and has to be updated in the applicaiton after relationship stubbing in each test.

Zainzzkk (approved):

PR #2766, opened 2024-07-17
chore(EMS-3465): dry cypress "click link" commands

ttbarnes:
## Introduction :pencil2:
This PR introduces 3x new cypress commands to be DRY.

  - `cy.clickHeaderApplicationsLink();`
  - `cy.clickSignInButtonLink();`
  - `cy.clickSignInResetPasswordLink();`
- Update E2E tests.

- Update `cspell`.


e2e-tests/insurance/cypress/e2e/journeys/account/password-reset/account-password-reset-retry/account-password-reset-retry-maximum-retries.spec.js
```diff
@@ -9,70 +7,73 @@ const {
...
+         * so that we can assert that the URL has the correct account ID.
+         */
+        const accountEmail = Cypress.env('GOV_NOTIFY_EMAIL_RECIPIENT_1');
 
-      cy.getAccountByEmail(accountEmail).then((responseData) => {
-        const [firstAccount] = responseData;
-        account = firstAccount;
+        cy.getAccountByEmail(accountEmail).then((responseData) => {
+          const [firstAccount] = responseData;
+          account = firstAccount;
```
Zainzzkk, line 54: could we have a let accountUrl and construct it here then we don't need to construct it twice?

ttbarnes: This wasn't changed in this PR, this is just eslint/prettier, but i've updated this so there is an `expectedUrl` const.

Zainzzkk (approved):

PR #2778, opened 2024-07-19
feat(EMS-3535): Declarations - Anti-bribery and Corruption - Content changes

Zainzzkk:
## Introduction :pencil2:
This PR adds content changes to the Anti-bribery and Corruption page

* Updated definitions in content-strings for declarations
* Updated test content-strings


ttbarnes (changes requested):
Needs to be a new version

Also need to look at data migration and see how this declarations content version is obtained and declared in the DB.

I can help with this

ttbarnes (approved):

PR #2811, opened 2024-07-24
feat(EMS-1921):  custom domain names for Azure front door

abhi-markan:
## Introduction :pencil2:
Ensure custom domains are created during IaC execution.
⚠️ Please not `domains` job will fail unless SSLs are manually added to KV which has not been covered as part of this PR.

* Managed identity creation at FD level.
* Ensure Key Vault policy to allow correct permissions for FD to fetch respective certificates.
* Introduction of a new domains job responsible for:
  * Add custom certificates from KV to FD secrets
  * Creation of custom domains with FD secrets.
  * Association of endpoints with respective custom domains.

* Added values to `.env.sample` for API container.


.github/workflows/infrastructure.yml
```diff
@@ -564,6 +564,26 @@ jobs:
             --name IPAllowListRule \
             --policy-name waf${{ env.PRODUCT }}${{ env.TARGET }}${{ vars.VERSION }}
 
+      - name: Identity 🪪
```
Zainzzkk, line 567: emoji not showing for me, or should it be a ?

.github/workflows/infrastructure.yml
```diff
@@ -1058,3 +1078,84 @@ jobs:
...
+            --use-latest-version true \
+            --secret-source $(az keyvault list --query '[].id' -o tsv)/secrets/kv-certificate-quote-latest
+
+            az afd secret create \
+            --name kv-certificate-apply-latest \
+            --profile-name frontdoor-${{ env.PRODUCT }}-${{ env.TARGET }}-${{ vars.VERSION }} \
+            --use-latest-version true \
+            --secret-source $(az keyvault list --query '[].id' -o tsv)/secrets/kv-certificate-apply-latest
+
+      - name: Domains 🌐
```
Zainzzkk, line 1131: we have 2 Domains as name, could one be a different name?

.github/workflows/infrastructure.yml
abhi-markan: Have updated.

.github/workflows/infrastructure.yml
abhi-markan: Have updated.

Zainzzkk (approved):

PR #2856, opened 2024-07-31
feat(EMS-3665): export contract - how was contract awarded - data model

ttbarnes:
## Introduction :pencil2:
This PR introduces a data model/schema updates for the "how was the contract awarded" form in the "export contract" section.

- Create `ExportContractAwardMethod` DB table.
- Update `ExportContract` DB table to have `awardMethod` and `otherAwardMethod` fields.
- Create `EXPORT_CONTRACT_AWARD_METHOD` constants.
- Update keystone schema.

- Fix a typo.


src/api/constants/export-contract-award-method/index.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * List of custom "export contract award method" database IDs and values.
+ */
+export const EXPORT_CONTRACT_AWARD_METHOD = {
+  OPEN_TENDER: {
+    DB_ID: 'eg9qxlqw4edxa8b5mwbybsrfp',
```
Zainzzkk, line 6: is this the best way to do this? or do a db call and get the IDs that way?

ttbarnes: I believe so, we have the same approach with `COVER_PERIOD` and `TOTAL_CONTRACT_VALUE` and `DB_ID`'s.

By doing it this way, we avoid calling the API and DB in multiple scenarios (GET, POST with validation errors, maybe summary lists).

The only difference between the above mentioned constants and `EXPORT_CONTRACT_AWARD_METHOD` is that the `DB_ID` is a CUID whereas the others use incremental numbers (ideally we would update these as well)

Zainzzkk (approved):

Zainzzkk (approved):

PR #2859, opened 2024-07-31
chore(EMS-3660): export contract - how was contract awarded - form fields

ttbarnes:
## Introduction :pencil2:
This PR adds radio form fields to the "how was the contract awarded" form in the "export contract" section.

- Update `EXPORT_CONTRACT` field ID constants.
- Update `EXPORT_CONTRACT` field content strings.
- Update "how was the contract awarded" E2E test.
- Create `CONDITIONAL_OTHER_METHOD_HTML` nunjucks partial.
- Update "how was the contract awarded" GET controller and nunjucks template.


src/ui/templates/insurance/export-contract/how-was-the-contract-awarded.njk
```diff
@@ -16,19 +18,111 @@
...
+
+  {% set otherMethodHtml %}
+    {% include CONDITIONAL_OTHER_METHOD_HTML %}
+  {% endset %}
 
   <form method="POST" data-cy="form" enctype="application/x-www-form-urlencoded" novalidate>
 
     <input type="hidden" name="_csrf" value="{{ csrfToken }}">
 
+    {{ govukRadios({
```
Zainzzkk, line 33: could this be in its own component/partial and imported in?

ttbarnes: Yes, but I don't see much value in it, as it's specific to this pagea and won't be used anywhere else. Think it would overcomplicate it?

Zainzzkk: just thinking since the file is pretty big with this in, to make it slightly more readable

Zainzzkk (approved):

Zainzzkk (approved):

PR #2882, opened 2024-08-01
chore(EMS-3708): code scanning - Server-side URL redirect

Zainzzkk:
## Introduction :pencil2:
This PR fixes the Server-side URL redirect warning

* Added helper to check originalUrl is valid
* Added REGEX pattern
* Updated redirects function to check url and redirect to PROBLEM_WITH_SERVICE if invalid
* Added tests


src/ui/server/constants/regex.ts
```diff
@@ -30,6 +30,20 @@
...
+   * - numbers.
+   * - letters.
+   * - a hyphen.
+   * - a question mark.
+   * - a slash.
+   * - a dash.
+   * - &.
+   * Note: \d is exactly the same as [0-9]
+   */
+  ALPHA_NUMBER_HYPHEN_DASH_QUESTION_EQUALS_AMPERSAND: /^[A-z-=&?/\d]+$/,
```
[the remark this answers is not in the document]
abhi-markan: @Zainzzkk warning from CodeQL

src/ui/server/controllers/redirects/index.ts
```diff
@@ -1,15 +1,28 @@
...
  * @returns {Express.Response.redirect}
  */
 const get = (req: Request, res: Response) => {
-  const newUrl = generateRedirectUrl(req.originalUrl);
+  const { originalUrl } = req;
+
+  const validUrl = isValidReqOriginalUrl(originalUrl);
+
+  if (!validUrl) {
+    return res.redirect(PROBLEM_WITH_SERVICE);
```
abhi-markan, line 22: Might be good add a console.error I think.

src/ui/server/helpers/is-valid-req-original-url/index.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @returns {Boolean}
+ */
+const isValidReqOriginalUrl = (originalUrl: string) => {
+  const joiString = Joi.string();
+
+  const schema = () => joiString.regex(REGEX.ALPHA_NUMBER_HYPHEN_DASH_QUESTION_EQUALS_AMPERSAND).required();
+
+  const validation = schema().validate(originalUrl);
+
+  if (validation.error) {
```
abhi-markan, line 19:
```javascript
return !validate.error;
```

abhi-markan (changes requested):
Few minor comments please.

src/ui/server/controllers/redirects/index.ts
```diff
@@ -1,15 +1,28 @@
...
+import { ROUTES } from '../../constants';
 import { Request, Response } from '../../../types';
 
+const { PROBLEM_WITH_SERVICE } = ROUTES.INSURANCE;
+
 /**
  * get
- * Redirect an MVP_INSURANCE_ROOT URL to an INSURANCE_ROOT URL
+ * Redirect an MVP_INSURANCE_ROOT URL to an INSURANCE_ROOT URL if req.originalUrl is valid
+ * Redirects to PROBLEM_WITH_SERVICE if req.originalUrl is invalid
```
ttbarnes, line 11: Redirect/Rediects, can we remove the "s"

src/ui/server/helpers/is-valid-req-original-url/index.ts
```diff
@@ -0,0 +1,26 @@
...
+    return false;
+  }
+
+  return true;
+};
+
+export default isValidReqOriginalUrl;
```
ttbarnes, line 26: Feels strange to have this new helper, and an existing `isValidWebsiteAddress` helper function that uses a `isUrl` method from `is-url`

Zainzzkk: url will include https/http and .com etc whereas this is the bit after the url. Needs its own helper

src/ui/server/helpers/is-valid-req-original-url/index.ts
```diff
@@ -0,0 +1,22 @@
+import Joi from 'joi';
+import { REGEX } from '../../constants';
+
+/**
+ * Checks if the provided original URL is valid.
+ * checks against NUMBER_ALPHA_HYPHEN_DASH_QUESTION_EQUALS_AND regex
+ * if error on validation, returns false
+ * returns true if valid
```
ttbarnes, line 8:
```suggestion
  * if error on validation, returns false
  * otherwise, return true
```

src/ui/server/helpers/is-valid-req-original-url/index.test.ts
```diff
@@ -0,0 +1,73 @@
+import isValidReqOriginalUrl from '.';
+
+describe('helpers/is-valid-req-original-url', () => {
+  describe('when provided url is not value', () => {
```
ttbarnes, line 4: "is not value"?

src/ui/server/constants/regex.ts
```diff
@@ -30,6 +30,20 @@ export const REGEX = {
...
+   * - numbers.
+   * - letters.
+   * - a hyphen.
+   * - a question mark.
+   * - a slash.
+   * - a dash.
+   * - &.
+   * Note: \d is exactly the same as [0-9]
+   */
+  ALPHA_NUMBER_HYPHEN_DASH_QUESTION_EQUALS_AMPERSAND: /^[A-Za-z-=&?/\d]+$/,
```
ttbarnes, line 45: I think we should rename this to something like `VALID_REQUEST_ORIGINAL_URL`

src/ui/server/controllers/redirects/index.test.ts
```diff
@@ -12,12 +15,33 @@ describe('controllers/redirects', () => {
...
+      expect(res.redirect).toHaveBeenCalledTimes(1);
+      expect(res.redirect).toHaveBeenCalledWith(expected);
+    });
+  });
+
+  describe('when the req.original is invalid', () => {
+    beforeEach(() => {
+      const invalid = { forgery: true };
+      // @ts-ignore
+      req.originalUrl = invalid;
```
ttbarnes, line 37: Why do we need to do this, with the URL being an object and `ts-ignore`? Express controls `originalUrl`

src/ui/server/controllers/redirects/index.test.ts
```diff
@@ -12,12 +15,33 @@ describe('controllers/redirects', () => {
     res = mockRes();
   });
 
-  it('should redirect to a new URL via generateRedirectUrl helper function', () => {
-    get(req, res);
+  describe('when the req.original is valid', () => {
```
ttbarnes, line 18: Can we rename `req.original` to `req.originalUrl`

src/ui/server/controllers/redirects/index.ts
ttbarnes: Not a blocker (so we can progress PRs), can we change this to the following?:

```ts
Invalid original URL detected in "redirects" controller
```

So that it's more descriptive, otherwise it reads like `/redirects` is route/URL which is not the case

Happy to be done in the next PR.

src/ui/server/helpers/is-valid-req-original-url/index.ts
```diff
@@ -0,0 +1,22 @@
+import Joi from 'joi';
+import { REGEX } from '../../constants';
+
+/**
+ * Checks if the provided original URL is valid.
+ * checks against NUMBER_ALPHA_HYPHEN_DASH_QUESTION_EQUALS_AND regex
+ * if error on validation, returns false
+ * otherwise, return true
+ * @param {string} originalUrl - The original URL to validate.
```
ttbarnes, line 9: Not a blocker, this should be `String`

ttbarnes: I'll fix this in my PR as there's a few other instance

ttbarnes (approved):

abhi-markan (approved):

PR #2893, opened 2024-08-02
chore(max-sockets): lower npm_config_maxsockets to 5

Zainzzkk:
# ⚠️ WIP - Do not merge

## Introduction :pencil2:
This PR fixes an issue with E2E tests where some runners randomly got ERRCONNRESET when running e2e tests.
Reducing value of npm_config_max_sockets should reduce the load of concurrent requests which may reduce the likeliness of this ERRCONNRESET happening again

* Reduce npm_config_max_sockets to 5


.github/workflows/test.yml
```diff
@@ -459,6 +465,9 @@ jobs:
       name: ${{ needs.setup.outputs.environment }}
     runs-on: ubuntu-latest
 
+    env:
+      npm_config_maxsockets: 5  # Increased maxsockets value
```
abhi-markan, line 469: Can you set this at top `env` level so it can appear only once please.

abhi-markan (changes requested):

Zainzzkk:
We have a workaround for this now

PR #2916, opened 2024-08-07
chore(e2e-coverage): iterations - e2e - removed coverage

Zainzzkk:
## Introduction :pencil2:
This PR removes coverage from e2e tests

* Removes coverage from e2e tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -364,9 +364,6 @@ jobs:
         with:
           node-version: ${{ vars.NODE_VERSION }}
 
-      - name: Clear npm cache
-        run: npm cache clean --force
-
       - name: Dependencies
         working-directory: ./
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -432,6 +428,9 @@ jobs:
         with:
           node-version: ${{ vars.NODE_VERSION }}
 
+      - name: Clear npm cache
```
abhi-markan, line 431: This should no longer be required unless we want to reclaim disk space, latest `NPM` is self healing
Reference: https://docs.npmjs.com/cli/v10/commands/npm-cache

abhi-markan (changes requested):
Nice to remove E2E code coverage.

abhi-markan (approved):

PR #2931, opened 2024-08-09
chore(EMS-3633): data migration - required fields for in progress applications

ttbarnes:
## Introduction :pencil2:
This PR updates various conditions in the service so that any applications that have migrated from V1 to V2 (MVP to No-PDF) and have an "in progress" status, all conditional questions relating to the "contract value threshold" are required.

The reason for this change is:

- In MVP eligibility, we ask if the contract value is over £500k. A user can only proceed if the answer to this question is “no”.
- In No-PDF eligibility, we ask if the contract value is above or below £250k. A user can proceed with either option.
- If the contract value is above £250k, we ask more questions in the application.
- After data migration, it is not possible to determine if migrated MVP applications are specifically above £250k.
  - A user has only answered “it’s less than £500k” - it could be less than £250k.
  - Therefore, without these changes, migrated MVP applications do not ask the additional “above £250k” questions.

By enforcing the additional questions, we ensure that a complete data set is available if an MVP application's contract value is (or isn't) above £250k.

## TODO
- Manual test after running the migration script.

- Update all `totalContractValueOverThreshold` conditions to run the same execution if a new `migratedV1toV2` flag/param is true.
- Update unit tests.
- Update various summary list related functions to consume and pass `migratedV1toV2`.
- Update calls to the relevant summary lists to pass a `migratedV1toV2` flag.

- Update API types to include `migratedV1toV2`.

### DIFF
--- a/src/ui/server/controllers/insurance/check-your-answers/export-contract/index.ts
+++ b/src/ui/server/controllers/insurance/check-your-answers/export-contract/index.ts
@@ -48,7 +48,7 @@ export const get = async (req: Request, res: Response) => {
       return res.redirect(PROBLEM_WITH_SERVICE);
     }
 
-    const { referenceNumber, exportContract, totalContractValueOverThreshold } = application;
+    const { referenceNumber, exportContract, totalContractValueOverThreshold, migratedV1toV2 } = application;
 
     const {
       finalDestinationKnown,
@@ -67,7 +67,7 @@ export const get = async (req: Request, res: Response) => {
 
     const countries = await api.keystone.countries.getAll();
 
-    const summaryList = exportContractSummaryLists(exportContract, totalContractValueOverThreshold, referenceNumber, countries, checkAndChange);
+    const summaryList = exportContractSummaryLists(exportContract, totalContractValueOverThreshold, migratedV1toV2, referenceNumber, countries, checkAndChange);
 
     const exportContractFields = requiredFields({
       totalContractValueOverThreshold,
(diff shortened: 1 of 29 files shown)


src/ui/server/controllers/insurance/check-your-answers/export-contract/index.ts
```diff
@@ -67,7 +67,7 @@ export const get = async (req: Request, res: Response) => {
 
     const countries = await api.keystone.countries.getAll();
 
-    const summaryList = exportContractSummaryLists(exportContract, totalContractValueOverThreshold, referenceNumber, countries, checkAndChange);
+    const summaryList = exportContractSummaryLists(exportContract, totalContractValueOverThreshold, migratedV1toV2, referenceNumber, countries, checkAndChange);
```
Zainzzkk, line 70: with so many variables, would this be better in an object?

ttbarnes: Yes, I thought the same - want to update all summary list functions to be object structured. Really keen to do this but it's not quick and would double/triple the size of this PR.

We need to get these changes in ASAP for data migration testing, before public beta.

Can tackle this and all other summary list functions to be object structured in `main-iterations`

Zainzzkk (approved):

src/ui/server/controllers/insurance/check-your-answers/export-contract/index.ts
ttbarnes: Created EMS-3748 for this

PR #2932, opened 2024-08-09
fix(EMS-3633): data migration - xlsx - eligibility questions

ttbarnes:
## Introduction :pencil2:
This PR fixes an issue where, if an application is migrated from V1 to V2, 3x eligibility answers would render in the spreadsheet with answers in "no".

These 3x fields are not present in V1/MVP, only in V2/No-PDF. Therefore, the XLSX should have empty answers for these fields:
- `HAS_END_BUYER`
- `IS_PARTY_TO_CONSORTIUM`
- `IS_MEMBER_OF_A_GROUP`

- Update XLSX `mapEligibility` function to conditionally map the mentioned fields.
- Update `MapYesNoFieldParams` type.
- Update required eligibility field's `IRRELEVANT_FIELD_IDS`.


src/api/types/helpers/index.ts
```diff
@@ -1,4 +1,4 @@
 export interface MapYesNoFieldParams {
-  answer?: boolean;
+  answer?: boolean | null;
```
Zainzzkk, line 2: do you need the | null if it;s `answer?`

ttbarnes: yes, without this, when calling `mapYesNoField` with an answer that is null, TypeScript complains because it's not a boolean.

Zainzzkk (approved):

PR #2982, opened 2024-08-20
chore(test-retry): e2e - retry if dependencies or docker fail

Zainzzkk:
## Introduction :pencil2:
This PR adds a retry to test.yml if dependencies or docker fail 

* Added 3 retries for each Dependencies and Docker step in test.yml


.github/actions/retry/action.yml
```diff
@@ -0,0 +1,23 @@
...
+    default: "."
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        max_attempts=3
+        attempt_num=1
+        until [ $attempt_num -gt $max_attempts ]
+        do
+          echo "Attempt $attempt_num to run command"
```
abhi-markan, line 16: Please can you remove this.

.github/actions/retry/action.yml
```diff
@@ -0,0 +1,23 @@
...
+  command:
+    required: true
+  working-directory:
+    required: false
+    default: "."
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        max_attempts=3
```
abhi-markan, line 12:
```suggestion
        limit=3
```

.github/actions/retry/action.yml
```diff
@@ -0,0 +1,23 @@
...
+        attempt_num=1
```
abhi-markan, line 13:
```suggestion
        run=1
```

.github/actions/retry/action.yml
```diff
@@ -0,0 +1,23 @@
+name: "Retry Steps"
+inputs:
+  command:
+    required: true
+  working-directory:
+    required: false
```
abhi-markan, line 6: Should this be `true`?

.github/actions/retry/action.yml
```diff
@@ -0,0 +1,23 @@
+name: "Retry Steps"
```
abhi-markan, line 1: Can the directory be renamed from `actions` to `modules` please and file to be `retry.yml` so it can be used for other similar instances?

.github/workflows/test.yml
```diff
@@ -422,11 +435,15 @@ jobs:
           node-version: ${{ vars.NODE_VERSION }}
 
       - name: Dependencies
-        working-directory: ./
-        run: npm run ci:all
-
+        uses: ./.github/actions/retry
+        with:
+          command: npm run ci:all
+          working-directory: ./
```
abhi-markan, line 441: If you set working-directory as `./` as default in the module then you can remove all the working directory parameters.

abhi-markan (changes requested):

.github/actions/retry/action.yml
Zainzzkk: docker step doesn't pass a working-directory so keeping it as false

abhi-markan (approved):

PR #2988, opened 2024-08-22
chore(EMS-3495): various documentation updates, alignments

ttbarnes:
## Introduction :pencil2:
This PR updates the XLSX documentation and fixes some other minor documentation and test description issues.

- Update XLSX README.md
- Various documentation improvements/alignments.
- Fix some test descriptions.
- Remove an unused nunjucks template.

List any additional fixes or improvements.


src/api/XLSX/README.md
```diff
@@ -1,15 +1,13 @@
...
-This directory serves as a temporary storage location for generated XLSX files. These XLSX files are typically created and used during the application submission process.
+This directory acts as a transient storage point for these files, ensuring that they are available for email attachments while being promptly removed to maintain data privacy and efficient use of storage resources. :email:
 
 Here's how this directory is used:
 
-1. **Temporary Storage:** When an application is submitted, the system generates an XLSX file containing relevant data or reports. These files are temporarily stored here.
+1. **Temporary Storage:** When an application is submitted, the system generates an XLSX file containing relevant data or reports. The XLSX file is temporarily stored here.
 
-2. **Email Notification:** The XLSX files are typically sent as attachments in email notifications to the underwriting team or relevant stakeholders for review or further processing.
+2. **Email Notification:** The XLSX file is then sent to the underwriting team via a Notify email for review and further offline processing.
```
Zainzzkk, line 9: I think this would be better as `The XLSX file is then sent by email to the underwriting team via Notify` ...

src/api/helpers/create-a-buyer/index.ts
```diff
@@ -2,7 +2,7 @@ import { Context } from '../../types';
 
 /**
  * createABuyer
- * Create a buyer with:
+ * Create a buyer.
```
Zainzzkk, line 5: we don't use full stops everywhere else, should we remove it?

src/api/XLSX/README.md
ttbarnes: The file is not sent _by_ email though, we call Notify - Notify then creates a link which is added to the email, file has to be downloaded outside of the email, so it's via an email

Zainzzkk: should we say `via a Notify email call` in that case?

Zainzzkk (approved):

PR #2989, opened 2024-08-22
fix(EMS-3787): export contract - agent name with an apostrophe

ttbarnes:
## Introduction :pencil2:
This PR fixes an issue where if an apostrophe is included in the "Export contract - agent name" field, the apostrophe would not be decoded when revisiting to form.

Note - unlike some other fields where special characters are allowed, this name field does not allow any special characters, but an apostrophe is allowed.

- Add E2E test coverage.
- Update `mapNameFields` to include the agent name field.

### DIFF
--- a/src/ui/server/helpers/mappings/map-name-fields/index.test.ts
+++ b/src/ui/server/helpers/mappings/map-name-fields/index.test.ts
@@ -35,7 +35,7 @@ const mockApplicationWithCharacterCodes = {
     ...mockApplication.exportContract,
     agent: {
       ...mockApplication.exportContract.agent,
-      [AGENT_NAME]: `${mockApplication.exportContract.agent}'`,
+      [AGENT_NAME]: mockStringWithCharacterCodes,
     },
   },
   policyContact: {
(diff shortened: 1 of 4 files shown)


src/ui/server/helpers/mappings/map-name-fields/index.test.ts
```diff
@@ -28,6 +31,13 @@ const mockApplicationWithCharacterCodes = {
     ...mockApplication.company,
     [COMPANY_NAME]: mockStringWithCharacterCodes,
   },
+  exportContract: {
+    ...mockApplication.exportContract,
+    agent: {
+      ...mockApplication.exportContract.agent,
+      [AGENT_NAME]: `${mockApplication.exportContract.agent}'`,
```
Zainzzkk, line 38: Rather than a ' at the end, could this be changed to be like O'something so more realistic?

ttbarnes: Good shout, i've updated all agent name test mocks to contain `O'Neill`

Zainzzkk (approved):

PR #3000, opened 2024-08-22
chore(EMS-3789): dry cypress commands - task link clicks

ttbarnes:
## Introduction :pencil2:
This PR introduces some new cypress commands to click the task links in the "All sections" task list.

  - `cy.clickTaskBusiness();`
  - `cy.clickTaskBuyer();`
  - `cy.clickTaskPolicy();`
  - `cy.clickTaskExportContract();`
  - `cy.clickTaskCheckAnswers();`
  - `cy.clickTaskDeclarationsAndSubmit();`
- Update E2E tests.

### DIFF
--- a/e2e-tests/commands/insurance/check-your-answers/complete-and-submit-check-your-answers.js
+++ b/e2e-tests/commands/insurance/check-your-answers/complete-and-submit-check-your-answers.js
@@ -1,9 +1,3 @@
-import partials from '../../../partials';
-
-const { taskList } = partials.insurancePartials;
-
-const task = taskList.submitApplication.tasks.checkAnswers;
-
 /**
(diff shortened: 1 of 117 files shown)


Zainzzkk (commented):
Does it not sound better the other way around? as in clickBusinessTask and the same for the rest?

ttbarnes:
> Does it not sound better the other way around? as in clickBusinessTask and the same for the rest?

I did think about this, I think having a `clickTask` prefix is cleaner and consistent, so in the E2E tests you very quickly know it's "clicking a task"

Especially when considering long task names e.g `clickDeclarationsAndSubmitTask`, I think the prefix is nicer. What do you think?

Zainzzkk (approved):

PR #3027, opened 2024-08-30
chore(deps): bump dependencies

ttbarnes:
## Introduction :pencil2:

- Update dependencies.
- Update relevant intergrity hashes.


src/ui/server/constants/integrity.ts
```diff
@@ -1,7 +1,7 @@
 export const INTEGRITY = {
   JS: 'sha512-BFOXngFJ+9msC1y5Q1LD6AG+LG635vp8Lfv/+HVCUU8w/N8VWH13wdTyZeLGvu4d7zoTz83nfz329ENYfe9x5Q==',
```
abhi-markan, line 2: Is this `main.js`, if yes does SRI for this also needs to be updated give `main.js` has also been updated?

abhi-markan (changes requested):

src/ui/server/constants/integrity.ts
ttbarnes: I don't believe it does @abhi-markan , `main.js` purely imports SCSS, there's no actual client side JS used in this file.

I think we should remove this integrity hash instead

abhi-markan (approved):

src/ui/public/js/govukFrontend.js.LICENSE.txt
```diff
@@ -77,6 +77,12 @@
  * @preserve
  */
 
+/**
```
Zainzzkk, line 80: was this done automatically?

Zainzzkk (approved):

abhi-markan (approved):

src/ui/public/js/govukFrontend.js.LICENSE.txt
ttbarnes: Yes, it's populated by GOV design library.

Zainzzkk (approved):

PR #3101, opened 2024-09-20
feat(EMS-3828-3876): currency of agent charges - save and back, improve e2e assertions

ttbarnes:
## Introduction :pencil2:
This PR adds "save and back" functionality to the "export contract - currency of agent charges" form.

Previously, the common currency forms did not have "save and back" functionality.

Therefore, as part of this feature, the DRY E2E currency form assertion functions needed some refactoring to support DRY "save and back" tests.

- Update the E2E currency form assertion function (`formSubmissionAssertions`) so that:
  - It can be executed with 1x line of code, `formSubmission({}).executeTests();`
  - It handles high level describe blocks (instead of repeating in individual tests.
  - It tests "via continue/submit" and "via save/back" button scenarios.
  - The main file is now an `index` file, with previously existing functions extracted into their own files.
- Update `assertCurrencyFormFields` to conditionally click submit or "Save and back" buttons.
- Rename `completeNonCurrencyFields` to `completeNonCurrencyFieldsFunction`.
- Update E2E tests.
- Create "save and back" UI POST routes for "export contract - currency of agent charges"
- Update "your buyer - currency of late payments" controllers to have a page variables funtion with `SAVE_AND_BACK_URL`.

- Add various missing documentation.
- Fix some typos.


e2e-tests/insurance/cypress/e2e/journeys/export-contract/currency-of-agent-charges/save-and-back.spec.js
Zainzzkk, line 1: empty file?

src/ui/server/controllers/insurance/export-contract/currency-of-agents-charge/save-and-back/index.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  beforeEach(() => {
+    req = mockReq();
+    res = mockRes();
+
+    req.body = mockFormBody;
+
+    mapAndSave.exportContractAgentServiceCharge = mapAndSaveSpy;
+  });
+
+  describe('when the form has data', () => {
```
Zainzzkk, line 39: do we have a test for when there are validation errors and when there aren't validation errors like the other save and back tests?

e2e-tests/insurance/cypress/e2e/journeys/export-contract/currency-of-agent-charges/save-and-back.spec.js
ttbarnes: Whoops, great spot!

ttbarnes: Forgot - as part of this PR, updated the DRY currency form assertions to also test "save and back", so no need for individual `save-and-back` E2E test

src/ui/server/controllers/insurance/export-contract/currency-of-agents-charge/save-and-back/index.test.ts
ttbarnes: Good question, we don't, they are not needed because:

```ts
// save-and-back controller

// ...

if (hasFormData(req.body)) {
  // ...

  mapAndSave.exportContractAgentServiceCharge(); // this handles the validation errors

  return res.redirect();
}

return res.redirect();
```

So that is tested in the `mapAndSave` functions, the "Save and back" controller only cares if there is form data or not, regardless of what state the fields are in

Zainzzkk (approved):

PR #3116, opened 2024-09-25
chore(EMS-3886): update filtering of invalid cis countries

ttbarnes:
## Introduction :pencil2:
- 2x objects returned in the CIS (Country Information System) API response are not countries. These need to be filtered out.
- The API is used by other teams so unfortunately we cannot remove these from the root of the problem.

- Update `INVALID_COUNTRIES` constants.

### DIFF
--- a/src/api/constants/external-apis.ts
+++ b/src/api/constants/external-apis.ts
@@ -20,7 +20,15 @@ export const EXTERNAL_API_DEFINITIONS = {
       NO: 'N',
     },
     NO_COVER: 'Off cover',
-    INVALID_COUNTRIES: ['EC Market n/k', 'Non EC Market n/k', 'Non UK', 'Third Country', 'Eastern and Southern African Trade and Development Bank'],
+    INVALID_COUNTRIES: [
+      'CABEI',
+      'Cor Andino Fom',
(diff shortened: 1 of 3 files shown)


src/api/constants/external-apis.ts
```diff
@@ -20,7 +20,15 @@ export const EXTERNAL_API_DEFINITIONS = {
       NO: 'N',
     },
     NO_COVER: 'Off cover',
-    INVALID_COUNTRIES: ['EC Market n/k', 'Non EC Market n/k', 'Non UK', 'Third Country', 'Eastern and Southern African Trade and Development Bank'],
+    INVALID_COUNTRIES: [
```
Zainzzkk, line 23: do we need to update any tests for this?

Zainzzkk: Are we sure these aren't being used as well?

ttbarnes: Good point RE tests, we don't have any tests other than the filtering function and related functions/unit tests.

I have created a ticket to add some E2E test coverage, EMS-3892.

Zainzzkk (approved):

PR #3123, opened 2024-09-26
chore(EMS-3828): export contract - agent charges - task list logic

ttbarnes:
## Introduction :pencil2:
This PR updates the task list logic so that the `FIXED_SUM_AMOUNT` field is a conditionally required field.

- Update `agentServiceChargeTasks` task list helper function.
- Re-enable all disabled tests.

- Remove instances of `clickAlternativeCurrencyLink`.


e2e-tests/insurance/cypress/e2e/journeys/export-contract/change-your-answers/agent-service/change-your-answers-agent-service-charging-no-to-yes.spec.js
```diff
@@ -72,10 +72,9 @@ context(
           const expectedUrl = `${agentChargesUrl}#${FIELD_ID}-label`;
           cy.assertUrl(expectedUrl);
 
-          // TODO: EMS-3828 - renable
           cy.completeAndSubmitAgentChargesForm({
-            // fixedSumMethod: true,
-            percentageMethod: true,
+            fixedSumMethod: true,
+            percentageMethod: false,
```
Zainzzkk, line 77: could this be defaulted to false maybe? then we don't need to add this line in some tests?

ttbarnes: It is defaulted to false, but I think we should be explicit, we already do this in several places

Zainzzkk (approved):

PR #3128, opened 2024-09-26
feat(EMS-3893): buyer - currency of late payments - save and back

ttbarnes:
## Introduction :pencil2:
This PR adds "save and back" functionality to the "Buyer - Currency of late payments" form.

- Update `assertCurrencyFormFields` to have a default `hasSaveAndBack` flag of true. This will be removed in an upcoming PR.
- Create new POST route and controller.


e2e-tests/shared-test-assertions/currency-form-fields/index.js
```diff
@@ -34,7 +34,7 @@ export const assertCurrencyFormFields = ({
   hint,
   legend,
   expectedRedirectUrl,
-  hasSaveAndBack,
+  hasSaveAndBack = true,
```
Zainzzkk, line 37: should they all have a save and back? if they do we could get rid of this condition?

src/ui/server/controllers/insurance/your-buyer/currency-of-late-payments/save-and-back/index.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import { Request, Response } from '../../../../../../types';
+import { EUR, mockReq, mockRes, mockSpyPromiseRejection, referenceNumber } from '../../../../../test-mocks';
+
+const {
+  CURRENCY: { CURRENCY_CODE },
+} = INSURANCE_FIELD_IDS;
+
+const { INSURANCE_ROOT, ALL_SECTIONS, PROBLEM_WITH_SERVICE } = ROUTES.INSURANCE;
+
+describe('controllers/insurance/your-buyer/curreny-of-late-payments/save-and-back', () => {
```
Zainzzkk, line 17: typo

src/ui/server/controllers/insurance/your-buyer/currency-of-late-payments/save-and-back/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      const saveResponse = await mapAndSave.buyerTradingHistory(payload, application, validationErrors);
+
+      if (!saveResponse) {
+        return res.redirect(PROBLEM_WITH_SERVICE);
+      }
+    }
+
+    return res.redirect(`${INSURANCE_ROOT}/${referenceNumber}${ALL_SECTIONS}`);
+  } catch (error) {
+    console.error('Error updating application - your buyer - currency of late payments (save and back) %O', error);
```
Zainzzkk, line 49: I know there is a PR for changing these all to %o but could the new pages not have %o so then we change it in less pages later?

e2e-tests/shared-test-assertions/currency-form-fields/index.js
ttbarnes: Yes, the last "save and back" is added in the next PR, which removes all instances of `hasSaveAndBack`: https://github.com/UK-Export-Finance/exip/pull/3129

src/ui/server/controllers/insurance/your-buyer/currency-of-late-payments/save-and-back/index.ts
ttbarnes: Agree - this PR was created after we discovered this needs to change. WIll update.

Zainzzkk (approved):

PR #3181, opened 2024-10-11
chore(deploy): remove feature environment flags

ttbarnes:
## Introduction :pencil2:
- The `feature` environment has been deleted.
- We have some feature related flags in the deployment scripts/YMLs.

- Remove instances of `feature`.

### DIFF
--- a/.github/workflows/deploy.sh
+++ b/.github/workflows/deploy.sh
@@ -25,9 +25,6 @@ if [ -n "$selection" ]; then
         ############### DEVELOPMENT ###############
         if [ "$destination" = "dev" ]; then
             branch=main
-        ############### FEATURE ###############
-        elif [ "$destination" = "feature" ]; then
-            branch=main
         ############### STAGING ###############
         elif [ "$destination" = "staging" ]; then
             branch=dev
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.sh
```diff
@@ -27,7 +27,7 @@ if [ -n "$selection" ]; then
             branch=main
         ############### FEATURE ###############
         elif [ "$destination" = "feature" ]; then
-            branch=main-iterations
+            branch=main
```
Zainzzkk, line 30: are we deleting the feature environment? in which case probably best to totally remove this section

ttbarnes: I think it is already deleted, can you double check in Azure please?

Zainzzkk: it is deleted so this can be deleted @abhi-markan

ttbarnes: Thanks @Zainzzkk i've removed `feature` from all deployment related scripts/YML

abhi-markan (approved):

Zainzzkk (approved):

PR #3276, opened 2024-11-01
chore(deps): bump dependencies

ttbarnes:
## Introduction :pencil2:

- Update dependencies.
- Remove `mysql2` package (EMS-3950)


Zainzzkk (dismissed):
do we need to bump gov front end here as well? like DTFS

Zainzzkk (commented):
and should this be into main or iterations-2?

ttbarnes:
> do we need to bump gov front end here as well? like DTFS

No, EXIPS is already up to date. DTFS is many versions behind

ttbarnes:
> and should this be into main or iterations-2?

It will be easier to keep this in `iterations-2`.

This branch will be the next release. It's very unlikely for us to push `main` to `production` with pure dependency updates.

By keeping dep updates in `iterations-2` it will make merging `main` => `iterations-2` or vice versa easier

ttbarnes:
> > > and should this be into main or iterations-2?
> > 
> > 
> > It will be easier to keep this in `iterations-2`.
> > This branch will be the next release. It's very unlikely for us to push `main` to `production` with pure dependency updates.
> > By keeping dep updates in `iterations-2` it will make merging `main` => `iterations-2` or vice versa easier
> 
> The only issue is dependabot is set to main, and we are making the changes in iterations-2 and dependabot will not see that we've made the change @abhi-markan

Dependabot will only flag new package updates - if a package is updated in `main-iterations-2`, but not in `main`, dependabot will not create the same alert in `main`. There will be no additional dependabot PRs because we're updating in the iterations branch.

This branch should be very short lived.

In the previous iterations branch, we did dependabot updates in `main` instead of `iterations` and it caused several headaches RE merge conflicts and dependabot updates. Whilst this was occuring, `main` was never pushed to production with the dep updates so it was almost pointless _not_ doing it in the iterations branch.

Zainzzkk (approved):

PR #3278, opened 2024-11-01
chore(EMS-3966): update dependabot to run during non working hours

ttbarnes:
## Introduction :pencil2:
- Dependabot runs at a random time. When this is during working hours, it uses a lot of runners.

- Change time to be 3am.

- Update dependabot to run weekly instead of monthly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,7 @@ updates:
   - package-ecosystem: 'npm'
     directory: '/'
     schedule:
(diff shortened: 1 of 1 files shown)


AlexBramhill (approved):

AlexBramhill (approved):

Zainzzkk (commented):
Are we agreed on weekly here or should it be monthly ?

Zainzzkk (approved):

PR #3290, opened 2024-11-07
chore(EMS-3968): rename formToStopAt cypress command params

ttbarnes:
## Introduction :pencil2:
- Some cypress commands that complete X forms in a section, have a param called `formToStopAt`.
- `formToStopAt` is a confusing param because it does not actually stop at the provided form name. Instead, the provided form name is the _last_ form to be submitted.

For example, if the flow is:

1) "Agent details" form
2) "Agent service" form

In the "Agent service" E2E test, we have the following:

```js
cy.completeAndSubmitExportContractForms({ formToStopAt: 'agentDetails' });
```

What actually happens is, the "Agent details" form is submitted and the flow is stopped at the "Agent service" form. It is not "stopping" on the `agentDetails` form.

This is misleading and so it should be renamed.

- Rename `formToStopAt` to `stopSubmittingAfter`


Zainzzkk (commented):
formToStopAt will never be forgotten! :P

Zainzzkk (approved):

PR #3308, opened 2024-11-15
feat(EMS-3998): broker confirm address - enter manually link

ttbarnes:
## Introduction :pencil2:
As part of "Broker - address lookup" functionality, we need to allow a user to enter an address manually via the "confirm address" page.

- Add `ENTER_ADDRESS_MANUALLY` to page content strings.
- Update GET controller/page variables to include `ENTER_ADDRESS_MANUALLY_URL`.
- Update nunjucks template.


e2e-tests/insurance/cypress/e2e/journeys/policy/broker-confirm-address/broker-confirm-address.spec.js
```diff
@@ -97,6 +99,22 @@ context.skip(
           cy.assertUrl(brokerDetailsUrl);
         });
       });
+
+      describe('`enter address manually` link', () => {
+        it('renders', () => {
```
Zainzzkk, line 104: should this be "should render"?

ttbarnes: Updated

PR #3319, opened 2024-11-21
feat(EMS-3996): broker addresses - form validation, redirect

ttbarnes:
## Introduction :pencil2:
This PR adds form validation and a redirect to the "Broker addresses" form

- Add E2E test.
- Create new POST route, controller.
- Update nunjucks template.

- Fix some nunjucks typos.


e2e-tests/insurance/cypress/e2e/journeys/policy/broker-addresses/broker-addresses-page.spec.js
```diff
@@ -69,12 +76,47 @@ context('Insurance - Policy - Broker addresses page', () => {
...
+      beforeEach(() => {
+        cy.navigateToUrl(url);
+        cy.clickSubmitButton();
+      });
+
+      it('should render validation errors', () => {
+        const field = fieldSelector(FIELD_ID);
+
+        cy.checkErrorSummaryListHeading();
+        cy.assertErrorSummaryListLength(1);
```
Zainzzkk, line 100: did we not have a shared command for checking these errors?

ttbarnes: We have a sparsely used`cy.assertFieldErrors`, but it will not work for radios.

Zainzzkk: submitAndAssertRadioErrors?

ttbarnes: Tried - that won't work for this - it relies on a single field ID - the field is called `selectTheAddress` but none of the options are called this because they are dynamic addresses.

We have this in "Type of policy" as well for example which have "single" and "multiple" radio options.

Zainzzkk (approved):

PR #3324, opened 2024-11-28
feat(EMS-4006): broker addresses - data saving

ttbarnes:
## Introduction :pencil2:
This PR adds data saving to the "Broker addresses" form.

- Update E2E test.
- Update `mapOrdnanceSurveyAddress` so that `postalCode` is `postcode` and `county` returns an empty string.
- Update Broker addresses POST controller to call `mapAndSave.broker`
- Update "Get application" GQL query to return broker address fields.
- Update `mapOrdnanceSurveyAddresses` so that the checkbox value is index based and has a `checked` value via a new helper function, `isChecked`.
- Update docker and github workflows to include Ordnance Survey environment variables.


src/ui/server/controllers/insurance/policy/broker-addresses/index.ts
```diff
@@ -154,6 +155,16 @@ export const post = async (req: Request, res: Response) => {
...
     }
 
+    const answer = payload[FIELD_ID];
+
+    const { __typename, ...chosenAddress } = addresses[answer];
+
+    const saveResponse = await mapAndSave.broker(chosenAddress, application);
+
+    if (!saveResponse) {
+      return res.redirect(PROBLEM_WITH_SERVICE);
```
abhi-markan, line 165: console.error?

src/ui/server/helpers/mappings/map-ordnance-survey-addresses/is-checked/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Check if a provided address is a submitted broker address.
+ * @param {MappedOrdnanceSurveyAddress} Ordnance survey address
+ * @param {ApplicationBroker} Application broker address
+ * @returns {Boolean}
+ */
+const isChecked = (address: MappedOrdnanceSurveyAddress, submittedAddress: ApplicationBroker) => {
+  const { addressLine1, addressLine2, town, postcode } = address;
+
+  if (
+    addressLine1 === submittedAddress.addressLine1 &&
```
abhi-markan, line 14: can this be a const please?
Can this be a ternary operator please?

abhi-markan (changes requested):
2 minor changes please, otherwise all looks good. Thanks

src/ui/server/controllers/insurance/policy/broker-addresses/index.ts
ttbarnes: We have this throughout. I don't think we need a console.error here - the flow is:

`POST controller` => `mapAndSave.broker` => `saveData` => API GQL handler

In `saveData`, an error will be thrown if the API call fails.

If ths API call fails, there will also be console errors and/or an error thrown.

So they'll be plenty of logs/errors if the data saving fails, but I don't see why not, to add a specific log.

Can look into this separately and apply throughout - have created EMS-4021 to do this.

src/ui/server/helpers/mappings/map-ordnance-survey-addresses/is-checked/index.ts
ttbarnes: No need for a ternary operator with consts

abhi-markan (approved):

src/ui/server/helpers/mappings/map-ordnance-survey-addresses/is-checked/index.ts
Zainzzkk, line 1: Are there any more fields that can come from OS or are they not mapped?
Could we otherwise use lodash or similar to check if both objects are the same?

ttbarnes: There are many OS fields, however EXIPS only needs a few fields - especially on the UI side and ultimately for data saving.

No need to worry about other properties in the UI that our API will not map or return, there's no reason to consume such fields.

This function is purely for design/UX, so that if a given Ordnance Survey address matches what's been saved in the database, a checkbox will be checked.

I do not see a reason to introduce lodash or a package to check if 4x values are the same. I also think object equality checking here could be risky especially as it's a small subset of data.

Would prefer to keep it simple and specific.

Zainzzkk (approved):

PR #3408, opened 2024-12-13
feat(EMS-4035): declarations - modern slavery - data migration

ttbarnes:
## Introduction :pencil2:
This PR adds a data migration script so that when `main-declarations-modern-slavery` is deployed, we can migrate previous applications data structure into the new data structure required for "declarations - modern slavery".

- Create data migration V3 to V4 index script.
- Create various migration helper functions.
- Move `getAllSubmittedApplications` and `createCuid` into a new parent helpers directory to be DRY.
- Create various migration functions so that when run, the following happens:
  - Application table - The V2->V3 field is replaced with a `migratedTo` field. `version` is updated.
  - A new `DeclarationModernSlavery` table is created.
  - A new `DeclarationModernSlaveryVersion` table is created.
  - For each existing declaration row - if an application has an `IN_PROGRESS` status, create `DeclarationModernSlavery` and `DeclarationModernSlaveryVersion` rows.
  - The `Declaration` table is updated appropriately.
  - Keys and constraints are added appropriately to the new tables.

- Minor documentation fixes/improvements.
- Data migration README improvements.
- Update `DECLARATION_VERSIONS` constants.


database/exip.sql
```diff
@@ -154,7 +154,7 @@ CREATE TABLE `Application` (
 	`dealType` varchar(4) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT 'EXIP',
   `policyContact` varchar(191) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   `nominatedLossPayee` varchar(191) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
-  `migratedV2toV3` tinyint(1) DEFAULT NULL,
```
Zainzzkk, line 157: should we not keep this flag for auditing purposes?

ttbarnes: I don't think it serves any value in all aspects. If we need to determine historic data migration versions from an auditing perspective, we can determine this from particular dates in the applications and release versions.

Also, if we were to keep the previous migration flag it:
- Could be confusing/misleading because e.g `migratedV1toV2` would not exist.
- The table would have a lot of "dead columns".

If this was a very high volume service and/or applications are long standing, then I would agree it's worth keeping all migration version flags in a new table, but at the moment, it is very low volume and low/no risk.

Also all data migration is quite minor really, excluding `migratedV1toV2` which had 100x more changes than V2 to V3 or V3 to V4.

src/api/data-migration/version-3-to-version-4/update-applications/update-declaration-modern-slaveries/create-declaration-modern-slavery-version-rows.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+const createDeclarationModernSlaveryVersionRows = async (connection: Connection) => {
+  const loggingMessage = 'Creating modern slavery version entries with modern slavery relationships';
+
+  console.info('✅ %s', loggingMessage);
+
+  try {
+    const modernSlaveries = await getAllDeclarationModernSlaveries(connection);
+
+    if (!modernSlaveries.length) {
```
Zainzzkk, line 25: should we not do the same !modernSlaveries || !modernSlaveries.length as other places?

ttbarnes: Nope, we have that check for "declarations for in progress applications", not "modern slaveries", as it works differently to the former.

This check (`getAllDeclarationModernSlaveries`); will always return an array.

Whereas the `getAllDeclarationsNonSubmittedApplications` check could return `false` or an array.

database/exip.sql
ttbarnes: @abhi-markan Can we get your input on this please.

abhi-markan: Is this migrating all applications?

Would it be possible to change the column to `migration` and specify the version i.e. `2`, `3` and etc.

Agree with @Zainzzkk we need to keep track of which application was migrated under which version.

src/api/data-migration/version-1-to-version-2/create-new-application-relationships/declaration-version-field.ts
```diff
@@ -18,7 +18,14 @@ const updateDeclarationVersionField = async (connection: Connection) => {
   console.info('✅ %s', loggingMessage);
 
   try {
-    const declarations = await getAllDeclarations(connection);
+    const declarations = await getAllDeclarationsNonSubmittedApplications(connection);
+
+    if (!declarations || !declarations.length) {
```
abhi-markan, line 23:
```suggestion
    if (!declarations?.length) {
```

src/api/data-migration/version-3-to-version-4/update-applications/update-declaration-modern-slaveries/create-declaration-modern-slavery-rows.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+const createDeclarationModernSlaveryRows = async (connection: Connection) => {
+  const loggingMessage = 'Creating modern slavery entries in the DeclarationModernSlavery table';
+
+  console.info('✅ %s', loggingMessage);
+
+  try {
+    const declarations = await getAllDeclarationsNonSubmittedApplications(connection);
+
+    if (!declarations || !declarations.length) {
```
abhi-markan, line 21:
```suggestion
    if (!declarations?.length) {
```

abhi-markan (changes requested):
Minor comments please.

src/api/data-migration/version-1-to-version-2/create-new-application-relationships/declaration-version-field.ts
ttbarnes: This is not possible because `.length` cannot exist. `getAllDeclarationsNonSubmittedApplications` can return a boolean or an array.

src/api/data-migration/version-3-to-version-4/update-applications/update-declaration-modern-slaveries/create-declaration-modern-slavery-rows.ts
ttbarnes: Same reply as above: https://github.com/UK-Export-Finance/exip/pull/3408#discussion_r1904160652

database/exip.sql
ttbarnes: Discussed offline - in summary:

- I've updated the script so we have a `migratedTo` integer field.
  - In progress applications will have this set to 4.
  - If any other applications have a `migratedV2toV3` value of 1 (`true`), `migratedTo` will be set to 3.
- `migratedV2toV3` has been deprecated.

RE migration - the data structure is changing for all applications regardless. However, only new data will be added for applications that are in progress. They will essentially gain an additional `DeclarationModernSlavery` relationship with the application, as this are now required questions.

abhi-markan (dismissed):

abhi-markan (approved):

PR #3411, opened 2024-12-14
feat(EMS-4034): declarations - modern slavery - update data model

ttbarnes:
## Introduction :pencil2:
- The "Declarations - Modern slavery" form will have conditional textarea fields.
- These fields need to be able to be saved and consumed.

- Update Keystone schema.
- Update SQL dump.
- Update API field IDs.
- Update the UI's "get application" GQL query.


src/api/schema.ts
```diff
@@ -715,6 +715,16 @@ export const lists = {
       willAdhereToAllRequirements: nullableCheckbox(),
       hasNoOffensesOrInvestigations: nullableCheckbox(),
       isNotAwareOfExistingSlavery: nullableCheckbox(),
+
+      cannotAdhereToAllRequirements: text({
```
Zainzzkk, line 719: do we need to add any tests?

Zainzzkk (approved):

PR #3437, opened 2024-12-23
feat(EMS-4059): task list - declarations - modern slavery

ttbarnes:
## Introduction :pencil2:
This PR updates the task list so that the "declarations and submit" task group requires the new "Modern slavery" declaration fields.

- Update the following functions to pass/consume a `declaration` param instead of `hasAntiBriberyCodeOfConduct`:
  - `generateGroupsAndTasks`
  - `requiredDeclarationsFields`
  - `submitApplicationTasks`
  - `createSubmitApplicationTasks`
- Update `canSubmitApplication` to _not_ flatten application data. This is now handled in `applicationIsComplete`.
- Create new `getModernSlaveryTasks` function.
- Update `requiredFields` (declarations) to consume `getModernSlaveryTasks`.

- Address TODO comments.
- Minor test description improvements.
- Remove unused helper function.


src/ui/server/helpers/can-submit-application/index.ts
```diff
@@ -22,9 +21,7 @@ const canSubmitApplication = (application: Application): boolean => {
   const { status, submissionCount, submissionDeadline } = application;
 
   // check the application is complete
-  const flatApplicationData = flattenApplicationData(application);
-
-  const isComplete = applicationIsComplete(flatApplicationData);
+  const isComplete = applicationIsComplete(application);
```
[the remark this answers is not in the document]
ttbarnes: It turns out this was not tested fully in the first place. Previously, the function called `flattenApplicationData` - the tests didn't call this. 

Now, this has been removed and so the test does not need to be updated

src/ui/server/helpers/required-fields/declarations/index.ts
```diff
@@ -31,16 +41,44 @@ export const getAntiBriberyCodeOfConductTasks = (hasAntiBriberyCodeOfConduct?: b
...
+/**
+ * getModernSlaveryTasks
+ * Get "modern slavery" tasks depending on the provided answers
+ * @param {ApplicationDeclarationModernSlavery} modernSlavery: Application "Modern slavery" answers
+ * @returns {Array} Modern slavery tasks
+ */
+export const getModernSlaveryTasks = (modernSlavery: ApplicationDeclarationModernSlavery) => {
+  const tasks = [WILL_ADHERE_TO_ALL_REQUIREMENTS, HAS_NO_OFFENSES_OR_INVESTIGATIONS, IS_NOT_AWARE_OF_EXISTING_SLAVERY];
+
+  if (modernSlavery.willAdhereToAllRequirements === false) {
```
abhi-markan, line 53: Switch-case please.

Zainzzkk (dismissed):

abhi-markan (changes requested):

src/ui/server/helpers/required-fields/declarations/index.ts
ttbarnes: I do you envison a switch/case here? The only way to do it would be like this:

```ts
switch (modernSlavery.willAdhereToAllRequirements) {
  case false:
    tasks.push(CANNOT_ADHERE_TO_ALL_REQUIREMENTS);
}

switch (modernSlavery.hasNoOffensesOrInvestigations) {
  case false:
    tasks.push(CANNOT_ADHERE_TO_ALL_REQUIREMENTS);
}

switch (modernSlavery.isNotAwareOfExistingSlavery) {
  case false:
    tasks.push(CANNOT_ADHERE_TO_ALL_REQUIREMENTS);
}
```

Which seems pointless / code smell, as it's a single `switch` and a single `case` for 3x different fields with individual conditions.

The if statements make a lot more sense IMO. If this was only for one field then i'd agree a switch/case would be cleaner.

abhi-markan: Thanks for the explanation.

abhi-markan (approved):

PR #3442, opened 2025-01-02
feat(EMS-4087): application submission - small export builder - non gbp

ttbarnes:
## Introduction :pencil2:
This PR updates the "eligible for small export builder" checks in application submission so that if the policy has a non GBP currency, we fetch the exchange rate and check if the value is less than the GBP threshold.

- Add E2E tests.
- Update `EXTERNAL_API_MAPPINGS` constants.
- Update `getSubmittedConfirmationTemplateId` to get an exchange rate via APIM if the policy currency code is not GBP.
- Create new `roundNumber` helper.
- Introduce APIM "currencies exchange rate" integration:
  - `apimCurrencyExchangeRate.get`
  - Add `getCurrenciesExchange` to `integrations/APIM`.
- Update test mocks and types.
- Create new `UNABLE_TO_DETERMINE_TEMPLATE_ID` constant. This will be very helpful for potential debugging in the future.

- Fix some typos.
- Improve `integrations/APIM` unit test coverage.


e2e-tests/insurance/cypress/e2e/journeys/application-submission/multiple-policy-type/policy-conditions/submit-an-application-multiple-policy-type-not-eligible-for-small-export-builder-alternative-currency.spec.js
```diff
@@ -0,0 +1,34 @@
...
+context(
+  'Insurance - submit an application - Multiple policy type - Not eligible for small export builder - Alternative currency - As an Exporter, I want to submit my completed credit insurance application, So that UKEF can process and make a decision on my application',
+  () => {
+    let referenceNumber;
+
+    before(() => {
+      cy.completeSignInAndSubmitAnApplication({
+        policyType: APPLICATION.POLICY_TYPE.MULTIPLE,
+        alternativeCurrencyPolicy: true,
+        maximumBuyerWillOwe: '5000000',
```
Zainzzkk, line 12: could this be a constant? or something like MAXIMUM_LIMIT +1?

src/api/emails/application/get-submitted-confirmation-template-id/index.ts
```diff
@@ -15,33 +17,52 @@ const {
...
+      if (eligibileForSmallExportBuilder) {
+        return CONFIRMATION.MULTIPLE_CONTRACT_POLICY.ELIGIBLE_FOR_SMALL_EXPORT_BUILDER_CONFIRMATION;
+      }
+
+      return CONFIRMATION.SINGLE_OR_MULTIPLE_CONTRACT_POLICY;
     }
 
-    return CONFIRMATION.SINGLE_OR_MULTIPLE_CONTRACT_POLICY;
-  }
+    return '';
```
Zainzzkk, line 60: could this be something else instead of an empty string?

src/api/emails/application/get-submitted-confirmation-template-id/index.ts
Zainzzkk, line 1: would it be better if this file was split into some helpers? for example for `if (policyCurrencyCode !== GBP) {`

e2e-tests/insurance/cypress/e2e/journeys/application-submission/multiple-policy-type/policy-conditions/submit-an-application-multiple-policy-type-not-eligible-for-small-export-builder-alternative-currency.spec.js
ttbarnes: I don't think it'd be wise because this is for an alternative currency.

If we use `SMALL_EXPORT_BUILDER.MAXIMUM_BUYER_WILL_OWE + 1` for example and the currency exchange rate changes, these tests could fail.

Therefore have opted for a very large amount here (5 million), which is well above `SMALL_EXPORT_BUILDER.MAXIMUM_BUYER_WILL_OWE`

src/api/helpers/get-APIM-currencies-exchange-rate/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        expect(response).toEqual(expected);
+      });
+    });
+
+    describe(`when the provided source is NOT ${GBP}`, () => {
+      it('should return an exchange rate ', async () => {
+        const response = await apimCurrencyExchangeRate.get(USD, mockTarget);
+
+        const expected = Number(Number(1 / mockCurrencyExchange.midPrice).toFixed(2));
```
Zainzzkk, line 36: is there a reason this has 2 Number wraps?

src/api/emails/application/get-submitted-confirmation-template-id/index.ts
ttbarnes: If we do that we'd need to have 2x return types e.g `String | null`, I think one type with an empty string is cleaner.

In any instance, the Notify call will fail due to unrecognised template ID.

Thoughts?

ttbarnes: I'm updating this to return a new constant, `UNABLE_TO_DETERMINE_TEMPLATE_ID` which would be very helpful for debugging.

Zainzzkk: I think a new constant would be better than an empty string

src/api/emails/application/get-submitted-confirmation-template-id/index.ts
ttbarnes: Great shout, have updated.

Zainzzkk (approved):

PR #3446, opened 2025-01-06
feat(EMS-4026): application submission - xlsx - modern slavery fields

ttbarnes:
## Introduction :pencil2:
This PR updates the XLSX sent during application submission to include the newly introduced "modern slavery" declaration fields.

- Create new `DECLARATIONS` XLSX row indexes.
- Update `XLSX_ROW_INDEXES`.
- Update and simplify API content strings.
- Create new `mapModernSlaveryFields` function and child functions.
- Update `mapDeclarations` to consume `mapModernSlaveryFields`.
- Create new `getPopulatedDeclaration` function.
- Update `getPopulatedApplication` to consume `getPopulatedDeclaration`.
- Update unit test mocks and types.

- Add additional application submission E2E tests.
- Improve `getPopulatedApplication` error handling.
- API unit test improvements.
- Add `objectHasKeysAndValues` helper function to the API.


src/api/constants/XLSX-CONFIG/INDEXES/DECLARATIONS/index.test.ts
```diff
@@ -0,0 +1,124 @@
...
+describe('api/constants/XLSX-CONFIG/INDEXES/DECLARATIONS', () => {
+  describe('DEFAULT_INDEXES', () => {
+    it('should return an object with indexes', () => {
+      const expected = {
+        CANNOT_ADHERE_TO_ALL_REQUIREMENTS: 0,
+        OFFENSES_OR_INVESTIGATIONS: 0,
+        AWARE_OF_EXISTING_SLAVERY: 0,
+      };
+
+      expect(DEFAULT_INDEXES()).toEqual(expected);
```
abhi-markan, line 20:
```suggestion
      const result = DEFAULT_INDEXES();
      expect(result).toEqual(expected);
```

ttbarnes: Good shout, have updated all instances of this.

src/api/constants/XLSX-CONFIG/INDEXES/DECLARATIONS/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const cannotAdhereAnswer = modernSlavery[CANNOT_ADHERE_TO_ALL_REQUIREMENTS];
+  const offensesInvestigationsAnswer = modernSlavery[OFFENSES_OR_INVESTIGATIONS];
+  const awareOfAnswer = modernSlavery[AWARE_OF_EXISTING_SLAVERY];
+
+  if (cannotAdhereAnswer) {
+    INDEXES.CANNOT_ADHERE_TO_ALL_REQUIREMENTS = 7;
+  }
+
+  if (offensesInvestigationsAnswer) {
```
abhi-markan, line 40: Not a big fan of nested if, can we please update the below logic to look something like this please.

```ts
if (offensesInvestigationsAnswer) {
  INDEXES.OFFENSES_OR_INVESTIGATIONS = 8 
    + (cannotAdhereAnswer ? 1 : 0) 
    + (awareOfAnswer ? 1 : 0);
}
```

or

```ts
if (offensesInvestigationsAnswer) {
const a = cannotAdhereAnswer ? 1 : 0;
const b = awareOfAnswer ? 1 : 0;

  INDEXES.OFFENSES_OR_INVESTIGATIONS = 8 + a + b;
}
```

src/api/generate-xlsx/map-application-to-XLSX/map-declarations/map-modern-slavery-fields/index.test.ts
```diff
@@ -0,0 +1,38 @@
...
+} = DECLARATIONS_FIELD_IDS;
+
+const CONTENT_STRINGS = DECLARATIONS_FIELDS.MODERN_SLAVERY;
+
+const {
+  declaration: { modernSlavery },
+} = mockApplication;
+
+describe('api/generate-xlsx/map-application-to-xlsx/map-declarations/map-modern-slavery-fields', () => {
+  it('should return an array of mapped declaration fields', () => {
```
abhi-markan, line 22: Can we please add an assertion for invalid argument or exception being thrown when calling `mapModernSlaveryFields`. Thanks

src/api/generate-xlsx/map-application-to-XLSX/map-declarations/map-modern-slavery-fields/index.ts
```diff
@@ -0,0 +1,39 @@
...
+const IS_NOT_AWARE_OF_EXISTING_SLAVERY_STRINGS = CONTENT_STRINGS[IS_NOT_AWARE_OF_EXISTING_SLAVERY];
+
+/**
+ * mapModernSlaveryFields
+ * Map an application's "modern slavery declaration" fields into an array of objects for XLSX generation
+ * @param {ApplicationDeclarationModernSlavery} modernSlavery
+ * @returns {Array<object>} Array of objects for XLSX generation
+ */
+const mapModernSlaveryFields = (modernSlavery: ApplicationDeclarationModernSlavery) => {
+  const mapped = [
```
abhi-markan, line 27: Do we expect any exception being thrown here, wondering if worth wrapping in `try-catch`?

src/api/helpers/get-populated-application/index.api-error.test.ts
```diff
@@ -279,6 +279,22 @@ describe('api/helpers/get-populated-application - error handling', () => {
...
+  it('should throw an error when declaration modernSlavery does not exist', async () => {
+    const mockApplication = {
+      ...fullApplication,
+      declaration: {
+        ...fullApplication.declaration,
+        modernSlaveryId: mockInvalidId,
+      },
+    };
+
+    try {
```
abhi-markan, line 291: Should the assertion not be written in a pattern where `response.rejects.toThrow(...)`?

src/api/helpers/get-populated-declaration/index.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param {Context} context: KeystoneJS context API
+ * @param {String} id: Declaration ID
+ * @returns {Promise<ApplicationDeclaration>}
+ */
+const getPopulatedDeclaration = async (context: Context, id: string) => {
+  try {
+    console.info('Getting populated declaration %s', id);
+
+    const declaration = await getDeclarationById(context, id);
+
```
abhi-markan, line 17: Should we validate `declaration` response?

abhi-markan (changes requested):
Nice work few minor comments to address please.

src/api/constants/XLSX-CONFIG/INDEXES/DECLARATIONS/index.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  describe('DECLARATIONS_INDEXES', () => {
+    const mockPopulatedModernSlavery = {
+      ...mockApplication.declaration.modernSlavery,
+      [CANNOT_ADHERE_TO_ALL_REQUIREMENTS]: 'mock',
+      [OFFENSES_OR_INVESTIGATIONS]: 'mock',
+      [AWARE_OF_EXISTING_SLAVERY]: 'mock',
+    };
+
+    describe(`when ${CANNOT_ADHERE_TO_ALL_REQUIREMENTS}, ${OFFENSES_OR_INVESTIGATIONS} and ${AWARE_OF_EXISTING_SLAVERY} are populated`, () => {
+      it('should return an object with indexes', () => {
```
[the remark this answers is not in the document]
ttbarnes: I'm not sure what you mean RE an assertion with `DECLARATIONS_INDEXES` as `id: 123`?

No property could have a field with the name of `id`, it's not in `ApplicationDeclarationModernSlavery`. The only fields that will be in `declaration.modernSlavery` are the ones defined in the `ApplicationDeclarationModernSlavery` type - there was actually 3x missing fields so i've added them in.

In this function (`DECLARATIONS_INDEXES`) we only need to specifically check 3x fields:

```
cannotAdhereToAllRequirements
offensesOrInvestigations
awareOfExistingSlavery
```

These are string fields - If any one of these is populated, the indexes need to be incremented.

If they are not populated, they will only ever be empty strings. So for each field we can just simply check `modernSlavery.fieldX` 😎

src/api/constants/XLSX-CONFIG/INDEXES/DECLARATIONS/index.ts
ttbarnes: I see what you mean, but I think the proposed approach is harder to read/follow.

We use nested if statements for all relevant XLSX indexing conditions. Previously, we had many bugs RE XLSX indexing, due to the sheer volume of potential conditions - they were very hard to debug and test all the different indexes, since then have tried to keep the mapping logic as simple and easy to follow as possible.

I agree we should steer away from nested if statements where possible, but I think there's a balance between readability and maintainability. With the above proposal, if we need to add another condition in the future, we've had to revert back to nested if statements.

I recommend we stick to the current approach RE above comments and to be consistent with other XLSX indexing.

Otherwise to be aligned/consistent, we would need to update other indexing to use ternaries where appropriate, but this would be out of scope and involve retesting a majority of the XLSX.

Interesting topic!

src/api/generate-xlsx/map-application-to-XLSX/map-declarations/map-modern-slavery-fields/index.ts
ttbarnes: I cannot think of a scenario where this - or any other XLSX mapping function would throw an error because:

1. We only map XLSX data when an application is submitted.
2. When an application is created (way before submission), it has `declaration` and `modernSlavery` relationships.
3. The mapping simply creates an object with 1x property from content strings, and a top level value from `modernSlavery`, e.g `modernSlavery[WILL_ADHERE_TO_ALL_REQUIREMENTS]`. 
4. When this function is called during submission - all values passed to `xlsxRow` will always be present.
5. An application cannot be submitted without these answers being provided (e.g `modernSlavery[WILL_ADHERE_TO_ALL_REQUIREMENTS]`). If this (for some bizarre reason), is not populated, it would be gracefully handled via `xlsxRow` and render an empty string.
6. I can only think of this being a concern if someone tries to submit an application directly via the API from another system - which we do not allow.

Also just to note:

- These mapping functions are tried and tested, currenty live in production.
- We have a `console.info` in `xlsxRow` that logs out the field name - whilst it's not a try/catch - has been very helpful for debugging any mapping issues before.

src/api/helpers/get-populated-declaration/index.ts
ttbarnes: Good suggestion - I don't see a value in doing this though.

`getDeclarationById` calls and returns a KeystoneJS method which we have no control over - `context.db.Declaration.findOne`.

In this helper (`getPopulatedDeclaration`), we only consume `declaration.modernSlaveryId`, which we call `getDeclarationModernSlaveryById` with.

If `declaration.modernSlaveryId` doesn't exist or is an invalid ID etc, it'll throw an error in `getDeclarationModernSlaveryById`'s try/catch handler.

src/api/generate-xlsx/map-application-to-XLSX/map-declarations/map-modern-slavery-fields/index.test.ts
abhi-markan: Missed this one.

ttbarnes: Forgot to reply - same reply as here https://github.com/UK-Export-Finance/exip/pull/3446/files#r1907205946

abhi-markan (dismissed):

src/api/helpers/get-populated-application/index.ts
```diff
@@ -99,11 +119,26 @@ const getPopulatedApplication = async ({
...
+    /**
+     * Check that all expected relationships are populated.
+     * If not, throw an error.
+     * Otherwise, unexpected behaviours could occur.
+     */
+    Object.keys(populatedApplication).forEach((relationshipKey: string) => {
+      if (EXPECTED_RELATIONSHIPS.includes(relationshipKey)) {
+        const populatedRelationship = populatedApplication[relationshipKey];
+
+        if (!populatedRelationship || !objectHasKeysAndValues(populatedRelationship)) {
```
abhi-markan, line 131: I would expect this function to evaluate whether `objectHasKeysAndValues` is falsy or not.

src/api/test-helpers/declaration.ts
```diff
@@ -45,9 +45,35 @@ const get = async (context: Context, declarationId: string) => {
...
+    console.info('Updating a declaration by ID (test helpers)');
+
+    const declaration = await context.db.Declaration.updateOne({
+      where: { id: declarationId },
+      data,
+    });
+
+    return declaration;
+  } catch (error) {
+    console.error(error);
```
abhi-markan, line 67: Add some text please.

abhi-markan (changes requested):

src/api/helpers/get-populated-application/index.ts
ttbarnes: This function only needs to check for false/negative scenarios i.e:

- `populatedApplication[relationship]` is false
- `populatedApplication[relationship]` is an object without keys or values 

We do not need to check if `objectHasKeysAndValues` is true, the function is purely for throwing an error (for debugging), if either of the above scenarios are met

src/api/test-helpers/declaration.ts
ttbarnes: Agree this would be nice to have.

We have 37 test helper files with multiple functions containing this; Do not want to bloat this PR or become misaligned - have created EMS-4099 to tackle this in a different PR.

ttbarnes: I would just note that these files are used purely for unit testing purposes.

abhi-markan: As long as it will rectified in this PR EMS-4099 I am happy to approve for now.

src/api/helpers/get-populated-application/index.ts
abhi-markan: I think the function can incorporate falsy argument check before validating above mentioned scenarios and then you if condition will be a simple function call.

ttbarnes: I'm not following what you mean @abhi-markan can you provide an example please? The function is checking if any relationship is false or has no key/values.

I do not see how the implementation could be cleaner/simpler

ttbarnes: It's checking the nested properties in `populatedApplication` i.e

```ts
const populatedApplication = {
  relationshipA, 
  relationshipB,
  relationshipC,
  // etc
};
```

So this line:

```ts
if (!populatedRelationship || !objectHasKeysAndValues(populatedRelationship)) {
```

Is checking relationshipA, relationshipB etc

ttbarnes: Updated as discussed 👍

abhi-markan (approved):
LGTM 👍🏼

PR #3461, opened 2025-01-08
feat(EMS-4008): broker addresses - save and back

ttbarnes:
## Introduction :pencil2:
This PR adds "save and back" functionality ot the "broker addresses" form.

- Create new `BROKER_ADDRESSES_SAVE_AND_BACK` route constant.
- Create new "save and back" POST route and controller.
- Update `mapSubmittedData` to delete `SELECT_THE_ADDRESS`.
- Create a new DRY function to get a chosen Ordnance Survey address, `getChosenOrdnanceSurveyAddress`.
- Add E2E test.


abhi-markan (dismissed):

src/ui/server/controllers/insurance/policy/broker-addresses/save-and-back/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    const payload = constructPayload(req.body, [FIELD_ID]);
+
+    const validationErrors = generateValidationErrors(payload, FIELD_ID, ERROR_MESSAGE);
+
+    if (!validationErrors) {
+      const { postcode, buildingNumberOrName } = broker;
+
+      const response = await api.keystone.getOrdnanceSurveyAddresses(String(postcode), String(buildingNumberOrName));
+
+      const { addresses } = response;
```
abhi-markan, line 37: This will fail if `response` is false, do we want to validate `response`?

abhi-markan (changes requested):
A comment to address please.

src/ui/server/controllers/insurance/policy/broker-addresses/save-and-back/index.ts
ttbarnes: No need to 😎 

1. `api.keystone.getOrdnanceSurveyAddresses` checks the API response and returned data. If no data is returned (which will contain `addresses`, `api.keystone.getOrdnanceSurveyAddresses` will throw an error.
  1a. Can see all the checks from [this file](https://github.com/UK-Export-Finance/exip/blob/main-broker-address-lookup/src/ui/server/api/keystone/get-ordnance-survey-addresses/index.ts#L20)
3. If any of the above errors occur, `broker-addresses/save-and-back` will in turn, fall into it's own try/catch handler.
4. So if the API response fails, it won't get to line 37 where it tries to obtain `addresses`.

By doing it this way, we avoid having to check the API response object in multiple controllers and instead, it's handled in the UI's DRY API call.

abhi-markan (approved):

PR #3468, opened 2025-01-09
feat(EMS-4015): application submission - xlsx - uk broker address

ttbarnes:
## Introduction :pencil2:
This PR updates the application submission's XLSX mapping to handle UK broker addresses.

Previously, the XLSX would only render a single `FULL_ADDRESS` string. Now, UK addresses are supported which contain multiple fields. Therefore we need to map these fields/substrings into a single string.

- Create new `mapBrokerAddress` helper function.
- Create new `mapBrokerAddressBasedInTheUk` helper function.
- Update `mapBroker` to consume `mapBrokerAddress`.
- Add `isEmptyString` helper function to the API.
- Update test mocks, types and mocks.


src/api/generate-xlsx/map-application-to-XLSX/map-policy/map-broker/map-broker-address/based-in-the-uk/index.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const {
+  BROKER_DETAILS: { BUILDING_NUMBER_OR_NAME, ADDRESS_LINE_1, ADDRESS_LINE_2, TOWN, COUNTY, POSTCODE },
+} = FIELD_IDS;
+
+const { broker } = mockApplication;
+
+const { buildingNumberOrName, addressLine1, addressLine2, town, county, postcode } = broker;
+
+describe('api/generate-xlsx/map-application-to-xlsx/map-policy/map-broker/map-broker-address/based-in-the-uk', () => {
```
abhi-markan, line 14: An assertion where all fields are '' please.

src/api/helpers/string/index.test.ts
```diff
@@ -0,0 +1,21 @@
+import { isEmptyString } from '.';
+
+describe('api/helpers/string', () => {
+  describe('isEmptyString', () => {
+    describe('when a string is empty', () => {
+      it('should return true', () => {
```
abhi-markan, line 6: Can you add one for `null` and `undefined` please.

src/api/helpers/string/index.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * isEmptyString
+ * Check if a string is empty
+ * @param {String}
+ * @returns {Boolean}
+ */
+export const isEmptyString = (str: string) => str === '';
```
abhi-markan, line 7: Do you want to `str.trim()`?

abhi-markan (changes requested):

src/api/generate-xlsx/map-application-to-XLSX/map-policy/map-broker/map-broker-address/based-in-the-uk/index.test.ts
ttbarnes: Good shout, added.

src/api/helpers/string/index.test.ts
ttbarnes: The only way to do this would be to add a `ts-ignore` for both `null` and `undefined` because the required type is a string. I don't think we need to do this.

src/api/helpers/string/index.ts
ttbarnes: We could, but would like to do a separate PR so that we can track and test it properly and progress this PR. Have created EMS-4103 to tackle this.

src/api/helpers/string/index.test.ts
abhi-markan: Apologies, I didn't notice `string` type.

abhi-markan (approved):

PR #3473, opened 2025-01-10
feat(EMS-3981-3991): broker address lookup - mapping updates

ttbarnes:
## Introduction :pencil2:
- Ordnance Survey integration was not rendering addresses in the UI with the building number.
- Ordnance Survey integration was not working as expected when multiple results for 1x building name/number were found.

- Update E2E commands to pass and consume new arguments:
  - `brokerBuildingNumberOrName`
  - `brokerPostcode`
- Add new E2E tests for submitting the "broker details" form and submitting an application as a building name or number.
- Update `getOrdnanceSurveyAddresses` GQL resolver to always uppercase the building name/number (The Ordnance Survey data we consume is in uppercase).
- Update the Ordnance Survey integration to filter _and_ map the results.
- Update `mapOrdnanceSurveyAddress` to:
  - Include the building number.
  - Simplify string concatenation.

- Rename `OrdnanceSurveyResponse` type to `OrdnanceSurveyAddress`.
- Minor test description, documentation improvements.


src/api/custom-resolvers/queries/get-ordnance-survey-addresses/index.test.ts
```diff
@@ -95,4 +81,18 @@ describe('getOrdnanceSurveyAddresses', () => {
...
+    beforeEach(() => {
+      ordnanceSurvey.get = mockSpyPromiseRejection;
+    });
+
+    it('should return object containing success=false and apiError=true', async () => {
+      const response = await getOrdnanceSurveyAddress({}, { postcode, houseNameOrNumber });
+
+      const expected = { success: false, apiError: true };
+
+      expect(response).toEqual(expected);
```
abhi-markan, line 95: Should this not be `rejects.toThrow(...)`, also can we add an assertion to check the error message text please?

src/api/helpers/filter-ordnance-survey-addresses/index.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      const result = filterOrdnanceSurveyAddresses(ordnanceSurveyResponse, substring);
+
+      const expected = [address];
+
+      expect(result).toEqual(expected);
+    });
+  });
+
+  describe('when no address is found', () => {
+    it('should return an empty array', () => {
```
abhi-markan, line 64: Can we add an assertion where `null` is returned from the function please.

src/api/helpers/map-ordnance-survey-address/index.ts
```diff
@@ -37,7 +44,7 @@ const mapOrdnanceSurveyAddress = (address: OrdnanceSurveyResponse): MappedOrdnan
   const county = '';
 
   return {
-    addressLine1,
+    addressLine1: addressLine1.trim(),
```
abhi-markan, line 47:
```suggestion
    addressLine1,
```

src/api/helpers/map-ordnance-survey-address/index.ts
```diff
@@ -37,7 +44,7 @@ const mapOrdnanceSurveyAddress = (address: OrdnanceSurveyResponse): MappedOrdnan
   const county = '';
```
abhi-markan, line 45:
```suggestion
const addressLine1 = addressLine1.trim();
```

src/api/helpers/map-ordnance-survey-addresses/index.test.ts
```diff
@@ -0,0 +1,15 @@
+import mapOrdnanceSurveyAddresses from '.';
+import mapAddress from '../map-ordnance-survey-address';
+import mockOrdnanceSurveyResponse from '../../test-mocks/mock-ordnance-survey-response';
+
+const ordnanceSurveyResponse = mockOrdnanceSurveyResponse.results;
+
+describe('api/helpers/map-ordnance-survey-addresses', () => {
+  it('should return an array of mapped addresses', () => {
+    const result = mapOrdnanceSurveyAddresses(ordnanceSurveyResponse);
```
abhi-markan, line 9: Can this function ever return an empty array, if yes then an assertion for the same please.

abhi-markan (changes requested):
Nice work, minor comments to address please.

src/api/custom-resolvers/queries/get-ordnance-survey-addresses/index.test.ts
ttbarnes: Good spot - have updated to throw an error in the catch handler.

src/api/helpers/filter-ordnance-survey-addresses/index.test.ts
ttbarnes: If `.filter` returns `null`, the `filtered` const returns an empty ar
ray so we cannot assert null values. I have double checked, it's handled in this unit test.[](url)

Test coverage for this is reported at 100%.

src/api/helpers/map-ordnance-survey-address/index.ts
ttbarnes: Moved the trim to above the return.

src/api/helpers/map-ordnance-survey-address/index.ts
ttbarnes: Moved the trim to above the return.

src/api/helpers/map-ordnance-survey-addresses/index.test.ts
ttbarnes: Yes, but we don't need to assert this. The implementation uses `.map` - the unit test mimics this.

If we were to add a test for an empty array, we would testing the native `.map` behaviour.

The test only needs to ensure that it's mapping over the addresses with `mapAddress`

Unit test coverage is reported at 100%.

src/api/helpers/filter-ordnance-survey-addresses/index.test.ts
abhi-markan: The only reason I ask is you are returning null at LOC:28, ideally a test case which can reach that point in code.

src/api/custom-resolvers/queries/get-ordnance-survey-addresses/index.test.ts
```diff
@@ -95,4 +81,14 @@ describe('getOrdnanceSurveyAddresses', () => {
...
     });
   });
+
+  describe('when ordnanceSurvey API throws an error', () => {
+    beforeEach(() => {
+      ordnanceSurvey.get = mockSpyPromiseRejection;
+    });
+
+    it('should throw an error', async () => {
+      await expect(getOrdnanceSurveyAddress({}, { postcode, houseNameOrNumber })).rejects.toThrow('Getting Ordnance Survey addresses');
```
abhi-markan, line 91:
```suggestion
      const response = await getOrdnanceSurveyAddress({}, { postcode, houseNameOrNumber });
      expect(response).rejects.toThrow('Getting Ordnance Survey addresses');
```

src/api/custom-resolvers/queries/get-ordnance-survey-addresses/index.test.ts
abhi-markan: Thanks

abhi-markan (changes requested):
2 comments to address please.

src/api/helpers/filter-ordnance-survey-addresses/index.test.ts
ttbarnes: I think it's slightly overkill because `.filter` handles it natively, but I have split up individual address filtering into a new function `filterOrdnanceSurveyAddress`, so we can test the `null` return.

src/api/custom-resolvers/queries/get-ordnance-survey-addresses/index.test.ts
ttbarnes: Great shout, updated.

abhi-markan (approved):

PR #3475, opened 2025-01-13
feat(EMS-4054): broker addresses - single address redirection

ttbarnes:
## Introduction :pencil2:
This PR updates the broker address lookup functionality so that if a broker is based in the UK and only one address is returned from Ordnance Survey, the user is redirected (from `BROKER_DETAILS`) to `BROKER_CONFIRM_ADDRESS`, instead of `BROKER_ADDRESSES`.

Otherwise, the user would be selected only one available option from the `BROKER_ADDRESSES` form.

This PR also updates the "confirm broker address" page to render an address appropriately via a helper function, depending on if the address was entered manually, or obtained via Ordnance Survey.

- Update the `BROKER_ADDRESSES` GET controller to map/save data and redirect to `BROKER_CONFIRM_ADDRESS` if only one address is returned from the API.
- Rename `getChosenOrdnanceSurveyAddress` to `getOrdnanceSurveyAddressById`.
- Create new `getChosenOrdnanceSurveyAddressByIndex` helper function.
- Create new `generateBrokerAddressInsetTextHtml` helper function.
- Update E2E tests.
- Create new cypress command, `cy.clickZeroAddressesEnterManuallyLink();`.

- Create new E2E constants, `ADDRESS_LOOKUP_INPUT_EXAMPLES`.
- E2E improvements, address a TODO comment.


src/ui/server/controllers/insurance/policy/broker-addresses/index.ts
```diff
@@ -103,6 +104,24 @@ export const get = async (req: Request, res: Response) => {
 
     const { addresses } = response;
 
+    /**
+     * If only one address is found - no need to ask the user to select an address. Therefore:
+     * 1) Generate an address to save.
+     * 2) Send to the API.
+     * 3) Redirect to the next part of the flow.
+     */
+    if (addresses.length === 1) {
```
abhi-markan, line 113: Is there a reason for a `1` length check?

src/ui/server/helpers/get-chosen-ordnance-survey-address/by-index/index.ts
```diff
@@ -0,0 +1,31 @@
+import { MappedOrdnanceSurveyAddress } from '../../../../types';
+
+interface GetChosenOrdnanceSurveyAddressByIndexParams {
```
abhi-markan, line 3: Should this be declared somewhere in a centralised interface collection file?

abhi-markan (changes requested):

src/ui/server/controllers/insurance/policy/broker-addresses/index.ts
ttbarnes: Yes, it's in the PR description and above documentation (line 108) 😛 

If only 1 address is returned, we need to redirect a user to a different URL (to skip 1x step). Otherwise a user is redirected to a form where there is only possible option to select.

src/ui/server/helpers/get-chosen-ordnance-survey-address/by-index/index.ts
ttbarnes: Yes - this PR is a WIP, need another PR merged first.

abhi-markan (approved):

src/ui/server/helpers/generate-broker-address-inset-text-html/index.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+      expect(result).toEqual(expected);
+    });
+  });
+
+  describe(`when ${FULL_ADDRESS} is NOT populated`, () => {
+    it('should return the result of generateMultipleFieldHtml', () => {
+      const mockBroker = {
+        ...broker,
+        [FULL_ADDRESS]: '',
```
abhi-markan, line 33: Can `FULL_ADDRESS` ever be an empty array?

src/ui/server/helpers/generate-broker-address-inset-text-html/index.ts
```diff
@@ -0,0 +1,44 @@
...
+const generateBrokerAddressInsetTextHtml = (broker: ApplicationBroker) => {
+  if (broker[FULL_ADDRESS]) {
+    /**
+     * Transform \r\n characters into <br />,
+     * so that the string will render with multiple lines.
+     */
+    return replaceNewLineWithLineBreak(broker[FULL_ADDRESS]);
+  }
+
+  const { addressLine1, addressLine2, town, county, postcode } = broker;
```
abhi-markan, line 29: Will it ever be possible for one of thes properties to not exists, wondering if this will be the case then this function will throw an exception and to be caught elsewhere in the call stack.

abhi-markan (changes requested):

src/ui/server/helpers/generate-broker-address-inset-text-html/index.test.ts
ttbarnes: No - it will only ever be a string. As with all other fields obtained from the DB:

- Keystone API schema defines the data type.
- TypeScript and GraphQL enforce the type and only returns a string.

src/ui/server/helpers/generate-broker-address-inset-text-html/index.ts
ttbarnes: Good question - they will always exist 😎 

When an application is created, a wide range of relationships are created, e.g `broker`, `policy` etc. 

If any of these relationships are _not_ created, the service will error and it would not be possible to access the application. The GraphQL call would fail.

When each relationship is created, it has default/empty fields for that relationship. Relating to the previous comment - in this case, all of the `broker` relationship's fields will be empty strings (`addressLine1`, `town`, `postcode` etc)

Also FYI for more context, **all** of the "insurance" UI routes consume the application from one single piece of middleware, which gets application data from the API via one DRY GraphQL call which contains the application fields:

https://github.com/UK-Export-Finance/exip/blob/main/src/ui/server/middleware/insurance/get-application-by-reference-number/index.ts

abhi-markan (approved):

PR #3477, opened 2025-01-14
chore(EMS-4011-4116): policy - check your answers - e2e improvements

ttbarnes:
## Introduction :pencil2:
- The "policy" summary list was not rendering a UK based broker address as a multi line string.
- "Policy - check your answers" has some missing/misaligned E2E test coverage for multiple contract policy conditions.

- Update `brokerAddressField` to use `generateMultipleFieldHtml` instead of `mapBrokerAddressBasedInTheUk`.
- Remove now unused `mapBrokerAddressBasedInTheUk` UI summary list helper.
- Update "Policy - check your answers" E2E tests to:
  - Be split into single/multiple directories.
  - Have aligned single/multiple tests, along with some simplification. (Multiple was missing some tests that only single had)
- Update `checkPolicySummaryList`.
- Create new E2E constant examples:
  - `EXPECTED_TREASURY_SINGLE_LINE_STRING`
  - `ORDNANCE_SURVEY_EXAMPLES`

- Remove unnecessary `referenceNumber` params from the following cypress commands:
  - `completeAndSubmitDeclarationsForms`
  - `completePrepareApplicationSinglePolicyType`


e2e-tests/commands/insurance/declarations/complete-and-submit-declarations-forms.js
```diff
@@ -3,10 +3,9 @@
  * completes declarations forms up to the specified form to stop at
  * eg, when 'antiBribery' is passed, it will complete all forms up to and including 'antiBribery'
  * @param {String} stopSubmittingAfter: The final form to submit
- * @param {String} referenceNumber: Application reference number
  */
-const completeAndSubmitDeclarationsForms = ({ stopSubmittingAfter, referenceNumber }) => {
-  cy.completePrepareApplicationSinglePolicyType({ referenceNumber });
+const completeAndSubmitDeclarationsForms = ({ stopSubmittingAfter }) => {
+  cy.completePrepareApplicationSinglePolicyType({});
```
abhi-markan, line 8:
```suggestion
  cy.completePrepareApplicationSinglePolicyType();
```

Wondering if it needs an empty object as an argument?

e2e-tests/insurance/cypress/e2e/journeys/task-list/complete-prepare-application-group.spec.js
```diff
@@ -16,7 +16,7 @@ context('Insurance - Task list - complete `prepare application` group', () => {
 
       cy.assertUrl(url);
 
-      cy.completePrepareApplicationSinglePolicyType({ referenceNumber });
+      cy.completePrepareApplicationSinglePolicyType({});
```
abhi-markan, line 19:
```suggestion
      cy.completePrepareApplicationSinglePolicyType();
```

abhi-markan (changes requested):

e2e-tests/commands/insurance/declarations/complete-and-submit-declarations-forms.js
ttbarnes: It is required because [completePrepareApplicationSinglePolicyType](https://github.com/UK-Export-Finance/exip/blob/main/e2e-tests/commands/insurance/complete-prepare-application-section-single-policy-type.js#L45) has object structured params. Without an empty object param, the command will fail.

We have this pattern throughout.

It might be possible to declare a default empty object, but we should look into this outside of this PR

e2e-tests/insurance/cypress/e2e/journeys/task-list/complete-prepare-application-group.spec.js
ttbarnes: Same reply as above

e2e-tests/commands/insurance/declarations/complete-and-submit-declarations-forms.js
abhi-markan: Happy to approve this if you can please create a back log ticket for above please to be picked up as a seprate PR. Ideally we should never pass an empty object. Thanks

abhi-markan (approved):
Happy to approve this if you can please create a back log ticket for above please to be picked up as a seprate PR. Thanks

e2e-tests/commands/insurance/declarations/complete-and-submit-declarations-forms.js
ttbarnes: Have created EMS-4123 to investigate this. Very unlikely to be picked up anytime soon. There are some instances where we will want and need to pass an empty object, but otherwise agree would be great to avoid this.

PR #3483, opened 2025-01-16
fix(EMS-4125): update keystone config

abhi-markan:
## Introduction :pencil2:
API container boot failure due to outdated keystone config.

* Executed `npm run dev` on `api` container.


ttbarnes (approved):
Great work

PR #3507, opened 2025-01-28
chore(EMS-4155): apim notify integration

ttbarnes:
## Introduction :pencil2:
- APIM has been updated so that file buffers can be sent to Notify.
- EXIPS can therefore now call APIM to send emails, instead of calling Notify directly.

- Create new `sendEmail` APIM integration function.
- Update relevant helper functions.
- Remove `notifications-node-client` NPM dependency.
- Create new types.

- Various documentation improvements.
- Improve some API error handling unit tests.
- Improve APIM integration validation checks


src/api/integrations/APIM/index.ts
```diff
@@ -115,6 +121,54 @@ const APIM = {
...
+          sendToEmailAddress,
+          personalisation,
+        },
+        validateStatus(status) {
+          const acceptableStatus = [201];
+          return acceptableStatus.includes(status);
+        },
+      });
+
+      if (response.data && response.status === 201) {
```
abhi-markan, line 155:
```suggestion
      if (response?.data && response?.status === 201) {
```

abhi-markan: Suggest using HttpStatusCode not sure if this is something EXIP uses widely?

abhi-markan: wondering if you can use `validateStatus` instead of explicit status check to `201` since same operation is being performed when calling `validateStatus`.

abhi-markan (changes requested):
Nice work, few comments to address additonally can you als please update `.env.sample` with removed `GovNotify` env variables please.

ttbarnes:
> Nice work, few comments to address additonally can you als please update `.env.sample` with removed `GovNotify` env variables please.

Thanks @abhi-markan - we cannot update `.env.sample`, the GOV Notify variables are still required. APIM requires `GOV_NOTIFY_API_KEY` to be sent and the `EMAIL_RECIPIENT` variables are used for testing purposes.

src/api/integrations/APIM/index.ts
ttbarnes: It doesn't seem wise to remove the specific status check.

If APIM returns a 500 status, there will be no `response.status` and instead a `response.statusCode`. 

In addition, there could or could not be a `data` object.

Therefore I think it should remain as is, as a fail safe. Especially considering that this call is absolutely essential for the entire EXIPS service.

There could be many different combinations of responses. The current specific status check - with the check to ensure a data object is in the response, seems to be a failsafe check for a successful response.

ttbarnes: We don't have `HttpStatusCode` in EXIP unfortunately - would be introduce to have in the future.

EXIPS only has ~10 instances of status specific checks.

abhi-markan: Apologies for the confusion, I meant it will be wise to do following to keep the code DRY?

```ts
const isValidResponse = response.data && validateStatus(response.status);

if(isValidResponse) {...
```

The function is Axios call can be centralised.

```ts
 const validateStatus = (status) => {
          const acceptableStatus = [201];
          return acceptableStatus.includes(status);
        },
```

abhi-markan (changes requested):

abhi-markan (approved):

PR #3544, opened 2025-03-20
feat(EMS-4156): acr purge workflow

abhi-markan:
## Introduction :pencil2:
Ensure ACR old artefacts are purged after specified number of days have elapsed.

* Created `purge.yml` GHA workflow.
* Created GHA compose module for AZ CLI ACR purge.
* Created GHA composite module for dispatching notifications to MS Teams.

* Updated deployment GHA workflow to consume `notify` composite module.
* Added ACR to cspell.json.


.github/actions/acr/action.yml
```diff
@@ -0,0 +1,49 @@
...
+  credentials:
+    description: 'Azure subscription and resource group service principal'
+    required: true
+  subscription:
+    description: 'Azure subscription ID'
+    required: true
+  group:
+    description: 'Azure resource group'
+    required: true
+  acr:
```
tony-barnes-ukef, line 17: Can we be explicit in the `actions/acr` directory? `actions/azure-container-registry`?

I had no idea what ACR meant until I read this.

.github/workflows/purge.yml
```diff
@@ -0,0 +1,120 @@
...
+            [
+            {
+            "os": "default",
+            "uri": "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}",
+            },
+            ],
+            },
+            ],
+            }
+            '
```
tony-barnes-ukef, line 120: Indentation above?

.github/workflows/purge.yml
```diff
@@ -0,0 +1,120 @@
+name: Purge
```
tony-barnes-ukef, line 1: Should we call this and the workflow yml, Azure Purge?

Otherwise, If we need to purge something else in the future, we'd either bloat this workflow or have to rename

tony-barnes-ukef (changes requested):

.github/workflows/purge.yml
abhi-markan: This is just one worded name, but the run name is set to `run-name: 🧹Azure purge` on the line below. Thanks.

.github/actions/acr/action.yml
abhi-markan: Good point, have amended.

tony-barnes-ukef (approved):
Not an expert on this but LGTM

PR #3556, opened 2025-05-09
feat(EMS-4222): upgraded GovUK package to 5.10.0

abhi-markan:
# Pull Request

## Introduction :pencil2:

Update GovUK package to `5.10.01.

* UI re-build.
* SRI updates.


src/ui/server/constants/integrity.ts
zainUKEF, line 1: do any other integrities need to be updated?

zainUKEF (approved):

src/ui/server/constants/integrity.ts
abhi-markan: No just this, this does mean our E2E tests do not pick up failing SRIs which also needs to be updated so it can be caught during the CICD.

zainUKEF: I think it is very difficult to test them in e2e tests when the SRI fails

abhi-markan: We test them in DTFS.

PR #3594, opened 2025-08-29
fix(EMS-4233): reverted Serbia country code

abhi-markan:
# Introduction :pencil2:

ISO country code for Serbia to be reverted back to `SCG` untill `SRB` has been implemented in downstream systems.

* Reverted back to `SCG`.

* Dependencies updates.
* Schema updates.


tony-barnes-ukef (approved):
It'd be wise to add an E2E tests for this country - in the quote tool and the application, since we're forced to have a country with multiple codes.
 
This way if EXIP E2E tests are running - and say, the CIS code is changed or removed, the test will fail and we know something needs to change.

Otherwise LGTM

